PR #26, opened 2024-09-19
feat(loaders): document loader pdf

### LINKED ISSUE #24: feat: Add PDF Loader to Document Loaders in Rig

## Feature Request
Add a PDF Loader to the Document Loaders in Rig

### Motivation
PDF is a widely used format for storing and sharing documents. Adding support for loading PDF files would significantly enhance Rig's capability to process and analyze a broader range of document types. This feature would allow users to easily incorporate PDF documents into their RAG (Retrieval-Augmented Generation) systems and other LLM tasks.

Use cases include:
1. Extracting information from technical documentation stored in PDFs
2. Analyzing academic papers and research reports
3. Processing business documents and reports
4. Incorporating legal documents into NLP workflows

### Proposal
Implement a `PdfLoader` as part of the `document_loaders` module. The implementation should:

1. Create a new file `src/document_loaders/pdf.rs`
2. Implement a `PdfLoader` struct that implements the `DocumentLoader` trait
3. Use the `lopdf` crate for parsing PDF files
4. Extract text content from PDF documents
5. Convert extracted content into `DocumentEmbeddings`

Basic structure:

```rust
use async_trait::async_trait;
use lopdf::Document;
use crate::embeddings::DocumentEmbeddings;
use super::DocumentLoader;

pub struct PdfLoader {
    path: String,
}

impl PdfLoader {
    pub fn new(path: &str) -> Self {
        Self { path: path.to_string() }
    }
}

#[async_trait]
impl DocumentLoader for PdfLoader {
    async fn load(&self) -> Result<Vec<DocumentEmbeddings>, Box<dyn std::error::Error + Send + Sync>> {
        // Implementation here
    }
}
```

Additional considerations:
- Handle potential errors gracefully (e.g., file not found, parsing errors)
- Implement chunking strategies for large PDFs
- Ensure proper encoding handling for various PDF text encodings
- Add unit tests for the `PdfLoader`
- Update documentation to include usage examples

### Alternatives
1. **Use a different PDF parsing library**: We could use libraries like `pdf-extract` or `pdf-rs` instead of `lopdf`. However, `lopdf` seems to offer a good balance of features and performance.

2. **Implement PDF parsing from scratch**: This would give us more control but would be time-consuming and potentially error-prone.

3. **Use external tools**: We could use external command-line tools like `pdftotext` and call them from Rust. This would be simpler to implement but would add external dependencies and potential security risks.

4. **Defer PDF support to users**: We could provide a trait for document loading and let users implement PDF support themselves. This would be simpler for Rig but would push complexity to the users.

The proposed solution (using `lopdf`) was chosen because it provides a good balance of functionality, ease of implementation, and integration with Rust. It keeps the implementation within Rig, providing a cohesive experience for users without external dependencies.

Tachikoma000:
# Add PDF Loader to Document Loaders

This PR implements a PDF loader as part of the document loaders module in Rig. It allows users to easily load and process PDF documents for use in RAG systems and other document processing tasks.

## Changes

- Implemented `PdfLoader` struct in `src/document_loaders/pdf.rs`
- Added `PdfLoader` to the `document_loaders` module
- Implemented `DocumentLoader` trait for `PdfLoader`
- Used the `lopdf` crate for PDF parsing
- Added error handling for file operations and PDF parsing
- Updated `Cargo.toml` with the `lopdf` dependency
- Added unit tests for `PdfLoader`
- Updated documentation with `PdfLoader` usage examples

## Implementation Details

The `PdfLoader` uses the `lopdf` crate to parse PDF files and extract text content. It handles potential errors such as file not found or parsing errors. The extracted text is converted into `DocumentEmbeddings` for further processing in Rig.

## Testing

Unit tests have been added to ensure the `PdfLoader` correctly loads PDF files and handles various edge cases. The tests cover:

- Loading a valid PDF file
- Handling a non-existent file
- Processing a PDF with multiple pages
- Dealing with empty PDF files

## Documentation

Code files are commented and some docstrings added

## Related Issue

Closes #24 

## Checklist

- [x] Code follows the project's coding style
- [x] Tests have been added and all tests pass
- [x] Documentation has been updated
- [x] Commit messages are clear and descriptive
- [x] Changes have been reviewed for potential performance impacts

## Additional Notes

This implementation focuses on text extraction from PDFs. Future enhancements could include handling PDFs with complex layouts or embedded images.


0xMochan:
I haven't done a formal review but I noticed a couple of things when I looked over this!
- This PR adds a lot of dependencies. We probably want to throw this as a [`cargo feature`](https://doc.rust-lang.org/cargo/reference/features.html) on the main rig library so that not all users download these deps if they install the library.
  - Also, `tokio` likely shouldn't be one of the dependencies since the library can work with multiple async backends.
  - Perhaps another route would be for a new crate `rig-pdf` instead of keeping this in the main library would be more suitable.
- It seems like there is also a CSV loader but I don't really see that documented in the PR.
- There's some code in the `mod.rs` file which I think we are trying to avoid. It could be spun out into a `document.rs` file.
- There shouldn't be any `println` code in our library to avoid erroneous output when people use our library. It should be using the `tracing` module so that logging can be turned off and on at will.
- I think some of the `Error` typing could be simplified using `thiserror` types for the return values.

These are just some first thought comments, I think stopher will do a main review of this. Looking forward to trying this out locally!

cvauclair:
This is being (re-)implemented in #55

PR #33, opened 2024-09-23
Feat(vector store): lancedb

marieaurore123:
1. Adds lanceDB as a vector store in `rig-lancedb` crate
2. Modifies `VectorStoreIndex` trait to include a generic type `type SearchParams` which can be passed to the vector store query to customize the query.


cvauclair:
@0xMochan @garance-buricatu would like your opinion on something here. This PR adds an associated type for the `VectorStoreIndex` trait `SearchParams` and adds it as an argument to the trait method, e.g.:
```rust
// in rig-core/src/vector_store/mod.rs
    fn top_n_from_query(
        &self,
        query: &str,
        n: usize,
        search_params: Self::SearchParams,   <----- this new arg
    ) -> ...;
```
However, a downstream effect is that you now need to pass an `VectorStoreIndex::SearchParams` argument when constructing a RAG agent, e.g.:
```rust
// in rig-core/examples/rag.rs
    let rag_agent = openai_client.agent("gpt-4")
        .preamble("
            You are a dictionary assistant here to assist the user in understanding the meaning of words.
            You will find additional non-standard word definitions that could be useful below.
        ")
        .dynamic_context(1, index, "".to_string())  <---- new empty string argument
        .build();
```

This feels a little awkward, even though it "makes sense". However, another approach would be to push whatever search/index parameters into the struct that implements the `VectorStoreIndex` trait. So instead of passing those params as an argument to the search functions, they would be fixed for a given `VectorStoreIndex`. E.g. you'd have this instead:
```rust
pub struct LanceDbVectorStore<M: EmbeddingModel> {
    model: M,
    document_table: lancedb::Table,
    embedding_table: lancedb::Table,
    params: ....   // <------- params would go inside the index object
}
```

This also has the advantage of making the `VectorStoreIndex` object safe avoid the need for the serde "hack" to get around the object safety requirements.

0xMochan:
Interesting, for context, can you provide an example for how the search params can be used? Is it reasonable to not have this change?

garance-buricatu:
@cvauclair 

Pro for new approach:
- wouldn't force the `SearchParams` struct to be deserializable/serializable, this has been annoying.
- improves readability like in the example you showed where the empty string was confusing
Con for new approach:
- can't use different settings of SearchParams for different calls to `top_n_from_query` or `top_n_from_embedding`. (unless we modify the `SearchParams` struct)

garance-buricatu:
@0xMochan 

in lanceDB, search params can be used to define the distance type (cosine, l2, dot, ...), search type (ann, enn), filters, and more fine tuning options on the search index.

cvauclair:
> Con for new approach:
> 
>     * can't use different settings of SearchParams for different calls to `top_n_from_query` or `top_n_from_embedding`. (unless we can modify the entire `LanceDbVectorStore` struct)

I think in the context of vector search, this is fine since these settings will most likely be static once the user has found the configuration that works for their RAG system. Plus, although these methods are public, users would usually interact with `VectorStoreIndex` types only insofar as they would create one and give it as a parameter to their `Agent`. Even in the current implementation, the `SearchParams` of an agent's dynamic context sources are static, so we'd just make it static across the board (if that makes sense).

0xMochan:
> > Con for new approach:
> > ```
> > - can't use different settings of SearchParams for different calls to `top_n_from_query` or `top_n_from_embedding`. (unless we can modify the entire `LanceDbVectorStore` struct)
> > ```
> 
> I think in the context of vector search, this is fine since these settings will most likely be static once the user has found the configuration that works for their RAG system. Plus, although these methods are public, users would usually interact with `VectorStoreIndex` types only insofar as they would create one and give it as a parameter to their `Agent`. Even in the current implementation, the `SearchParams` of an agent's dynamic context sources are static, so we'd just make it static across the board (if that makes sense).

I do think it _might_ be useful to change it on the fly, or if the user of an application wants to adjust this based on a UI / TUI. But you can always reconstruct an agent so it may not be that big of a deal. There can always be an extra search command to use a specific method of querying that overrides the default (so you choose a default when you create the `Agent`) but when you query, there's the ability to override that by using a specific function `top_n_from_query_using_search_param` (bad naming but ya know). This can be an extra function defined by a trait for this feature (theoretically) and can always be added in the future if there's a demand for it (perhaps we leave a comment about it?)

rig-core/src/providers/openai.rs
```diff
@@ -205,12 +208,60 @@ enum ApiResponse<T> {
...
-/// `text-embedding-3-small` embedding model
-pub const TEXT_EMBEDDING_3_SMALL: &str = "text-embedding-3-small";
-/// `text-embedding-ada-002` embedding model
-pub const TEXT_EMBEDDING_ADA_002: &str = "text-embedding-ada-002";
+#[derive(Debug, Clone)]
+pub enum OpenAIEmbeddingModel {
+    TextEmbedding3Large,
+    TextEmbedding3Small,
+    TextEmbeddingAda002,
+}
```
cvauclair, line 216: Keep as `const &str` and maybe match on string to get the embedding vector dimension?

The idea is to allow users to specify a model for which we don't have consts/enums for (e.g.: a model that was just release and that we did not have time to integrate yet)

cvauclair (changes requested):

.github/workflows/ci.yaml
```diff
@@ -49,5 +49,8 @@ jobs:
         with:
           components: clippy
 
+      - name: Install Protoc
+        uses: arduino/setup-protoc@v3
```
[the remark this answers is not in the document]
garance-buricatu: image

rig-core/src/providers/cohere.rs
```diff
@@ -63,7 +63,23 @@ impl Client {
...
+            EMBED_ENGLISH_V3 | EMBED_MULTILINGUAL_V3 | EMBED_ENGLISH_LIGHT_V2 => 1024,
+            EMBED_ENGLISH_LIGHT_V3 | EMBED_MULTILINGUAL_LIGHT_V3 => 384,
+            EMBED_ENGLISH_V2 => 4096,
+            EMBED_MULTILINGUAL_V2 => 768,
+            _ => 0,
+        };
+        EmbeddingModel::new(self.clone(), model, input_type, ndims)
+    }
+
+    pub fn embedding_model_with_ndims(
```
cvauclair, line 76: Add docstrings

rig-core/src/providers/cohere.rs
```diff
@@ -63,7 +63,23 @@ impl Client {
...
 
     pub fn embedding_model(&self, model: &str, input_type: &str) -> EmbeddingModel {
-        EmbeddingModel::new(self.clone(), model, input_type)
+        let ndims = match model {
+            EMBED_ENGLISH_V3 | EMBED_MULTILINGUAL_V3 | EMBED_ENGLISH_LIGHT_V2 => 1024,
+            EMBED_ENGLISH_LIGHT_V3 | EMBED_MULTILINGUAL_LIGHT_V3 => 384,
+            EMBED_ENGLISH_V2 => 4096,
+            EMBED_MULTILINGUAL_V2 => 768,
+            _ => 0,
+        };
```
cvauclair, line 72: Add dosctrings that mention a default ndims will be used if the model doesnt match one of the enum values

rig-core/src/providers/openai.rs
```diff
@@ -80,7 +80,27 @@ impl Client {
     /// let embedding_model = openai.embedding_model(openai::TEXT_EMBEDDING_3_LARGE);
     /// ```
     pub fn embedding_model(&self, model: &str) -> EmbeddingModel {
-        EmbeddingModel::new(self.clone(), model)
+        let ndims = match model {
+            TEXT_EMBEDDING_3_LARGE => 3072,
+            TEXT_EMBEDDING_3_SMALL | TEXT_EMBEDDING_ADA_002 => 1536,
+            _ => 0,
+        };
```
cvauclair, line 87: Add information about default ndims to docstring

rig-core/src/providers/openai.rs
```diff
@@ -80,7 +80,27 @@ impl Client {
...
+        EmbeddingModel::new(self.clone(), model, ndims)
+    }
+
+    /// Create an embedding model with the given name.
```
cvauclair, line 91: Update docstrings "[...] given name and dimension"

rig-core/src/vector_store/mod.rs
```diff
@@ -183,15 +183,15 @@ impl<I: VectorStoreIndex> VectorStoreIndexDyn for I {
         query: &'a str,
         n: usize,
     ) -> BoxFuture<'a, Result<Vec<(f64, DocumentEmbeddings)>, VectorStoreError>> {
-        Box::pin(self.top_n_from_query(query, n))
+        Box::pin(async move { self.top_n_from_query(query, n).await })
```
cvauclair, line 186: Why was this necessary?

rig-core/src/embeddings.rs
```diff
@@ -66,13 +66,20 @@ pub enum EmbeddingError {
     /// Error returned by the embedding model provider
     #[error("ProviderError: {0}")]
     ProviderError(String),
+
+    /// Http error (e.g.: connection error, timeout, etc.)
```
cvauclair, line 70: Edit docstring (also, where is this being used?)

rig-mongodb/examples/vector_search_mongodb.rs
```diff
@@ -32,7 +32,7 @@ async fn main() -> Result<(), anyhow::Error> {
     let mut vector_store = MongoDbVectorStore::new(collection);
 
     // Select the embedding model and generate our embeddings
-    let model = openai_client.embedding_model("text-embedding-ada-002");
+    let model = openai_client.embedding_model(&OpenAIEmbeddingModel::TextEmbeddingAda002);
```
cvauclair: This is no longer valid

rig-mongodb/examples/vector_search_mongodb.rs
```diff
@@ -3,10 +3,10 @@ use std::env;
 
 use rig::{
     embeddings::{DocumentEmbeddings, EmbeddingsBuilder},
-    providers::openai::Client,
+    providers::openai::{Client, OpenAIEmbeddingModel},
```
cvauclair, line 6: Import no longer valid

rig-mongodb/src/lib.rs
```diff
@@ -104,23 +102,67 @@ pub struct MongoDbVectorIndex<M: EmbeddingModel> {
...
             index_name: index_name.to_string(),
-            filter,
+            search_params,
+        }
+    }
+}
+
+/// See [MongoDB Vector Search](https://www.mongodb.com/docs/atlas/atlas-vector-search/vector-search-stage/) for more information
+/// on each of the fields
+#[derive(Deserialize)]
```
cvauclair, line 126: Un-needed `Deserialize`

rig-lancedb/src/lib.rs
```diff
@@ -0,0 +1,350 @@
...
+    document_table: lancedb::Table,
+    /// Table containing embeddings only.
+    /// Foreign key references the document in document table.
+    embedding_table: lancedb::Table,
+    /// Vector search params that are used during vector search operations.
+    search_params: SearchParams,
+}
+
+/// See [LanceDB vector search](https://lancedb.github.io/lancedb/search/) for more information.
+#[derive(Deserialize, Serialize, Debug, Clone)]
```
cvauclair, line 41: Unneeded `Deserialize` and `Serialize`

rig-lancedb/src/lib.rs
```diff
@@ -0,0 +1,350 @@
...
+pub enum SearchType {
+    // Flat search, also called ENN or kNN.
+    Flat,
+    /// Approximal Nearest Neighbor search, also called ANN.
+    Approximate,
+}
+
+#[derive(Deserialize, Serialize, Debug, Clone, Default)]
```
cvauclair, line 49: Unneeded `Deserialize` and `Serialize`

rig-lancedb/src/lib.rs
```diff
@@ -0,0 +1,350 @@
...
+            .execute()
+            .await?;
+
+        let embedding_table = db
+            .create_empty_table(
+                "embeddings",
+                Arc::new(Self::embedding_schema(model.ndims() as i32)),
+            )
+            .execute()
+            .await?;
```
cvauclair, line 115: Either make the tables arguments of the constructor, or add a check to see if they exist. Rn, this will fail if the tables already exist (e.g.: when running the same example twice in a row)

rig-lancedb/src/lib.rs
```diff
@@ -0,0 +1,350 @@
...
+        let SearchParams {
+            distance_type,
+            search_type,
+            nprobes,
+            refine_factor,
+            post_filter,
+        } = self.search_params.clone();
+
+        if let Some(distance_type) = distance_type {
+            query.clone().distance_type(distance_type);
```
cvauclair, line 301: This doesn't actually set anything since the `distance_type` method is invoked on a _clone_ of `query`. 

Make `query` mutable and change all the `query.clone().foo()` to `query = query.foo()`

cvauclair (changes requested):
Haven't reviewed everything yet, but there are a lot of things to change so will make it in two parts

rig-lancedb/src/lib.rs
```diff
@@ -0,0 +1,300 @@
...
+            .map(|value| {
+                Ok((
+                    match value.get("_distance") {
+                        Some(Value::Number(distance)) => distance.as_f64().unwrap_or_default(),
+                        _ => 0.0,
+                    },
+                    match value.get(self.id_field.clone()) {
+                        Some(Value::String(id)) => id.to_string(),
+                        _ => "".to_string(),
+                    },
```
cvauclair, line 260: Duplicate ids will cause an error for certain providers (e.g.: cohere) so the default returned here should not be the empty string, but something unique.

Suggestion: call enumerate before this map and use the value `unknown{n}` as the default id value

rig-lancedb/src/utils/deserializer.rs
```diff
@@ -0,0 +1,1098 @@
...
+    {
+        self.to_primitive::<T>()
+            .iter()
+            .map(serde_json::to_value)
+            .collect()
+    }
+}
+
+/// Trait used to "deserialize" an arrow_array::Array as as list of str objects.
+pub trait DeserializeByteArray {
```
cvauclair, line 395: Does this trait need to be public?

rig-lancedb/src/utils/deserializer.rs
```diff
@@ -0,0 +1,1098 @@
...
+                        acc
+                    })
+            })
+            .map(Value::Object)
+            .collect())
+    }
+}
+
+/// Trait used to "deserialize" an arrow_array::Array as as list of primitive objects.
+pub trait DeserializePrimitiveArray {
```
cvauclair, line 363: Does this trait need to be public?

rig-lancedb/src/utils/deserializer.rs
```diff
@@ -0,0 +1,1098 @@
...
+
+use crate::serde_to_rig_error;
+
+fn arrow_to_rig_error(e: ArrowError) -> VectorStoreError {
+    VectorStoreError::DatastoreError(Box::new(e))
+}
+
+/// Trait used to deserialize data returned from LanceDB queries into a serde_json::Value vector.
+/// Data returned by LanceDB is a vector of `RecordBatch` items.
+pub trait RecordBatchDeserializer {
```
cvauclair, line 30: Does this trait need to be public?

cvauclair (changes requested):

rig-lancedb/src/lib.rs
```diff
@@ -0,0 +1,300 @@
...
+            search_type,
+            nprobes,
+            refine_factor,
+            post_filter,
+            column,
+        } = self.search_params.clone();
+
+        if let Some(distance_type) = distance_type {
+            query = query.distance_type(distance_type);
+        }
```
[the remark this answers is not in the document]
garance-buricatu: No, I updated the doc strings to answer questions about defaults.

rig-lancedb/src/lib.rs
```diff
@@ -0,0 +1,300 @@
...
+
+        if let Some(SearchType::Approximate) = search_type {
+            if let Some(nprobes) = nprobes {
+                query = query.nprobes(nprobes);
+            }
+            if let Some(refine_factor) = refine_factor {
+                query = query.refine_factor(refine_factor);
+            }
+        }
```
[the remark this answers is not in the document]
garance-buricatu: I didn't do match here because I don't need to handle `None` on any of these cases

rig-lancedb/src/lib.rs
```diff
@@ -0,0 +1,300 @@
...
+
+        if let Some(true) = post_filter {
+            query = query.postfilter();
+        }
+
+        if let Some(column) = column {
+            query = query.column(column.as_str())
+        }
```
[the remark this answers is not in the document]
garance-buricatu: See above

rig-lancedb/src/utils/mod.rs
```diff
@@ -0,0 +1,28 @@
...
+use deserializer::RecordBatchDeserializer;
+use futures::TryStreamExt;
+use lancedb::query::ExecutableQuery;
+use rig::vector_store::VectorStoreError;
+
+use crate::lancedb_to_rig_error;
+
+/// Trait that facilitates the conversion of columnar data returned by a lanceDb query to serde_json::Value.
+/// Used whenever a lanceDb table is queried.
+pub trait Query {
```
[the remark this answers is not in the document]
garance-buricatu: It's only public in the crate (in `lib.rs`, I define `mod utils` instead of `pub mod utils`). But yeah, it needs to be public in the crate bc it is accessed from outside that file

rig-lancedb/src/utils/deserializer.rs
garance-buricatu: yeah this one is used outside the file

rig-lancedb/examples/vector_search_local_ann.rs
cvauclair, line 1: Remove AI generated data in favor of repeated hardcoded data

rig-lancedb/examples/vector_search_s3_ann.rs
cvauclair, line 1: Remove AI generated data in favor of repeated hardcoded data

cvauclair (changes requested):

cvauclair (approved):

Issue #41, opened 2024-09-30
Refactor vector store API


cvauclair:
The `VectorStoreIndex` trait should be simplified to this:

```rust
/// Gets the top `n` items from vector store that are the most relevant to the `query`.
/// Includes each item's relevance score.
pub trait VectorStoreIndex: Send + Sync {
    /// Return the top n relevant items from the vector store as a vector of deserializable objects.
    async fn top_n<T: for<'a> Deserialize<'a>>(
        &self,
        query: &str,
        n: usize,
    ) -> Result<Vec<(f64, T)>, VectorStoreError>;

    /// Same as `top_n_from_query` but returns the ids only.
    async fn top_n_ids(
        &self,
        query: &str,
        n: usize,
    ) -> Result<Vec<(f64, String)>, VectorStoreError>;
}
```
**Case 1:** `Self` defines an embedding model.
* Generate the vector embedding based on the input `query` and `self.model`
* Invoke vector search
* This **must** apply to MongoDB, Pinecone. Ie. `model` field of `MongoDBStore` must be `impl EmbeddingModel`
* This **could** apply to LanceDB and Weaviate. Ie. `model` field of `LanceDBStore` must be `Option<impl EmbeddingModel>`

**Case 2:** `Self` does not define an embedding model (`model` field of `Self` is `None`).
* Weaviate - user must have defined `vectorizer_config` on the weaviate collection beforehand. Invoke vector search with string.
* LanceDB - user must define embedding function in `Self`. Use the embedding function to generate the embedding for `query`. Invoke vector search with embedding.

cvauclair:
Closed by #42

PR #42, opened 2024-09-30
refactor: Update VectorStoreIndex trait

marieaurore123:
Github issue: https://github.com/0xPlaygrounds/rig/issues/41

* Refactor `VectorStoreIndex` trait

**Notes:**
* kept `DocumentEmbeddings` as the `InMemoryVectorStore` value type. Did this because if I change it to a more general type, I would need to change the `VectorStore` trait which should be done in a separate branch.
* method `top_n_ids` of `VectorStoreIndex` contains a generic type: `<T: for<'a> Deserialize<'a>`. Technically, it we don't need type `T` to get the `ids` of the top n documents. However, I found it simpler to get the `ids` of the top n by calling the existing method `top_n` which needs to be called like this `top_n::<T>()` and filtering the ids only. This also allows the implementation of the method `top_n_ids` to be to same for all trait implementations (ie. implementation defined in trait def)

Please let me know what you think about the second point! Thanks:)

### DIFF
--- a/rig-core/src/vector_store/mod.rs
+++ b/rig-core/src/vector_store/mod.rs
@@ -52,29 +52,19 @@ pub trait VectorStore: Send + Sync {
 /// Trait for vector store indexes
 pub trait VectorStoreIndex: Send + Sync {
     /// Get the top n documents based on the distance to the given query.
-    /// The result is a list of tuples with the distance and the document.
+    /// The result is a list of tuples of the form (score, id, document)
     fn top_n<T: for<'a> Deserialize<'a> + std::marker::Send>(
         &self,
         query: &str,
         n: usize,
     ) -> impl std::future::Future<Output = Result<Vec<(f64, String, T)>, VectorStoreError>> + Send;
 
-    /// Same as `top_n_from_query` but returns the document ids only.
-    fn top_n_ids<T: for<'a> Deserialize<'a> + std::marker::Send>(
+    /// Same as `top_n` but returns the document ids only.
+    fn top_n_ids(
         &self,
         query: &str,
         n: usize,
-    ) -> impl std::future::Future<Output = Result<Vec<(f64, String)>, VectorStoreError>> + Send
-    {
-        async move {
-            Ok(self
-                .top_n::<T>(query, n)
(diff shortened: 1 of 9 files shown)


0xMochan (approved):
Seems like a lot of code got simplified 👍

rig-core/src/vector_store/mod.rs
```diff
@@ -50,167 +51,78 @@ pub trait VectorStore: Send + Sync {
...
-            let documents = self.top_n_from_embedding(prompt_embedding, n).await?;
-            Ok(documents
+            Ok(self
+                .top_n::<T>(query, n)
+                .await?
                 .into_iter()
-                .map(|(distance, doc)| (distance, doc.id))
+                .map(|(distance, id, _)| (distance, id))
                 .collect())
         }
```
cvauclair, line 76: Remove the provided implementation of `top_n_ids` since this can vary based on the DB

cvauclair (changes requested):

cvauclair:
@garance-buricatu for the same of record keeping, here are my opinions on your points:
> kept DocumentEmbeddings as the InMemoryVectorStore value type. Did this because if I change it to a more general type, I would need to change the VectorStore trait which should be done in a separate branch.

That's fine for now, we'll remove the `DocumentEmbeddings` type in a subsequent PR (or maybe keep it only for the internal representation of the `InMemoryVectorStore`)

> method top_n_ids of VectorStoreIndex contains a generic type: <T: for<'a> Deserialize<'a>. Technically, it we don't need type T to get the ids of the top n documents. However, I found it simpler to get the ids of the top n by calling the existing method top_n which needs to be called like this top_n::<T>() and filtering the ids only. This also allows the implementation of the method top_n_ids to be to same for all trait implementations (ie. implementation defined in trait def)

IMO we should remove the provided implementation since 1) some DBs might have a more optimized way to query the ids only (e.g.: `SELECT id FROM ...` instead of `SELECT * FROM ...`); and 2) we don't know the name of the primary key used by the documents in advance.

cvauclair (approved):

PR #43, opened 2024-10-01
fix: move context documents to user prompt message

0xMochan:
Context documents get prepended to the prompt user message rather than to system messages.

### DIFF
--- a/rig-core/src/completion.rs
+++ b/rig-core/src/completion.rs
@@ -120,6 +120,28 @@ pub struct Document {
     pub additional_props: HashMap<String, String>,
 }
 
+impl std::fmt::Display for Document {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let metadata = self
+            .additional_props
+            .iter()
+            .map(|(k, v)| format!("{}: {}", k, v))
+            .collect::<Vec<_>>()
+            .join(" ");
+
+        write!(
+            f,
+            concat!("<file id: {}>\n", "{}\n", "</file>\n"),
+            self.id,
+            if self.additional_props.is_empty() {
+                self.text.clone()
+            } else {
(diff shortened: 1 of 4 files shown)


rig-core/src/completion.rs
cvauclair, line 1: Add a test or two for the `Display` implementation of `Document`

rig-core/src/providers/perplexity.rs
```diff
@@ -227,25 +227,28 @@ impl completion::CompletionModel for CompletionModel {
...
+                    .documents
+                    .iter()
+                    .map(|doc| doc.to_string())
+                    .collect::<Vec<_>>()
+                    .join("\n"),
+                completion_request.prompt
+            )
+        } else {
+            completion_request.prompt
+        };
```
cvauclair, line 243: Since this is copied in many provider implementation, this can probably be refactored into it's own utility method `CompletionRequest::prompt_with_context(&self) -> String`

cvauclair (changes requested):

rig-core/src/completion.rs
```diff
@@ -432,3 +472,78 @@ impl<M: CompletionModel> CompletionRequestBuilder<M> {
...
+        };
+
+        let expected = concat!(
+            "<file id: 123>\n",
+            "<metadata author: \"John Doe\" length: \"42\" />\n",
+            "This is a test document.\n",
+            "</file>\n"
+        );
+        assert_eq!(format!("{}", doc), expected);
+    }
```
cvauclair, line 511: This test is not deterministic since the ordering of KVs in `additional_props` is not deterministic. 
Either change the test to compare the two scenarios, or change the `Display` implementation to get a deterministic output (e.g.: KVs ordered alphabetically)

cvauclair (changes requested):
Looking good! Just need to fix one of the tests

cvauclair (approved):

PR #48, opened 2024-10-02
devops: Add nextest test runner to CI

cvauclair:


0xMochan (approved):
nextrunner looks cool!

PR #53, opened 2024-10-09
refactor: remove DocumentEmbeddings from in memory vector store

marieaurore123:
1. Refactor in memory vector store
2. Remove `VectorStore` trait
3. All other changes are side-effects of this

### DIFF
--- a/rig-core/examples/rag_dynamic_tools.rs
+++ b/rig-core/examples/rag_dynamic_tools.rs
@@ -1,10 +1,10 @@
 use anyhow::Result;
 use rig::{
     completion::{Prompt, ToolDefinition},
-    embeddings::EmbeddingsBuilder,
+    embeddings::{DocumentEmbeddings, EmbeddingsBuilder},
     providers::openai::{Client, TEXT_EMBEDDING_ADA_002},
     tool::{Tool, ToolEmbedding, ToolSet},
-    vector_store::{in_memory_store::InMemoryVectorStore, VectorStore},
+    vector_store::in_memory_store::InMemoryVectorStore,
 };
 use serde::{Deserialize, Serialize};
 use serde_json::json;
@@ -150,9 +150,6 @@ async fn main() -> Result<(), anyhow::Error> {
 
     let embedding_model = openai_client.embedding_model(TEXT_EMBEDDING_ADA_002);
 
-    // Create vector store, compute tool embeddings and load them in the store
-    let mut vector_store = InMemoryVectorStore::default();
-
     let toolset = ToolSet::builder()
         .dynamic_tool(Add)
         .dynamic_tool(Subtract)
@@ -163,7 +160,18 @@ async fn main() -> Result<(), anyhow::Error> {
         .build()
         .await?;
 
-    vector_store.add_documents(embeddings).await?;
+    let vector_store = InMemoryVectorStore::default().add_documents(
+        embeddings
+            .into_iter()
+            .map(
+                |DocumentEmbeddings {
+                     id,
+                     document,
+                     embeddings,
+                 }| { (id, document, embeddings) },
+            )
+            .collect(),
+    )?;
 
     // Create vector store index
     let index = vector_store.index(embedding_model);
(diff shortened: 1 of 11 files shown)


rig-core/examples/rag_dynamic_tools.rs
```diff
@@ -163,10 +160,20 @@ async fn main() -> Result<(), anyhow::Error> {
...
+                .map(
+                    |DocumentEmbeddings {
+                         id,
+                         document,
+                         embeddings,
+                     }| { (id, document, embeddings) },
+                )
+                .collect(),
+        )?
+        .index(embedding_model);
```
cvauclair, line 176: This is not ideal, but would presumably get fixed in subsequent PR

rig-core/src/vector_store/in_memory_store.rs
cvauclair, line 1: In this entire file, rename type parameter `D` to `T` to follow conventions (using other type parameter names is warranted when the type parameter has strict bounds, e.g.: `M: CompletionModel`, but in this case it's generic enough that `T` is more appropriate.

rig-core/src/vector_store/in_memory_store.rs
```diff
@@ -7,28 +7,29 @@ use std::{
...
+    /// Hashmap value is a tuple of the serializable document and its corresponding embeddings.
+    embeddings: HashMap<String, (D, Vec<Embedding>)>,
 }
 
-impl InMemoryVectorStore {
+impl<D: Serialize + Eq> InMemoryVectorStore<D> {
     /// Implement vector search on InMemoryVectorStore.
     /// To be used by implementations of top_n and top_n_ids methods on VectorStoreIndex trait for InMemoryVectorStore.
-    fn vector_search(&self, prompt_embedding: &Embedding, n: usize) -> EmbeddingRanking {
+    fn vector_search(&self, prompt_embedding: &Embedding, n: usize) -> EmbeddingRanking<D> {
```
cvauclair, line 26: Add unit tests for `vector_search` (at least one with multiple embeddings per item). 

Populate the store with fake embeddings with very few dimensions but make sure that are sensible. I.e.: Make the matching vectors close to each other (e.g.: `[0.1, 0.1, 0.5]` and `[0.1, 0.1, 0.6]`) and non-matching vectors away from each other (e.g.: `[0.1, 0.1, 0.5]` and `[-1.0, -0.6, 0.1]`)

cvauclair (changes requested):

rig-core/src/vector_store/in_memory_store.rs
```diff
@@ -56,84 +52,69 @@ impl InMemoryVectorStore {
         tracing::info!(target: "rig",
             "Selected documents: {}",
             docs.iter()
-                .map(|Reverse(RankingItem(distance, id, _, _))| format!("{} ({})", id, distance))
+                .map(|Reverse(RankingItem(distance, id, _, embed_doc))| format!("{} ({}). Specific match: {}", id, distance, embed_doc))
```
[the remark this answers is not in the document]
garance-buricatu: Ok that makes sense but i though the whole point of having the document in the Embedding object was for debugging. So maybe i'll log it with tracing::debug! ?

rig-core/src/vector_store/in_memory_store.rs
garance-buricatu: But T is already in use by `top_n` and `get_document`

cvauclair: Ok, I think we'll change `VectorStoreIndex` to have a trait-level generic type anyway so this won't be a problem after that

rig-core/src/vector_store/in_memory_store.rs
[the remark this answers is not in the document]
cvauclair: The user can decide to log the full document after generating the embedding if they want to inspect the data, but Rig shouldnt pollute the logs with large data dumps ourselves

cvauclair (changes requested):

PR #56, opened 2024-10-11
feat(model-providers): 16 Add Gemini Completion and Embedding Models

### LINKED ISSUE #16: feat: Add support for Gemini completion and embedding models
## Feature Request
Add Gemini as a new model provider for Rig.
 
### Motivation
Gemini is a popular model provider that Rig should support and they have a REST API which should be relatively easy to integrate.

### Proposal
Create a new module `rig::providers::gemini` which implements completion and embedding model clients for Google's Gemini.

Use the existing provider integrations as a guide (see [`rig::providers::openai`](https://github.com/0xPlaygrounds/rig/blob/main/rig-core/src/providers/openai.rs) and [`rig::providers::cohere`](https://github.com/0xPlaygrounds/rig/blob/main/rig-core/src/providers/cohere.rs)).

You can find a collection of examples using the REST API (implemented in Python) [here](https://github.com/google-gemini/cookbook/tree/main/quickstarts/rest)

mateobelanger:
### 🎯 Description & Goal
This pull request adds support for the Google Gemini LLM provider. 

- Add impl. for `EmbeddingModel` and `CompletionModel` traits for Google Gemini API
- Update to the README files, 
- New Gemini-agent/embeddings example 

### DOCS Updates:
* [`README.md`](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L37-R37): Added Google Gemini to the list of supported LLM providers. [[1]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L37-R37) [[2]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L73-R73)
* [`rig-core/README.md`](diffhunk://#diff-1ba6f89723a0990d647034467c2ed5c4356818b9df914158e9274cf400e58bb9L14-R14): Updated to include Google Gemini in the list of LLM providers. [[1]](diffhunk://#diff-1ba6f89723a0990d647034467c2ed5c4356818b9df914158e9274cf400e58bb9L14-R14) [[2]](diffhunk://#diff-1ba6f89723a0990d647034467c2ed5c4356818b9df914158e9274cf400e58bb9L50-R50)
* [`rig-core/examples/gemini_agent.rs`](diffhunk://#diff-cc6f83ed1e7400ac0fa900837e018553a16bf1ee1e2ca6541d6a160587899176R1-R24): Added an example demonstrating how to use the Google Gemini client to create an agent and prompt it.

### Gemini Models Integration:
* [`rig-core/src/providers/gemini/client.rs`](diffhunk://#diff-9576c852df4ce35061650477d4b260d04046eab6b0980359d61bcf3465b79d3cR1-R152): Implemented the Google Gemini client with methods for creating embedding models, completion models, agents, and extractors.
* [`rig-core/src/providers/gemini/completion.rs`](diffhunk://#diff-c4e2b5b27ea89b3e1f6c4b97305607d7662a42ec521ec2be842e33c25c0c9c31R1-R387): Added support for Google Gemini completion models and defined the necessary request and response structures.
* [`rig-core/src/providers/gemini/embedding.rs`](diffhunk://#diff-7e033028f58eb540f0baeb2b25666047fac8ee5c564db58214dab4265e29fd61R1-R97): Implemented the embedding model for Google Gemini, including methods for embedding documents.

> [!note] 
> -  Anthropic & OpenAi implementation of the CompletionModel trait differ `async fn completion()`. In OpenAI we add the preamble but not with Anthropic


rig-core/src/providers/gemini/completion.rs
```diff
@@ -0,0 +1,607 @@
...
+}
+
+impl completion::CompletionModel for CompletionModel {
+    type Response = GenerateContentResponse;
+
+    async fn completion(
+        &self,
+        mut completion_request: CompletionRequest,
+    ) -> Result<completion::CompletionResponse<GenerateContentResponse>, CompletionError> {
+        // QUESTION: Why do Anthropic/openAi implementation differ here? OpenAI adds the preamble but Anthropic does not.
```
0xMochan, line 417: For Anthropic, the Preamble is set as the `system` API field. The other ones differ add it as a `system` message in the chat history. Differences in API design.

rig-core/src/providers/gemini/completion.rs
```diff
@@ -0,0 +1,607 @@
...
+        let prompt_with_context = completion_request.prompt_with_context();
+
+        full_history.push(completion::Message {
+            role: "user".into(),
+            content: prompt_with_context,
+        });
+
+        // Handle Gemini specific parameters
+        let mut generation_config =
+            GenerationConfig::try_from(completion_request.additional_params.unwrap_or_default())?;
```
0xMochan, line 439: This is interesting. One negative of this is that if Gemini adds new fields and rig isn't updated, you won't be able to use them since this struct will ignore them. I chose not to do this for Anthropic.

rig-core/src/providers/gemini/completion.rs
```diff
@@ -0,0 +1,607 @@
...
+
+        // Set temperature from completion_request or additional_params
+        if let Some(temp) = completion_request.temperature {
+            generation_config.temperature = Some(temp);
+        }
+
+        // Set max_tokens from completion_request or additional_params
+        if let Some(max_tokens) = completion_request.max_tokens {
+            generation_config.max_output_tokens = Some(max_tokens);
+        }
```
0xMochan, line 449: Why can you set these extra parameters in two locations?

0xMochan (approved):
Overall approved, but there's a consequence I wrote about related to parsing the `additional_parameters` into an explicit requests Model that is significant to think about. You might want to collect additional keys into it's own field and then merge that onto the reuqest after the fact (might be good to do this for the other providers in a later PR).

rig-core/src/providers/gemini/completion.rs
mateobelanger: Because it's both a Rig (CompletionRequest) optionnal param and a  Gemini GenerationConfig field.

So depending if the user fills `temperature` or `max_output_token` I need to check both locations.

rig-core/src/providers/gemini/completion.rs
mateobelanger: @cvauclair What do you think? I thought that from a dev UX perspective having access to types/docs definitions was useful. IMO field updates/additions should be *relatively* infrequent so that we can wait for someone to raise a repo issue to implement the change.

rig-core/src/providers/gemini/completion.rs
mateobelanger: Made the change! This is `system_instruction` in the request object.

README.md
```diff
@@ -34,7 +34,7 @@ We'd love your feedback. Please take a moment to let us know what you think usin
 
 ## High-level features
 - Full support for LLM completion and embedding workflows
-- Simple but powerful common abstractions over LLM providers (e.g. OpenAI, Cohere) and vector stores (e.g. MongoDB, in-memory)
+- Simple but powerful common abstractions over LLM providers (e.g. OpenAI, Cohere, Google Gemini) and vector stores (e.g. MongoDB, in-memory)
```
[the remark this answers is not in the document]
mateobelanger: fair!

rig-core/src/providers/gemini/completion.rs
0xMochan: This isn't a case of rare and breaking changes, new fields being added would _not_ warrant a new version of the API (it would just be added) yet they would cause this model to break. It's not that infrequent as the providers match each other's capabilities (see OpenAI prompt caching) and it also forces users to wait for a rig update unable to use the newest features of their provider.

This might be a fine compromise, but I want to make it clear that this is the consequence.

rig-core/src/providers/gemini/completion.rs
0xMochan: It's a little odd that it's in both, this is not how it's down in the other providers. Is this a suggestion for changing the internal DX for new providers?

rig-core/src/providers/gemini/completion.rs
```diff
@@ -0,0 +1,649 @@
...
+                raw_response: response,
+            }),
+            _ => Err(CompletionError::ResponseError(
+                "No candidates found in response".into(),
+            )),
+        }
+    }
+}
+
+impl TryFrom<serde_json::Value> for GenerationConfig {
```
garance-buricatu, line 521: You may not need this `TryFrom` implementation. Instead of doing `GenerationConfig::try_from(completion_request.additional_params.unwrap_or_default())?;`, you can do `serde_json::from_value::<GenerationConfig>(completion_request.additional_params.unwrap_or_default())?`. Since `GenerationConfig` has Option for each field, if the additional params don't contain the generation_config field, it won't fail. 

This won't work though if field X of `completion_request.additional_params` is a string but field X of GenerationConfig expects a float for example.

marieaurore123 (changes requested):
I added a few comments!

rig-core/src/providers/gemini/completion.rs
cvauclair: Yeah the Rig hardcoded set of parameters should be favored over whatever is in `additional_params`

rig-core/src/providers/gemini/completion.rs
```diff
@@ -0,0 +1,649 @@
...
+            frequency_penalty: None,
+            response_logprobs: None,
+            logprobs: None,
+        };
+
+        fn unexpected_type_error(field: &str) -> CompletionError {
+            CompletionError::ResponseError(format!("Unexpected type for field '{}'", field))
+        }
+
+        if let Some(obj) = value.as_object() {
```
cvauclair, line 544: This check fails if `additional_params` is not set, even though the `GenerationConfig` is not required to query the API.

cvauclair (changes requested):
Good first draft but couple things to change (especially the bug which makes `GenerationConfig` required)

rig-core/examples/gemini_agent.rs
```diff
@@ -0,0 +1,35 @@
...
+        .agent(gemini::completion::GEMINI_1_5_PRO)
+        .preamble("Be creative and concise. Answer directly and clearly.")
+        .temperature(0.5)
+        .max_tokens(8192)
+        .additional_params(serde_json::to_value(GenerationConfig {
+            top_k: Some(1),
+            top_p: Some(0.95),
+            candidate_count: Some(1),
+            ..Default::default()
+        })?) // Unwrap the Result to get the Value
```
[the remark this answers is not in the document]
mateobelanger: Yes! Keeping the params docs was what I had in mind.

rig-core/src/providers/gemini/completion.rs
```diff
@@ -0,0 +1,798 @@
...
+        let prompt_with_context = completion_request.prompt_with_context();
+
+        full_history.push(completion::Message {
+            role: "user".into(),
+            content: prompt_with_context,
+        });
+
+        // Handle Gemini specific parameters
+        let mut generation_config =
+            GenerationConfig::try_from(completion_request.additional_params.unwrap_or_default())?;
```
cvauclair, line 63: This like here causes requests to fail if the user does not set `additional_params`. 

FYI the default value for `serde_json::Value` is `serde_json::Value::Null` and not an empty `serde_json::Value::Object`, so if the user does not set `additional_params`, this code will set it to `null` before calling `GenerationConfig::try_from` which will fail.

Two possible fixes:
1. Use `unwrap_or(serde_json::Value::Object({}))` instead of `unwrap_or_default()`
2. Only call `GenerationConfig::try_from` if `additional_params` is not `None`

cvauclair (changes requested):
The bug that cases requests to fail if `additional_params` is not set still exists!

cvauclair (approved):

cvauclair:
@garance-buricatu can you make sure your comments were addressed and approve? Would like to merge this before eod, thanks!

0xMochan (approved):

rig-core/src/providers/gemini/completion.rs
garance-buricatu: @mateobelanger is there a reason why you didn't change this?

marieaurore123 (approved):

PR #59, opened 2024-10-15
Feat(embeddings): Embeddable derive macro

marieaurore123:
**TODO**: RENAME `Embeddable` trait to `ToEmbedabble`, `ToEmbedTargets`, `Extract`, `ExtractEmbeddable`, ...

This PR accomplishes:
* Defines `Embeddable` trait which selects the portions of a struct that need ot be embedded. The `embeddable()` method returns a `OneOrMany` object which returns either one or many items to embed. 
* Write custom macro that derives the `Embeddable` trait for any user-defined object


rig-core/src/embeddings/builder.rs
```diff
@@ -0,0 +1,209 @@
...
+        for doc in documents.into_iter() {
+            let embed_targets = doc.embeddable()?;
+
+            self.documents.push((doc, embed_targets));
+        }
+
+        Ok(self)
+    }
+}
+
```
garance-buricatu, line 111: I don't like adding Clone here but I couldn't find a way around this.

rig-core/src/embeddings/tool.rs
```diff
@@ -0,0 +1,25 @@
+use crate::{self as rig, tool::ToolEmbeddingDyn};
```
garance-buricatu, line 1: The only way I  found to add tools to the embeddings builder was by converting all tools in a `Toolset` to the type `EmbeddableTool` which implements `Embeddable`.

Tool items defined in `Toolset` can all be of different types due to dynamic dispatch. However, I haven't implemented dynamic dispatch for items in the embeddings builder. Therefore, all documents added to the embeddings builder need to be of the same type (that implements `Embeddable`). So before added tools from `Toolset` to `EmbeddingsBuilder`, they are all converted to `EmbeddableTool`. @cvauclair what do you think?

cvauclair:
@garance-buricatu can you split the PR into two: 1) a PR that removes `DocumentEmbeddings`, adds the `Embeddable` trait, etc.; and 2) a PR specifically for the macro implementation? 

That would make it much easier to review. Thank you!

rig-core/src/embeddings/embeddable.rs
```diff
@@ -0,0 +1,220 @@
...
+}
+
+/// Create a OneOrMany object with a list of items.
+impl<T> From<Vec<T>> for OneOrMany<T> {
+    fn from(items: Vec<T>) -> Self {
+        let mut iter = items.into_iter();
+        OneOrMany {
+            first: match iter.next() {
+                Some(item) => item,
+                None => panic!("Cannot create OneOrMany with an empty vector."),
```
garance-buricatu, line 97: I know that this will seem bad but here's the reason why I did this:

I have to implement `Embeddable` trait for a vector of items that implement `Embeddable`, ie:
```
impl<T: Embeddable> Embeddable for Vec<T> {
      type Error = <SOMETHING>;

     fn embeddable(&self) -> Result<OneOrMany<String>, Self::Error> {
        let items = self
            .iter()
            .map(|item| item.embeddable())
            .collect::<Result<Vec<_>, _>>()?; // <-- T::Error raised here maybe

        Ok(OneOrMany::try_from(items)) // <-- OneToMany empty list error raised here maybe    }
}
```
(see line 209). This is necessary because a big use case is having multiple strings to embed per object. In the `embeddable` function implementation, there are 2 possible error types:
1. `T::Error`
2. Some error which is raised if you try to create a `OneToManyError` with a empty vector.

So what do we set `type Error` to? Obviously, we can't convert `T::Error` to `OneToManyError` and vise versa.
**Option 1:** Create an error type that wraps the 2 error types together, like:
```
pub enum SomeError {
    #[error("OneToManyError: Can't create OneToMany with empty vector")]
    OneToManyError(()),
    #[error("Error: {}")]
    Error(String),
}
```
this isn't the best IMO because I would have to convert `T::Error` to a string, in essence losing `T`'s actual error type:
```
     let items = self
            .iter()
            .map(|item| item.embeddable())
            .collect::<Result<Vec<_>, _>>().map(|e| SomeError::Error(e.to_string()))?;
```
**Option 2:** Use `T::Error` only
* panic if we attempt to create `OneToMany` with an empty vector. Now, the function has a single error type `T::Error` and we can return it.

rig-core/src/embeddings/embeddable.rs
```diff
@@ -0,0 +1,359 @@
...
+    /// Get the first item in the list.
+    pub fn first(&self) -> T {
+        self.first.clone()
+    }
+
+    /// Get the rest of the items in the list (excluding the first one).
+    pub fn rest(&self) -> Vec<T> {
+        self.rest.clone()
+    }
+
```
garance-buricatu, line 71: Note: I created the methods `one` and `many` to create objects of type `OneOrMany` instead of doing 
`impl From<T> for OneOrMany<T>` and `impl From<Vec<T>> for OneOrMany<T>` because these are conflicting. See my comment below for why I would use  `impl From<Vec<T>> for OneOrMany<T>` instead of  `impl TryFrom<Vec<T>> for OneOrMany<T>`.

rig-core/src/embeddings/embeddable.rs
```diff
@@ -0,0 +1,359 @@
+//! The module defines the [Embeddable] trait, which must be implemented for types that can be embedded.
+//! //! # Example
```
cvauclair, line 2: Docstring typo

rig-core/src/lib.rs
cvauclair, line 1: Fix references in module docstrings

rig-core/src/embeddings/embedding.rs
```diff
@@ -0,0 +1,99 @@
+//! The module defines the [EmbeddingModel] trait, which represents an embedding model that can
+//! generate embeddings for documents. It also provides an implementation of the [EmbeddingsBuilder]
```
cvauclair, line 2: Fix docstring reference to `EmbeddingsBuilder`

rig-core/src/embeddings/embeddable.rs
```diff
@@ -0,0 +1,359 @@
...
+/// Error type used for when the `embeddable` method fails.
+/// Used by default implementations of `Embeddable` for common types.
+#[derive(Debug, thiserror::Error)]
+pub enum EmbeddableError {
+    #[error("SerdeError: {0}")]
+    SerdeError(#[from] serde_json::Error),
+}
+
+/// Trait for types that can be embedded.
+/// The `embeddable` method returns a OneOrMany<String> which contains strings for which embeddings will be generated by the embeddings builder.
```
cvauclair, line 40: Put `OneOrMany<String>` inside backticks

rig-core/src/embeddings/embeddable.rs
```diff
@@ -0,0 +1,359 @@
...
+/// If multiple items are present, `first` will contain the first item and `rest` will contain the rest.
+/// IMPORTANT: this struct cannot be created with an empty vector.
+/// OneOrMany objects can only be created using OneOrMany::one() or OneOrMany::many().
+#[derive(PartialEq, Eq, Debug, Clone)]
+pub struct OneOrMany<T> {
+    /// First item in the list.
+    first: T,
+    /// Rest of the items in the list.
+    rest: Vec<T>,
+}
```
cvauclair, line 59: Move `OneOrMany` and all its implementations to a new toplevel module `vec_utils.rs` (this will probably be reused throughout the lib)

rig-core/src/embeddings/embeddable.rs
```diff
@@ -0,0 +1,359 @@
...
+//! struct FakeDefinition {
+//!     id: String,
+//!     word: String,
+//!     definitions: Vec<String>,
+//! }
+//!
+//! let fake_definition = FakeDefinition {
+//!     id: "doc1".to_string(),
+//!     word: "hello".to_string(),
+//!     definition: "used as a greeting or to begin a conversation".to_string()
```
cvauclair, line 17: Typo: should be `definitions` with an `s` (and the value should be a `Vec`)

rig-core/src/embeddings/embeddable.rs
```diff
@@ -0,0 +1,359 @@
...
+        if self.index == 0 {
+            item = Some(&self.one_or_many.first)
+        } else if self.index - 1 < self.one_or_many.rest.len() {
+            item = Some(&self.one_or_many.rest[self.index - 1]);
+        };
+
+        self.index += 1;
+        item
+    }
+}
```
cvauclair, line 123: Idk if this is the best option as opposed to simply implementing `IntoIterator` for `OneOrMany` like so:
```rust
impl<T: Clone> IntoIterator for OneOrMany<T> {
    type Item = T;
    type IntoIter = std::iter::Chain<std::iter::Once<T>, std::vec::IntoIter<T>>;

    fn into_iter(self) -> Self::IntoIter {
        std::iter::once(self.first).chain(self.rest)
    }
}
```

rig-core/src/embeddings/embeddable.rs
```diff
@@ -0,0 +1,359 @@
...
+            .flat_map(|one_or_many| one_or_many.into_iter())
+            .collect::<Vec<_>>();
+
+        OneOrMany::many(items)
+    }
+}
+
+//////////////////////////////////////////////////////
+/// Implementations of Embeddable for common types ///
+//////////////////////////////////////////////////////
```
cvauclair, line 175: Might seem pedantic, but for consistency with the rest of the lib, could you change this to:
```rust
// ================================================================
// Implementations of Embeddable for common types
// ================================================================
```
We use this delimiter format in the rest of the lib (with exactly 64 `=` characters)

rig-core/src/embeddings/embeddable.rs
cvauclair: You could change the `EmbeddableError` to this:
```rust
#[error("{0}")]
pub struct EmbeddableError(#[from] Box<dyn std::error::Error + Send + Sync>);
```
And you can then convert any error into it, e.g.
```rust
impl Embeddable for serde_json::Value {
    type Error = EmbeddableError;

    fn embeddable(&self) -> Result<OneOrMany<String>, Self::Error> {
        Ok(OneOrMany::one(
            serde_json::to_string(self).map_err(|err| EmbeddableError(Box::new(err)))?,  // <-------- THIS
        ))
    }
}
```

Once that is done, then `OneOrMany::many` can return a `Result`, for which the error can be converted into an `EmbeddableError`

cvauclair (changes requested):
Not done reviewing, but there's a lot of stuff to change so here is the first part

rig-core/tests/embeddable_macro.rs
```diff
@@ -0,0 +1,187 @@
...
+            speech: "noun".to_string(),
+            word: "a building in which people live; residence for human beings.".to_string(),
+            link: "https://www.dictionary.com/browse/house".to_string(),
+        },
+    };
+
+    println!(
+        "FakeDefinition: {}, {}",
+        fake_definition.id, fake_definition.word
+    );
```
cvauclair, line 41: Can we remove the `println!` statements?

rig-core/rig-core-derive/src/basic.rs
```diff
@@ -0,0 +1,29 @@
+use syn::{parse_quote, Attribute, DataStruct, Meta};
+
+use crate::EMBED;
+
+/// Finds and returns fields with simple #[embed] attribute tags only.
+pub(crate) fn basic_embed_fields(data_struct: &DataStruct) -> impl Iterator<Item = syn::Field> {
+    data_struct.fields.clone().into_iter().filter(|field| {
```
cvauclair, line 7: Unneeded clone, replace `.clone().into_iter()` by `.iter()`

rig-core/rig-core-derive/src/basic.rs
```diff
@@ -0,0 +1,29 @@
...
+        field
+            .attrs
+            .clone()
+            .into_iter()
```
cvauclair, line 11: Replace `.clone().into_iter()` by `.iter()`

rig-core/rig-core-derive/src/custom.rs
```diff
@@ -0,0 +1,125 @@
...
+
+/// Finds and returns fields with #[embed(embed_with = "...")] attribute tags only.
+/// Also returns the "..." part of the tag (ie. the custom function).
+pub(crate) fn custom_embed_fields(
+    data_struct: &syn::DataStruct,
+) -> syn::Result<Vec<(syn::Field, syn::ExprPath)>> {
+    data_struct
+        .fields
+        .clone()
+        .into_iter()
```
cvauclair, line 16: Replace `.clone().into_iter()` by `.iter()`

rig-core/rig-core-derive/src/custom.rs
```diff
@@ -0,0 +1,125 @@
...
+        .filter_map(|field| {
+            field
+                .attrs
+                .clone()
+                .into_iter()
```
cvauclair, line 21: Replace `.clone().into_iter()` by `.iter()`

cvauclair (changes requested):

rig-core/src/one_or_many.rs
```diff
@@ -0,0 +1,278 @@
...
+/// Error type for when trying to create a OneOrMany object with an empty vector.
+#[derive(Debug)]
+pub struct EmptyListError;
+
+impl std::fmt::Display for EmptyListError {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "Cannot create OneOrMany with an empty vector.")
+    }
+}
+impl std::error::Error for EmptyListError {}
```
cvauclair, line 23: Use `derive(thiserror::Error)` instead of implementing `Display` and `Error` manually

cvauclair (changes requested):
Excellent! Just a tiny change to the OneOrMany error definition and then you can merge :)

PR #64, opened 2024-10-18
Feat(embeddings): Add embeddable trait to embeddings builder

marieaurore123:
This PR deals with the `EmbeddingsBuilder`.

* The builder now only handles items that implement the `Embeddable` trait.
* The builder contains documents of type `(T: Embeddable, OneOrMany<String>)`

All examples using the embeddings builder had to be changed as well.


rig-core/src/one_or_many.rs
```diff
@@ -27,6 +27,22 @@ impl<T: Clone> OneOrMany<T> {
         self.rest.clone()
     }
 
+    /// After `OneOrMany<T>` is created, add an item of type T to the `rest`.
+    pub fn add(&mut self, item: T) {
+        self.rest.push(item);
+    }
```
cvauclair, line 33: We should follow the same method naming convention as the `std::vec::Vec`, so `push` instead of `add` (see https://doc.rust-lang.org/std/vec/struct.Vec.html#method.push)

cvauclair (changes requested):

rig-core/src/one_or_many.rs
```diff
@@ -27,6 +27,22 @@ impl<T: Clone> OneOrMany<T> {
...
+
+    /// Length of all items in `OneOrMany<T>`.
+    pub fn len(&self) -> usize {
+        1 + self.rest.len()
+    }
+
+    /// If `OneOrMany<T>` is empty. This will always be false because you cannot create an empty `OneOrMany<T>`.
+    /// This methos is required when the method `len` exists.
```
cvauclair, line 41: `methos` -> `method` :upside_down_face:

cvauclair (changes requested):
Found a small typo in docstring, but once that is fixed you can merge!

PR #67, opened 2024-10-21
Fix/anthropic bad request

### LINKED ISSUE #65: bug(anthropic): Client doesn't work without temperature and max_tokens

## Bug Report

I kept getting a 400 error when I tried to make a request to Anthropic (incidentally printing the response body for that error would be very helpful to debug this). I couldn't get what I was doing wrong, until I finally figured out rig is sending a 'temperature' and 'max_tokens' as null, and Anthropic rejects the request if you do so.

Leaving *out* `temperature` and `max_tokens` works, but they cannot be null.

## Reproduction

```rust

    let api_key = std::env::var("ANTHROPIC_API_KEY").unwrap_or_default();

    let anthropic_client = anthropic::ClientBuilder::new(&api_key).build();

    let agent = anthropic_client
        .agent(anthropic::CLAUDE_3_5_SONNET)
        // without max tokens and temperature it won't work
        .max_tokens(1000)
        .temperature(1.0) 
        .build();

    // Prompt the model and print its response
    let response = agent
        .prompt("Who are you?")
        .await
        .expect("Failed to prompt Claude Sonnet");

    println!("Claude Sonnet: {response}");
```

## Expected behavior

I expect it to work without setting max_tokens and temperature explicitly. Either send the default temperature (1.0) and max_tokens (varies per model) or set these fields to undefined (if possible this would be simpler).

### LINKED ISSUE #66: bug(anthropic): tool use returns error

## Bug Report

I'm trying to use with Anthropic. I can't get it to work, I get 400 as soon as I include any tools.

## Reproduction

Any anthropic call with tools should replicate this behavior.

## Expected behavior

Anthropic should use tools

## Additional context

I did some debugging and there are a few things that seem to hinder tool use being initiated at all:

* For some reason there's a 'cache_control' in `ToolDefinition` for Anthropic. That's not documented as far as I can see, and disabling seemed to improve the situation.

* `tool_choice: {"type": "auto"}` seems to be required in the request JSON to get tool usage at all.

When I make both modifications to rig, things seem to go further; Anthropic thinks for a while, but finally I get the following error:

```
Failed to prompt Claude Sonnet: CompletionError(HttpError(reqwest::Error { kind: Decode, source: Error("data did not match any variant of untagged enum Content", line: 0, column: 0) }))
```

So it appears Anthropic returns something that isn't properly handled by rig?

## Testing?

I like Rig's API so far, and I know it's alpha, but while the Anthropic functionality is in the API my impression (also from #65)  is that not really ready yet. It strikes me that some kind of integration testing is in order in order to determine a client really works for the various works cases, and doesn't get broken.  

Automated testing would be nice, but testing with an external service is always tricky to do automatically. Right now I'm testing it. :)

cvauclair:
## Description

Fixes multiple bugs with the Anthropic integration related to request generation and response handling. Also adds the `from_env` constructor to Anthropic and Cohere's provider clients, as well as proper provider error reporting.

Fixes #66 
Fixes #65 

## Type of change

Please delete options that are not relevant.

- [x] Bug fix
- [x] New feature

## Testing

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce your results.

- [ ] Ran the `anthropic_agent` example without specifying `temperature` (succeeded)
- [ ] Ran the `anthropic_agent` example without specifying `max_tokens` (failed as expected with proper error message)
- [ ] Ran the `agent_with_tools` example with a Anthropic's `CLAUDE_3_HAIKU` (succeeded)

## Checklist:
- [x] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
  - Will be added with other integration tests at some later point 

## Notes
These bugs highlight the need for typed provider specific Request models so that these kinds of errors can be caught earlier while reducing the need for manipulating raw JSON while building the underlying request (which is error prone)

### DIFF
--- a/rig-core/examples/agent_with_tools.rs
+++ b/rig-core/examples/agent_with_tools.rs
@@ -6,7 +6,6 @@ use rig::{
 };
 use serde::{Deserialize, Serialize};
 use serde_json::json;
-use std::env;
 
 #[derive(Deserialize)]
 struct OperationArgs {
@@ -92,25 +91,13 @@ impl Tool for Subtract {
 #[tokio::main]
 async fn main() -> Result<(), anyhow::Error> {
     // Create OpenAI client
-    let openai_api_key = env::var("OPENAI_API_KEY").expect("OPENAI_API_KEY not set");
-    let openai_client = providers::openai::Client::new(&openai_api_key);
+    let openai_client = providers::openai::Client::from_env();
 
     // Create agent with a single context prompt and two tools
-    let gpt4_calculator_agent = openai_client
-        .agent("gpt-4")
-        .context("You are a calculator here to help the user perform arithmetic operations.")
-        .tool(Adder)
-        .tool(Subtract)
-        .build();
-
-    // Create OpenAI client
-    let cohere_api_key = env::var("COHERE_API_KEY").expect("COHERE_API_KEY not set");
-    let cohere_client = providers::cohere::Client::new(&cohere_api_key);
-
-    // Create agent with a single context prompt and two tools
-    let coral_calculator_agent = cohere_client
-        .agent("command-r")
-        .preamble("You are a calculator here to help the user perform arithmetic operations.")
+    let calculator_agent = openai_client
+        .agent(providers::openai::GPT_4O)
+        .preamble("You are a calculator here to help the user perform arithmetic operations. Use the tools provided to answer the user's question.")
+        .max_tokens(1024)
         .tool(Adder)
         .tool(Subtract)
         .build();
@@ -118,12 +105,8 @@ async fn main() -> Result<(), anyhow::Error> {
     // Prompt the agent and print the response
     println!("Calculate 2 - 5");
     println!(
-        "GPT-4: {}",
-        gpt4_calculator_agent.prompt("Calculate 2 - 5").await?
-    );
-    println!(
-        "Coral: {}",
-        coral_calculator_agent.prompt("Calculate 2 - 5").await?
+        "Calculator Agent: {}",
+        calculator_agent.prompt("Calculate 2 - 5").await?
     );
 
     Ok(())
--- a/rig-core/src/providers/anthropic/completion.rs
+++ b/rig-core/src/providers/anthropic/completion.rs
@@ -196,14 +196,15 @@ impl completion::CompletionModel for CompletionModel {
             .post("/v1/messages")
             .json(&request)
             .send()
-            .await?
-            .error_for_status()?
-            .json::<ApiResponse<CompletionResponse>>()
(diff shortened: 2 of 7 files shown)


rig-core/src/providers/anthropic/completion.rs
```diff
@@ -157,9 +151,20 @@ impl completion::CompletionModel for CompletionModel {
...
         completion_request: completion::CompletionRequest,
     ) -> Result<completion::CompletionResponse<CompletionResponse>, CompletionError> {
+        // Note: Ideally we'd introduce provider-specific Request models to handle the
+        // specific requirements of each provider. For now, we just manually check while
+        // building the request as a raw JSON document.
+
         let prompt_with_context = completion_request.prompt_with_context();
 
-        let request = json!({
+        // Check if max_tokens is set, required for Anthropic
```
faassen, line 160: I would have expected this to be done in the `ClientBuilder`, but I see that while the API docs right now encourage you to use the builder they don't *require* it.

Since `max_tokens` is required, shouldn't it prevent use of Anthropic without this on the type level though? That is, make `max_tokens` required for `Client::new` and fail `ClientBuilder::build` if it's not provided, or alternatively make `new` non-public so that the builder is the only way.

This would have the compiler guide you in the right direction, rather than getting an error in runtime.

faassen (commented):
Thanks for looking into this! I can't judge most of the changes yet but I left a comment about max_tokens.

rig-core/src/providers/anthropic/completion.rs
0xMochan: This would be ideal, but probably should be done as a further PR since it's something we might want to standardize for all the providers. ATM, Anthropic is the only one w/ a `ClientBuilder`.

rig-core/examples/agent_with_tools.rs
0xMochan, line 1: Is this file supposed to be in this PR, this doesn't seem to be related to anthropic as far as I can tell.

0xMochan (approved):
There seems to be more changes than just anthropic here but overall looks good. The comment about making the `ClientBuilder` validate via compile-time type checking is def. something we should do in a future PR.

cvauclair:
> There seems to be more changes than just anthropic here but overall looks good. The comment about making the `ClientBuilder` validate via compile-time type checking is def. something we should do in a future PR.

One of the fixes (proper error reporting) was also applied to all other providers (which had the same problem as Anthropic) so the PR does touch on more than Anthropic because of that. As for the example, I just noticed it was a bit messy before so I cleaned it up (but true this could have been done in another PR :upside_down_face: )

PR #69, opened 2024-10-21
refactor(vector-store): update generics

garance-buricatu:
Converts this:
```
pub trait VectorStoreIndex: Send + Sync {
    /// Get the top n documents based on the distance to the given query.
    /// The result is a list of tuples of the form (score, id, document)
    fn top_n<T: for<'a> Deserialize<'a> + Send>(
        &self,
        query: &str,
        n: usize,
    ) -> impl std::future::Future<Output = Result<Vec<(f64, String, T)>, VectorStoreError>> + Send;

    /// Same as `top_n` but returns the document ids only.
    fn top_n_ids(
        &self,
        query: &str,
        n: usize,
    ) -> impl std::future::Future<Output = Result<Vec<(f64, String)>, VectorStoreError>> + Send;
}
```

to this:
```
pub trait VectorStoreIndex<T>: Send + Sync
where
    T: for<'a> Deserialize<'a> + Send,
{
    /// Get the top n documents based on the distance to the given query.
    /// The result is a list of tuples of the form (score, id, document)
    fn top_n(
        &self,
        query: &str,
        n: usize,
    ) -> impl std::future::Future<Output = Result<Vec<(f64, String, T)>, VectorStoreError>> + Send;

    /// Same as `top_n` but returns the document ids only.
    fn top_n_ids(
        &self,
        query: &str,
        n: usize,
    ) -> impl std::future::Future<Output = Result<Vec<(f64, String)>, VectorStoreError>> + Send;
}
```

**Note side-effects:**
This is not possible:
```
impl<T, I> VectorStoreIndexDyn for I
where
    I: VectorStoreIndex<T> + Send + Sync,
    T: for<'a> Deserialize<'a> + Serialize + Send,
```
as it create ambiguous implementations of VectorStoreIndexDyn. Instead I had to do:
```
impl<I: VectorStoreIndex<serde_json::Value>> VectorStoreIndexDyn for I
```
ie. we have to define a specific type for `T` if we want to implement `VectorStoreIndexDyn` for `VectorStoreIndex<T>`.

This means that for methods that expect type `impl VectorStoreIndexDyn` , the vector store can only contain items of type `Value`. **This is the case for anything that uses the rag agent**. For example:

```
let index = InMemoryVectorStore::default()
        .add_tools(embeddings)?
        .index(embedding_model);

    // Create RAG agent with a single context prompt and a dynamic tool source
    let calculator_rag = openai_client
        .agent("gpt-4")
        .preamble("" )
        .dynamic_tools(4, index, toolset)
        .build();
```
Since `.dynamic_tools()` expects `index` to implement `VectorStoreIndexDyn` and only `VectorStoreIndex<serde_json::Value>` implements `VectorStoreIndexDyn`, the vector store must contain items of type `serde_json::Value`. The method `add_tools` converts `EmbeddableTool` to a `Value`.

### DIFF
--- a/rig-core/examples/calculator_chatbot.rs
+++ b/rig-core/examples/calculator_chatbot.rs
@@ -252,7 +252,12 @@ async fn main() -> Result<(), anyhow::Error> {
         .await?;
 
     let index = InMemoryVectorStore::default()
-        .add_tools(embeddings)?
+        .add_documents(
+            embeddings
+                .into_iter()
+                .map(|(tool, embedding_vec)| (tool.name.clone(), tool, embedding_vec))
+                .collect(),
+        )?
         .index(embedding_model);
 
     // Create RAG agent with a single context prompt and a dynamic tool source
--- a/rig-core/examples/calculator_chatbot.rs
+++ b/rig-core/examples/calculator_chatbot.rs
@@ -252,12 +252,7 @@ async fn main() -> Result<(), anyhow::Error> {
         .await?;
 
     let index = InMemoryVectorStore::default()
-        .add_documents(
-            embeddings
-                .into_iter()
-                .map(|(tool, embedding)| (tool.name.clone(), tool, embedding))
-                .collect(),
-        )?
+        .add_tools(embeddings)?
         .index(embedding_model);
 
     // Create RAG agent with a single context prompt and a dynamic tool source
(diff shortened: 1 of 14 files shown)


garance-buricatu:
Needs more work on fixing issue mentioned in PR description. Potential solution: custom macro to implement `VectorStoreIndexDyn` on specific types defined by user.

PR #70, opened 2024-10-22
feat(vector-store): Add support for Neo4j 

mateobelanger:
This pull request introduces a new Neo4j-based vector store implementation for the Rig project. The key changes include adding the new `rig-neo4j` crate, updating the project structure, and providing documentation and examples for the new functionality.

### New Crate Addition:
* **`rig-neo4j` crate**: Added a new crate `rig-neo4j` to the project, which includes the implementation of a Neo4j-based vector store.

### Examples:
- [examples/vector_search_simple.rs](examples/vector_search_simple.rs) shows how to create an index on simple data.
- [examples/vector_search_movies_consume.rs](examples/vector_search_movies_consume.rs) shows how to query an existing index.
- [examples/vector_search_movies_create.rs](examples/vector_search_movies_create.rs) shows how to create embeddings & index on a large DB and query it in one go.


rig-neo4j/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+/// A Neo4j vector store.
+pub struct Neo4jVectorStore {
+    //collection: mongodb::Collection<DocumentEmbeddings>,
+    pub client: Neo4jClient,
+    pub database_name: String,
+}
+
+fn neo4j_to_rig_error(e: neo4rs::Error) -> VectorStoreError {
+    VectorStoreError::DatastoreError(Box::new(e))
+}
```
0xMochan, line 23: I think you should be able to use `thiserror` to avoid manually `Box`'ing errors when converting from library errors to crate errors

rig-neo4j/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+
+impl VectorStore for Neo4jVectorStore {
+    type Q = neo4rs::Query;
+
+    async fn add_documents(
+        &mut self,
+        documents: Vec<DocumentEmbeddings>,
+    ) -> Result<(), VectorStoreError> {
+        for doc in documents {
+            let query = Query::new("CREATE (d:DocumentEmbeddings {id: $id, document: $document, embeddings: $embeddings})".to_string())
```
0xMochan, line 65: Maybe use a const for premade queries.

rig-neo4j/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+    embedding_model: M,
+    index_name: String,
+    search_params: SearchParams,
+}
+
+impl<M: EmbeddingModel> Neo4jVectorIndex<M> {
+    const BASE_VECTOR_SEARCH_QUERY: &str = "
+    CALL db.index.vector.queryNodes($index_name, $num_candidates, $queryVector)
+    YIELD node, score
+    ";
```
0xMochan, line 139: Same as earlier

rig-neo4j/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+            "
+            {}
+            {}
+            RETURN score, ID(node) as element_id {}
+            ",
+            Self::BASE_VECTOR_SEARCH_QUERY,
+            where_clause,
+            if return_node { ", node as node" } else { "" }
+        ))
+        .param("queryVector", prompt_embedding.vec)
```
0xMochan, line 178: Is it intended that some params are camelCased while others are snake_cased?

rig-neo4j/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+    ) -> Result<Vec<(f64, String)>, VectorStoreError> {
+        let prompt_embedding = self.embedding_model.embed_document(query).await?;
+
+        let query = self.build_vector_search_query(prompt_embedding, false, n);
+
+        #[derive(Debug, Deserialize)]
+        struct RowResult {
+            score: f64,
+            element_id: i64,
+        }
```
0xMochan, line 286: Ditto as earlier

0xMochan (changes requested):
I have some questions I brought up here, there's some small strings that should be `const`s. I also noticed there's a lot of `TODO`s here about incomplete client behavior. Is this bc the rust port of neo4j isn't finished yet? If so, you should clarify this in the docstrings.

rig-neo4j/src/lib.rs
mateobelanger: nope, thanks! These are queries from the neo4j docs

rig-neo4j/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+    ) -> Result<Vec<(f64, String, T)>, VectorStoreError> {
+        let prompt_embedding = self.embedding_model.embed_document(query).await?;
+        let query = self.build_vector_search_query(prompt_embedding, true, n);
+
+        #[derive(Debug, Deserialize)]
+        struct RowResult<T> {
+            node: T,
+            score: f64,
+            element_id: i64,
+        }
```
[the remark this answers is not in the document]
mateobelanger: The struct is only used to wrap the user's type and deserialize the query result. Thus it's only relevant inside this function (for now).

Makes sense or bad practice you would say?

rig-neo4j/src/lib.rs
mateobelanger: This will be removed as the VectorStore trait is no more 🪦.

rig-neo4j/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+
+use rig::{
+    embeddings::{DocumentEmbeddings, Embedding, EmbeddingModel},
+    vector_store::{VectorStore, VectorStoreError, VectorStoreIndex},
+};
+use serde::Deserialize;
+
+pub struct Neo4jClient {
+    pub graph: Graph,
+}
```
cvauclair, line 12: Is this type necessary?

rig-neo4j/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+            .map_err(neo4j_to_rig_error)?;
+        let mut results = Vec::new();
+        while let Ok(Some(row)) = result.next().await {
+            let row_parsed = row
+                .to::<T>()
+                .map_err(|e| VectorStoreError::DatastoreError(Box::new(e)))?;
+
+            results.push(row_parsed);
+        }
+        Ok(results)
```
cvauclair, line 200: You can replace this whole thing by:
```rust
self.graph
    .neo4j
    .execute(query)
    .await
    .map_err(neo4j_to_rig_error)?
    .into_stream_as::<T>()
    .try_collect::<Vec<_>>()
    .await
    .map_err(|e| VectorStoreError::DatastoreError(Box::new(e)))?
```

cvauclair (changes requested):
Good work! But a couple things to change

rig-neo4j/src/lib.rs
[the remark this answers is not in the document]
cvauclair: I don't think it's strictly bad practice, but maybe this type could be useful elsewhere and made public :thinking:

rig-neo4j/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+
+[dependencies]
+futures = "0.3.30"
+neo4rs = "0.8.0"
+rig-core = { path = "../rig-core", version = "0.3.0" }
+serde = { version = "1.0.203", features = ["derive"] }
+serde_json = "1.0.117"
+tracing = "0.1.40"
+textwrap = { version = "0.16.1", optional = true }
+term_size = { version = "0.3.2", optional = true }
```
cvauclair, line 20: Move these dependencies to the `dev-dependencies` since they are not used in the actual crate. Note: `dev-dependencies` apply to both tests and examples!

rig-neo4j/examples/data/recommendations-50.dump
cvauclair, line 1: Any way to make this file downloadable via a command as opposed to having it in the repo? Im thinking of something like a `wget` command in the README that users can just copy to run the examples

rig-neo4j/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+
+[dev-dependencies]
+anyhow = "1.0.86"
+tokio = { version = "1.38.0", features = ["macros"] }
+
+
+[features]
+default = ["display"]
+display = ["textwrap", "term_size"]
```
cvauclair, line 29: Remove those feature flags

rig-neo4j/examples/vector_search_simple.rs
```diff
@@ -0,0 +1,138 @@
...
+    .await;
+
+    // Unwrap the results in the vector _create_nodes
+    for result in _create_nodes {
+        result.unwrap(); // or handle the error appropriately
+    }
+
+    // Create a vector index on our vector store
+    println!("Creating vector index...");
+    let _ = neo4j_client
```
cvauclair, line 82: No need for `let _ = `

rig-neo4j/examples/vector_search_simple.rs
```diff
@@ -0,0 +1,138 @@
...
+    #[derive(serde::Deserialize)]
+    struct Document {
+        #[allow(dead_code)]
+        id: String,
+        document: String,
+        #[allow(dead_code)]
+        embedding: Vec<f32>,
+    }
+
+    let _create_nodes = futures::future::join_all(embeddings.into_iter().map(|doc| {
```
cvauclair, line 55: This is ok, but using `Streams` from the `futures` crate is usually preferred (`Streams` are like the async version of `Iterators`, i.e.: closures can be async)

rig-neo4j/src/lib.rs
```diff
@@ -0,0 +1,266 @@
...
+        providers::openai::{Client, TEXT_EMBEDDING_ADA_002},
+        vector_store::VectorStoreIndex,
+    };
+    use serde::Deserialize;
+    use std::env;
+
+    const NEO4J_URI: &str = "neo4j+s://demo.neo4jlabs.com:7687";
+    const NEO4J_DB: &str = "recommendations";
+    const NEO4J_USERNAME: &str = "recommendations";
+    const NEO4J_PASSWORD: &str = "recommendations";
```
cvauclair, line 192: Secrets in repo :grimacing:

rig-neo4j/src/vector_index.rs
```diff
@@ -0,0 +1,358 @@
...
+                neo4rs::query("CALL db.awaitIndex($index_name, 10000)")
+                    .param("index_name", self.index_config.index_name.clone()),
+            )
+            .await;
+
+        if index_exists.is_err() {
+            tracing::warn!(
+                "Index with name `{}` is not ready or could not be created.",
+                self.index_config.index_name
+            );
```
cvauclair, line 149: Shouldn't this also return an `Error`? Cause right now the function returns `Ok` even if the index was not created

rig-neo4j/src/vector_index.rs
```diff
@@ -0,0 +1,358 @@
...
+    }
+}
+
+// ===============================
+// Utilities to print results from a vector search
+// ===============================
+
+#[cfg(feature = "display")]
+#[allow(dead_code)]
+pub mod display {
```
cvauclair, line 303: I'd move in the examples directory, as it's really only useful there and not really part of the scope of the companion crate. See the examples in `rig-lancedb` for how to have modules in the `examples` folder.

cvauclair (changes requested):
Couple things to change still but after that (and after tests pass) should be good to merge!

rig-neo4j/src/lib.rs
mateobelanger: Not a secret! It's a demo database. 

https://github.com/neo4j-graph-examples/recommendations/tree/main?tab=readme-ov-file#setup

rig-neo4j/src/vector_index.rs
mateobelanger: Good question. I was not able to parse the error coming from this call. Partly because if the creation fails (for instance if there is already an index of another name) the DB does not fire an error.

image

rig-neo4j/src/vector_index.rs
mateobelanger: I thought it was useful to have a way to visualize nicely the query results. 

You don`t think it's relevant? Otherwise yes I'll remove the feature flags and move this to /examples

rig-neo4j/examples/vector_search_simple.rs
mateobelanger: true! I was mistakenly using .execute() instead of run() before.

rig-neo4j/examples/data/recommendations-50.dump
mateobelanger: Will remove the file! Not useful anymore.

rig-neo4j/examples/vector_search_simple.rs
mateobelanger: will change it to streams!

rig-neo4j/src/lib.rs
```diff
@@ -0,0 +1,272 @@
...
+        Neo4jVectorIndex::new(self.graph.clone(), model, index_config, search_params)
+    }
+}
+
+#[allow(dead_code)]
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use display;
+    use display::SearchResult;
```
cvauclair, line 189: The only test that needs this... is a test for the display functionality! So remove both the test and this import

rig-neo4j/examples/vector_search_movies_add_embeddings.rs
```diff
@@ -0,0 +1,165 @@
...
+
+use neo4rs::*;
+use rig_neo4j::{
+    vector_index::{IndexConfig, SearchParams},
+    Neo4jClient, ToBoltType,
+};
+use serde::{Deserialize, Serialize};
+
+#[path = "./display/lib.rs"]
+mod display;
```
cvauclair, line 23: This is not needed (see my comment in the test)

cvauclair (changes requested):

rig-neo4j/src/lib.rs
cvauclair: @mateobelanger the `display` import is still present in the `test` module

cvauclair (approved):

0xMochan (approved):
Looks good!

PR #89, opened 2024-11-06
refactor(embeddings): embed trait definition

marieaurore123:
Update `Embed` trait definition. Change all dependent code accordingly.


cvauclair (changes requested):
Good work! Couple comments but overall this is good stuff!

cvauclair (approved):

PR #104, opened 2024-11-13
fix(rig-lancedb): rag embedding filtering

marieaurore123:


rig-core/src/providers/anthropic/completion.rs
```diff
@@ -214,7 +214,13 @@ impl completion::CompletionModel for CompletionModel {
 
         if response.status().is_success() {
             match response.json::<ApiResponse<CompletionResponse>>().await? {
-                ApiResponse::Message(completion) => completion.try_into(),
+                ApiResponse::Message(completion) => {
+                    tracing::info!(target: "rig",
+                        "Anthropic completion token usage: {:?}",
```
cvauclair, line 219: Implement `Diplay` for `Usage` and have it display the stats as a nice string instead of using the debug output which is not very good for logging stuff

rig-core/src/providers/cohere.rs
```diff
@@ -217,6 +217,11 @@ impl embeddings::EmbeddingModel for EmbeddingModel {
         if response.status().is_success() {
             match response.json::<ApiResponse<EmbeddingResponse>>().await? {
                 ApiResponse::Ok(response) => {
+                    tracing::info!(target: "rig",
+                        "Cohere embeddings billed units: {:?}",
+                        response.meta.map(|meta| meta.billed_units)
```
cvauclair, line 222: Same here, instead of using the debug output, just convert the `billed_units` to a string (with `n/a` as default if `billed_units` is `None`)

rig-core/src/providers/gemini/completion.rs
```diff
@@ -122,6 +122,11 @@ impl completion::CompletionModel for CompletionModel {
             .json::<GenerateContentResponse>()
             .await?;
 
+        tracing::info!(target: "rig",
+            "Gemini completion token usage: {:?}",
+            response.usage_metadata
```
cvauclair, line 127: Same here, use custom `Display` implementation instead of `Debug` trait

rig-core/src/providers/openai.rs
```diff
@@ -258,6 +258,11 @@ impl embeddings::EmbeddingModel for EmbeddingModel {
         if response.status().is_success() {
             match response.json::<ApiResponse<EmbeddingResponse>>().await? {
                 ApiResponse::Ok(response) => {
+                    tracing::info!(target: "rig",
+                        "OpenAI embedding token usage: {:?}",
+                        response.usage
```
cvauclair, line 263: Use `Display` instead of `Debug`

rig-core/src/providers/perplexity.rs
```diff
@@ -235,7 +235,13 @@ impl completion::CompletionModel for CompletionModel {
 
         if response.status().is_success() {
             match response.json::<ApiResponse<CompletionResponse>>().await? {
-                ApiResponse::Ok(completion) => Ok(completion.try_into()?),
+                ApiResponse::Ok(completion) => {
+                    tracing::info!(target: "rig",
+                        "Perplexity completion token usage: {:?}",
+                        completion.usage
```
cvauclair, line 241: Use `Display` instead of `Debug`

rig-lancedb/src/utils/mod.rs
```diff
@@ -26,3 +26,58 @@ impl QueryToJson for lancedb::query::VectorQuery {
         record_batches.deserialize()
     }
 }
+
+pub trait FilterEmbeddings {
+    fn filter(&mut self, embeddings_col: Option<String>) -> serde_json::Result<serde_json::Value>;
+}
+
+impl FilterEmbeddings for serde_json::Value {
+    fn filter(&mut self, embeddings_col: Option<String>) -> serde_json::Result<serde_json::Value> {
```
cvauclair, line 35: Either `self` is mutable and `filter` modifies the object inplace, or `self` is not mutable and a copy is returned. Right now it's both which is not good practice

cvauclair (changes requested):
Nicely done! But there are a couple things to change, mostly for logging

cvauclair (approved):

PR #106, opened 2024-11-14
feat(xai): initial xai (grok) implementation

0xMochan:
Implements xAI (grok) support.

## Notes
This implementation follows very closely to the openai implementation. It seems like xai has modeled their api to match both openai _and_ anthropic (simultaneously). The docs are very unclear on most of the sub-models so the implementation was essentially matched against openai. It might be possible to test alternative BASE_URLs schemas on the openai and anthropic APIs as suggested in the docs via python/js official SDK examples.

The `v1/embeddings` endpoint is documented and lists a model `v1` in the docs. Afaik, this is an internal model not available to the public as it's not listed as available in the xAI console (see photo). Therefore, this is untested (but likely would work if you had access to this model).

image


rig-core/src/providers/xai/embedding.rs
```diff
@@ -0,0 +1,123 @@
...
+
+use crate::embeddings::{self, EmbeddingError};
+
+use super::{
+    client::xai_api_types::{ApiErrorResponse, ApiResponse},
+    Client,
+};
+
+// ================================================================
+// OpenAI Embedding API
```
mateobelanger, line 17:
```suggestion
// xAI Embedding API
```

rig-core/src/providers/xai/client.rs
```diff
@@ -0,0 +1,172 @@
...
+    embeddings::{self},
+    extractor::ExtractorBuilder,
+};
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+use super::{completion::CompletionModel, embedding::EmbeddingModel, EMBEDDING_V1};
+
+// ================================================================
+// Google Gemini Client
```
mateobelanger, line 12:
```suggestion
// xAI Client
```

mateobelanger (commented):
Can you add a reference to xAI in the readme's integration section? With the logo ideally!

cvauclair (changes requested):
Nicely done! Only thing left is to clean up the main example (which is pretty messy atm) and addressing @mateobelanger 's comments.

Also, don't forget to add `xAI` to the list of providers in the repo README and the `rig-core/src/lib.rs` lib docstrings!

0xMochan:
- Added general style comment changes
- Added xai to various READMEs
- Added some printlines to grok example

rig-core/README.md
```diff
@@ -48,5 +50,6 @@ Rig supports the following LLM providers natively:
 - OpenAI
 - Cohere
 - Google Gemini
+- xAI
 Additionally, Rig currently has the following integration sub-libraries:
```
cvauclair, line 54: This is missing a line break (renders as a single line rn)

cvauclair (changes requested):
Just a small thing to fix in the README and it should be good to go!

cvauclair (approved):

PR #115, opened 2024-11-18
fix(mongodb): remove embeddings from `top_n` lookup

0xMochan:
Eliminate the embeddings field from the top_n lookup results to streamline the response structure. Also adds a `DocumentResponse` type that doesn't include the embeddings field to be used with `top_n` responses.

Additionally, this PR un-hardcoded `embeddings.vec` and dynamically looks up the embedded field from the vector search index (which also confirms whether it exists or not properly during construction).

Update: We make a call to mongodb to check all search indexes. We only handle 1 index w/ 1 indexed embedded field but this could be updated in the future if needed.

### DIFF
--- a/rig-core/src/vector_store/mod.rs
+++ b/rig-core/src/vector_store/mod.rs
@@ -17,6 +17,9 @@ pub enum VectorStoreError {
 
     #[error("Datastore error: {0}")]
     DatastoreError(#[from] Box<dyn std::error::Error + Send + Sync>),
+
+    #[error("Vector store error: {0}")]
+    Error(String),
 }
 
 /// Trait for vector stores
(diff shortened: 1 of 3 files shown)


rig-core/src/vector_store/mod.rs
```diff
@@ -17,6 +17,9 @@ pub enum VectorStoreError {
 
     #[error("Datastore error: {0}")]
     DatastoreError(#[from] Box<dyn std::error::Error + Send + Sync>),
+
+    #[error("Vector store error: {0}")]
+    Error(String),
```
cvauclair, line 22: I'd use `DatastoreError` instead of creating a new error type here.

FYI `String` implements `Into<Box<dyn std::error::Error + Send + Sync>>`, so you can do `DatastoreError(format!("some error {some_stuff}").into())`

cvauclair (changes requested):

cvauclair (approved):

PR #118, opened 2024-11-19
fix(neo4j): remove embeddings from top_n lookup

### LINKED ISSUE #116: bug(neo4j): top_n function returns embeddings data

mateobelanger:
## Description of the Fix
`Neo4jVectorIndex::top_n::<Value>()` now prevents returning the embedding vector property. When a Rig index object is created, a DB query retrieves the associated Neo4j index and the property name that stores it's embeddings vector.

### Note

Incidentally, I also improved the flow of creating indexes from an empty DB. The client now is in charge of both DB index creation and returning the Rig index object matching an existing DB index.


rig-neo4j/examples/vector_search_movies_consume.rs
```diff
@@ -32,20 +29,27 @@ mod display;
...
     // Initialize OpenAI client
     let openai_api_key = env::var("OPENAI_API_KEY").expect("OPENAI_API_KEY not set");
     let openai_client = Client::new(&openai_api_key);
 
-    let neo4j_uri = env::var("NEO4J_URI").expect("NEO4J_URI not set");
-    let neo4j_username = env::var("NEO4J_USERNAME").expect("NEO4J_USERNAME not set");
-    let neo4j_password = env::var("NEO4J_PASSWORD").expect("NEO4J_PASSWORD not set");
+    let neo4j_uri = "neo4j+s://demo.neo4jlabs.com:7687";
+    let neo4j_username = "recommendations";
+    let neo4j_password = "recommendations";
```
0xMochan, line 45: Was there a specific reason this was hardcoded in the example?

rig-neo4j/src/lib.rs
```diff
@@ -162,19 +176,165 @@ impl Neo4jClient {
...
+            index_config: IndexConfigDetails,
+        }
+
+        #[derive(Deserialize)]
+        struct IndexConfigDetails {
+            #[serde(rename = "vector.dimensions")]
+            vector_dimensions: i64,
+            #[serde(rename = "vector.similarity_function")]
+            vector_similarity_function: String,
+        }
```
0xMochan, line 225: Why define these structs within the method here? Doesn't this make exterior usage of these awkward?

rig-neo4j/src/lib.rs
```diff
@@ -162,19 +176,165 @@ impl Neo4jClient {
...
+            )
+            .await?;
+            return Err(VectorStoreError::DatastoreError(Box::new(
+                std::io::Error::new(
+                    std::io::ErrorKind::NotFound,
+                    format!(
+                        "Index `{}` not found in database. Available indexes: {:?}",
+                        index_name, indexes
+                    ),
+                ),
```
0xMochan, line 259: This error construction seems complicated, it seems cleaner to just pass a `format!(...).into()` directly to the `DatastoreError`.

0xMochan (commented):
Left some comments!

rig-neo4j/examples/vector_search_movies_consume.rs
mateobelanger: To make the example work out of the box, it uses a demo database from Neo4j!

rig-neo4j/src/lib.rs
mateobelanger: These structures would be of no use (for now). It's to make the deserialization more readable while keeping the module less clutered. Don't know if there is a better way to do this though.

mateobelanger: @cvauclair

rig-neo4j/examples/vector_search_simple.rs
cvauclair, line 1: This example currently fails because of line 119: `.top_n::<Document>("What is a glarb?", 1)`.

`Document` expects a field `embedding` of type `Vec<f32>`. However the PR prunes this field from the Neo4j response.

Solution is simply to change `Document` to something without an `embedding` field

cvauclair (changes requested):

PR #120, opened 2024-11-23
feat: embeddings overhaul

### LINKED ISSUE #52: feat: EmbeddingsBuilder redesign
## Feature Request
Any object that implements the trait `Embedabble` can be added to the embeddings builder and therefore can be embedded.

### Motivation
Currently, embeddings builder is tightly coupled with the `DocumentEmbedding` type (only objects of this type can be used by the builder). **Downsides**: user needs to type cast their types to `DocumentEmbedding` and the structure of `DocumentEmbedding` is predetermined so the user needs to design their embedding code around this.

### Proposal
1. Define the trait `Embedabble` which contains a method `fn embedabble()` which returns a list of strings (the strings that should be embedded).
2. Change the `EmbeddingsBuilder` to accept any type that implements the trait `Embedabble`.
3. Write a custom derive macro that can be used on any struct which derives the `Embedabble` trait automatically.

### Linked issues:
#60 
#61
#62

#### Todo:

- [x] Give trait and macro same name like serde

- [x] implement Embeddable trait for many primitive types

- [x] macro fails when there are no embed tags

- [x] macro fails when embed with is empty string

- [x] need to handle error in custom embed function

- [x] review borrow and owning eveywhere. **done for embeddings.rs**

- [x] update docstrings. **done for embeddings.rs**

- [ ] clean up embedding builder definition with traits maybe?

- [x] should build() return `Embedding` object or just string? **return `Embedding` object - document string used for debugging**

- [ ] in memory vector store, rethink type of document D --> move to another branch

- [x] why is vector search of in memory vector store so bad? **bug found**

- [ ] in memory store - implement from trait from both returns types of build into add_document
- [x]  tests
- [x] make macro as feature flag

marieaurore123:
Complete overhaul of embeddings. 

See Issue: 
https://github.com/0xPlaygrounds/rig/issues/52

The biggest changes can be found in:
* `rig-core/src/embeddings` --> notice the test and doctests written to validate functionlities.
* `rig-core/riig-core-derive`

Other significant changes in all examples related to vector searching for all of the sibling crates. All of the examples have been re-compiled and ran.


0xMochan (approved):
This code is very well written, especially with how internal code comments clearly define everything and how elegant and simplified the internals have become. When I start to get to the point of nitpicking, I feel like that indicates the PR is in a great spot!

1. I'm still not a fan of how `rig-core-derive` is a subcrate of `rig-core`. I understand namewise / recommendations led to this decision but I feel, because we are a monorepo, having `rig-core-dervive` be inside the `rig-core` crate sorta hides the crate a bit. I don't think this should block the PR though since it might be a larger, more annoying move to make at the last hour of the PR.

2. A couple of the code documentation could use some `[]` so that the comments directly link to source.
  - Very much a nitpick as it can be tedious to link every single object mention (like `EmbedError` -> `[EmbedError]`).

Overall, I think this is fantastic. I think some docs on both a) how one can migrate to the new embeddings setup b) how the derive macro is leveraged and c) general usage docs would be very helpful once we get our documentation up and running!

tarrencev (commented):
It's a little confusing to refer to embedded content as both documents and texts

cvauclair (approved):

PR #122, opened 2024-11-25
feat: Add support for Sqlite vector store

### LINKED ISSUE #124: feat: Add support for Sqlite vector store
Add vector store support for Sqlite (with sqlite-vec extension).

Motivation
Sqlite is one of the most widely used open source databases and supports vector search via the [sqlite-vec](https://github.com/asg017/sqlite-vec) extension. This makes the database a good integration candidate for Rig.

Proposal
Create a new companion crate in the repository called rig-sqlite.
Implement the VectorStore and VectorStoreIndex traits for the selected Sqlite client.

Note: The integration should be similar to the MongoDB vector store integration (see [rig-mongodb](https://github.com/0xPlaygrounds/rig/blob/main/rig-mongodb/src/lib.rs)).

### Resources
[sqlite-vec](https://github.com/asg017/sqlite-vec)
[tokio-rusqlite](https://github.com/programatik29/tokio-rusqlite)

tarrencev:
# New vector store: Sqlite

## Description

Sqlite using [sqlite-vec](https://github.com/asg017/sqlite-vec)

Fixes # (issue)

Resolves #124 

## Testing

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce your results.

- Run example with `cargo run --package rig-sqlite --example vector_search_sqlite`

## Checklist:

- [x] I have performed a self-review of my own code
- [x I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] Any dependent changes have been merged and published in downstream modules
- [x] I've reviewed the vector store API documentation and implemented the types of response accurately

## Notes

I think this might have some conflicts with #120?

Perhaps it would be easiest to merge this first then adapt it in that PR or I can adapt it after that is merged. It seems like it might remove a bunch of the code.

### DIFF
--- a/rig-sqlite/src/lib.rs
+++ b/rig-sqlite/src/lib.rs
@@ -1,0 +1,360 @@
+use rig::embeddings::{DocumentEmbeddings, Embedding, EmbeddingModel};
+use rig::vector_store::{VectorStore, VectorStoreError, VectorStoreIndex};
+use rusqlite::OptionalExtension;
+use serde::Deserialize;
+use tokio_rusqlite::Connection;
+use tracing::{debug, info};
+
+#[derive(Debug)]
+pub enum SqliteError {
+    DatabaseError(Box<dyn std::error::Error + Send + Sync>),
+    SerializationError(Box<dyn std::error::Error + Send + Sync>),
+}
+
+#[derive(Clone)]
+pub struct SqliteVectorStore {
+    conn: Connection,
+}
+
+impl SqliteVectorStore {
+    pub async fn new(conn: Connection) -> Result<Self, VectorStoreError> {
+        debug!("Running initial migrations");
+        // Run migrations or create tables if they don't exist
+        conn.call(|conn| {
+            conn.execute_batch(
+                "BEGIN;
+                -- Document tables
+                CREATE TABLE IF NOT EXISTS documents (
+                    id INTEGER PRIMARY KEY AUTOINCREMENT,
+                    doc_id TEXT UNIQUE NOT NULL,
+                    document TEXT NOT NULL
+                );
+                CREATE INDEX IF NOT EXISTS idx_doc_id ON documents(doc_id);
+                CREATE VIRTUAL TABLE IF NOT EXISTS embeddings USING vec0(embedding float[1536]);
+
+                COMMIT;",
+            )
+            .map_err(tokio_rusqlite::Error::from)
+        })
+        .await
+        .map_err(|e| VectorStoreError::DatastoreError(Box::new(e)))?;
+
+        Ok(Self { conn })
+    }
+
+    fn serialize_embedding(embedding: &Embedding) -> Vec<f32> {
+        embedding.vec.iter().map(|x| *x as f32).collect()
+    }
+
+    /// Create a new `SqliteVectorIndex` from an existing `SqliteVectorStore`.
+    pub async fn index<M: EmbeddingModel>(
+        &self,
+        model: M,
+    ) -> Result<SqliteVectorIndex<M>, VectorStoreError> {
+        Ok(SqliteVectorIndex::new(model, self.clone()))
+    }
+}
+
+impl VectorStore for SqliteVectorStore {
+    type Q = String;
+
+    async fn add_documents(
+        &mut self,
+        documents: Vec<DocumentEmbeddings>,
+    ) -> Result<(), VectorStoreError> {
+        info!("Adding {} documents to store", documents.len());
+        self.conn
+            .call(|conn| {
+                let tx = conn.transaction().map_err(tokio_rusqlite::Error::from)?;
+
+                for doc in documents {
+                    debug!("Storing document with id {}", doc.id);
+                    // Store document and get auto-incremented ID
+                    tx.execute(
+                        "INSERT OR REPLACE INTO documents (doc_id, document) VALUES (?1, ?2)",
+                        [&doc.id, &doc.document.to_string()],
+                    )
+                    .map_err(tokio_rusqlite::Error::from)?;
+
+                    let doc_id = tx.last_insert_rowid();
+
+                    // Store embeddings
+                    let mut stmt = tx
+                        .prepare("INSERT INTO embeddings (rowid, embedding) VALUES (?1, ?2)")
+                        .map_err(tokio_rusqlite::Error::from)?;
+
+                    debug!(
+                        "Storing {} embeddings for document {}",
+                        doc.embeddings.len(),
+                        doc.id
+                    );
+                    for embedding in doc.embeddings {
+                        let vec = Self::serialize_embedding(&embedding);
+                        let blob = rusqlite::types::Value::Blob(
+                            vec.iter().flat_map(|x| x.to_le_bytes()).collect(),
+                        );
+                        stmt.execute(rusqlite::params![doc_id, blob])
+                            .map_err(tokio_rusqlite::Error::from)?;
+                    }
+                }
+
+                tx.commit().map_err(tokio_rusqlite::Error::from)?;
+                Ok(())
+            })
+            .await
+            .map_err(|e| VectorStoreError::DatastoreError(Box::new(e)))?;
+
+        Ok(())
+    }
+
+    async fn get_document<T: for<'a> Deserialize<'a>>(
+        &self,
+        id: &str,
+    ) -> Result<Option<T>, VectorStoreError> {
+        debug!("Fetching document with id {}", id);
+        let id_clone = id.to_string();
+        let doc_str = self
+            .conn
+            .call(move |conn| {
+                conn.query_row(
+                    "SELECT document FROM documents WHERE doc_id = ?1",
+                    rusqlite::params![id_clone],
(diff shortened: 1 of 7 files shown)


0xMochan:
Hello, @tarrencev, welcome to the Rig community! We appreciate your pull request. I'll be doing a deeper dive and replicating your sqlite setup locally soon but I had a couple of preliminary comments.

1. In the future, make sure you check out the [CONTRIBUTING.md](https://github.com/0xplaygrounds/rig/blob/main/contributing.md) guide. In general, we request all contributions have a corresponding issue made alongside using the PR templates available for new provider and vector store integrations.

2. I was glancing through your PR and I noticed that you've leveraged `zerocopy` for memory transmutation in one of the examples (in the `unsafe` block). This is an unusual dependency and method here being used for an example, is there a specific reason this is needed for the example? I think if you had some documentation on the `rusqlite` side that made this a normal code pattern, it would make it a bit more clear to us!

3. Please do link the specific documentation being used for the vector store implementation and even consider adding it to the code documentation as it makes it convenient for future engineers to be able to maintain this crate!

4. As per your question about #120, we are looking to merge this very soon. In any case, if this PR merges before that one, we can add a follow up commit will be added to #120 post-merge. I'll follow up with the engineer leading that PR to match up what parts are changing related to this PR.

tarrencev:
Hey there, thanks for the response:

> 2. I was glancing through your PR and I noticed that you've leveraged `zerocopy` for memory transmutation in one of the examples (in the `unsafe` block). This is an unusual dependency and method here being used for an example, is there a specific reason this is needed for the example? I think if you had some documentation on the `rusqlite` side that made this a normal code pattern, it would make it a bit more clear to us!

In this case, we were using it to efficiently copy the embedding vector to a bytes representation for storage in the db but we can do it safely with a copy.

> 3. Please do link the specific documentation being used for the vector store implementation and even consider adding it to the code documentation as it makes it convenient for future engineers to be able to maintain this crate!

I've added a reference to the sqlite-vec rust docs for initializing the extension. 

> 4. As per your question about [Feat/embeddings overhaul #120](https://github.com/0xPlaygrounds/rig/pull/120), we are looking to merge this very soon. In any case, if this PR merges before that one, we can add a follow up commit will be added to [Feat/embeddings overhaul #120](https://github.com/0xPlaygrounds/rig/pull/120) post-merge. I'll follow up with the engineer leading that PR to match up what parts are changing related to this PR.

Sounds good!

tarrencev:
The failing test here seems to depend on a api token

0xMochan:
> The failing test here seems to depend on a api token

Looking into it!

rig-sqlite/src/lib.rs
```diff
@@ -0,0 +1,360 @@
...
+                -- Document tables
+                CREATE TABLE IF NOT EXISTS documents (
+                    id INTEGER PRIMARY KEY AUTOINCREMENT,
+                    doc_id TEXT UNIQUE NOT NULL,
+                    document TEXT NOT NULL
+                );
+                CREATE INDEX IF NOT EXISTS idx_doc_id ON documents(doc_id);
+                CREATE VIRTUAL TABLE IF NOT EXISTS embeddings USING vec0(embedding float[1536]);
+
+                COMMIT;",
```
0xMochan, line 35: This query hardcodes a couple of names and values such as `idx_doc_id`, `doc_id`, `1536`, etc. These should be configurable via the `new` (and potentially could be done via a builder for the vector store if you wanted to provide defaults). If you feel some should be hardcoded, then they should be constants but do provide a reason.

For the length of the embeddings, `1536`, we would use `EmbeddingModel::ndims` to configure this. This would mean `new` would also require an `EmbeddingModel` as well (and would need to be generic over one).

Note: in the embeddings revamp, the actual `VectorStore` general trait is gone so that would all collapse into your `SqliteVectorStore`.

rig-sqlite/src/lib.rs
```diff
@@ -0,0 +1,360 @@
...
+                            serde_json::from_str(&doc_str).map_err(|e| {
+                                rusqlite::Error::FromSqlConversionFailure(
+                                    0,
+                                    rusqlite::types::Type::Text,
+                                    Box::new(e),
+                                )
+                            })?;
+                        let vec = bytes
+                            .chunks(4)
+                            .map(|chunk| f32::from_le_bytes(chunk.try_into().unwrap()) as f64)
```
0xMochan, line 173: Rather than using `unwrap`, please opt to use `expect` so that a defined reason can be provided here (even if it's impossible). This makes it easier for future maintainers as it's self-documenting.

rig-sqlite/src/lib.rs
```diff
@@ -0,0 +1,360 @@
...
+
+        let results = self
+            .store
+            .conn
+            .call(move |conn| {
+                let mut stmt = conn.prepare(
+                    "SELECT d.doc_id, e.distance 
+                     FROM embeddings e
+                     JOIN documents d ON e.rowid = d.id
+                     WHERE e.embedding MATCH ?1  AND k = ?2
```
0xMochan, line 336: Extra space between `?1` and `AND`

0xMochan (changes requested):
Looks great! A couple of suggested changes (thanks for adding to the PR after my prelim glance). The `unsafe` block makes sense since it's straight from the docs so it should be fine to users of sqlite (the comment helps a lot!)

w.r.t the earlier `zerocopy` usage, I accidentally confused the usage of `zerocopy` with the `unsafe` block in the example (which, I didn't see why an extra dep. was needed for an example). I see that you reverted this change already (but I don't see it in the commit history) but I think it's actually fine to use it here. Apologies!

rig-sqlite/src/lib.rs
tarrencev: I'm not sure extracting these to be configurable is worth the additional complexity, since the interface encapsulates those details, and imo reduces the readability of the queries when there is a lot of string interpolation going on.

Either way, I think it might be better to save it for the embeddings refactor, since that will impact this pretty significantly from what I can tell.

tarrencev:
Addressed most feedback, if you feel strongly about extracting queries / columns / etc I can try find time to land that too.

0xMochan:
> Addressed most feedback, if you feel strongly about extracting queries / columns / etc I can try find time to land that too.

I do think it would be proper, since the user might be using a sqlite database file that is defined outside of rig. We could use `prepare` and have a sanitized sql query constructed via a builder pattern (since column naming could be theoretically malicious).

If timing is an issue, I could also quickly scaffold or complete this last feedback.

Also, for the `EmbeddingModel`, it might be possible to save only a reference rather than cloning the model since it's only used for `ndim`. This is fairly minor though.

Great work on the feedback, this will be a very handy addition as a general replacement for `InMemoryVectorStore` when I'm handling testing scripts!

tarrencev:
Sounds good! I've updated the PR to pass the embedding model by reference

0xMochan:
Last thing I think is a usage comment on the `SqliteVectorStore` struct. Something similar to Lancedb's: https://github.com/0xPlaygrounds/rig/blob/2cba66272d5af328fa9177c8a1d8b82d276748b7/rig-lancedb/src/lib.rs#L23-L90

We actually do more of this in the #120 (mongodb gets a usage comment).

w.r.t the builder pattern, etc. we'll do a follow-up PR after embeddings get merged.
LGTM 🎉

0xMochan (approved):

rig-sqlite/src/lib.rs
```diff
@@ -0,0 +1,452 @@
...
+}
+
+impl<E: EmbeddingModel + 'static, T: SqliteVectorStoreTable + 'static> SqliteVectorStore<E, T> {
+    pub async fn new(conn: Connection, embedding_model: &E) -> Result<Self, VectorStoreError> {
+        let dims = embedding_model.ndims();
+        let table_name = T::name();
+        let schema = T::schema();
+
+        // Build the table schema
+        let mut create_table = format!("CREATE TABLE IF NOT EXISTS {} (", table_name);
```
0xMochan, line 100: Here, I see you are using normal string interpolation rather than proper SQL [named parameters](https://docs.rs/rusqlite/latest/rusqlite/macro.named_params.html). I saw you used parameters elsewhere in the codebase so I was wondering if there was a specific reason.

It would be best to use proper SQL practices here to avoid SQL injections (even if it would require the app developer to improperly expose fields like `table_name` to the user, I still think it's worthwhile to avoid these bugs).

0xMochan (changes requested):
The comments and updates from the embeddings merge are fantastic. I made one comment about the change with un-hardcoding of the table names and such, w.r.t sql injections. Once that is fixed, this should be ready to merge!

rig-sqlite/src/lib.rs
tarrencev: Hm afaict Sqlite doesn't support parameterized queries for table / column names, only for values

0xMochan: Hmm, I see. Well I suppose we should leave a disclaimer in the documentation, perhaps in the docstring as well.

PR #123, opened 2024-11-25
feat: add dynamic documents

garance-buricatu:
Currently, when an agent needs to execute a vector search, we pass the vector store index as a parameter to the agent (with `dynamic_context`) and the agent executes said vector search internally. 

The problem I am facing with my use case is that the rig app needs access to the results of the vector search separately from the RAG process. Ex:
1. Execute vector search
2. Rig app does stuff with the vector search result
3. Rig app sends vector search results to agent --> agent adds RAGged documents to context window

Right now, step 2 is not possible, unless vector search is executed twice.

### Proposed solution
* Execute the vector search outside of the agent and have the option to pass the results of the vector search to the agent (instead of doing the vector search inside).
* Add `dynamic_documents: Vec<Document>` to the agent


cvauclair:
What is the difference between this and the existing `static_context` documents?

PR #126, opened 2024-11-26
Feat(integration test): MongoDB

marieaurore123:
Uses `testcontainers` to create an integration test which validates vector searching with rig on a local mongodb instance.


0xMochan (approved):
Looks great! `testcontainer` is really straightforward. I would clarify somewhere that a docker service does need to be running for the integration test to work!

rig-mongodb/src/lib.rs
```diff
@@ -27,10 +27,9 @@ struct SearchIndex {
 impl SearchIndex {
     async fn get_search_index(
         collection: mongodb::Collection<bson::Document>,
-        index_name: &str,
     ) -> Result<SearchIndex, VectorStoreError> {
         collection
-            .list_search_indexes(index_name, None, None)
+            .list_search_indexes()
```
cvauclair, line 32: Need to add `.name(...)` to builder, otherwise this will fail if there are multiple indices:
```rust
collection
    .list_search_indexes()
    .name(index_name)
    .await
    ...
```

cvauclair (changes requested):

rig-mongodb/src/lib.rs
cvauclair: See [ListSearchIndexes](https://docs.rs/mongodb/latest/mongodb/action/struct.ListSearchIndexes.html)

PR #148, opened 2024-12-13
docs(ollama-example): implement example showcasing ollama

0xMochan:
This example showcases using [ollama](ollama.com) via a custom endpoint in the `OpenAI` client. Should be accompanied with a blog post showcasing local model support!


cvauclair:
@0xMochan looks good! Is that ready to merge?

0xMochan:
> @0xMochan looks good! Is that ready to merge?

Yea lets do it! A blog post will accompany this next week :)

PR #154, opened 2024-12-17
fix: mongodb integration test

cygaar:
Previously, the mongo integration test was failing with `Error connecting to Search Index Management service.`. This is because it takes a bit of time for the container to initialize the service.

The extra delay added in #153 doesn't quite fix the issue because the index creation happens before the delay.

Our integration tests will now retry a couple of times before attempting to create the index.

### DIFF
--- a/rig-mongodb/tests/integration_tests.rs
+++ b/rig-mongodb/tests/integration_tests.rs
@@ -94,7 +94,7 @@ async fn vector_search_test() {
 async fn create_search_index(collection: &Collection<bson::Document>) {
     let mut attempts = 0;
     let max_attempts = 5;
-    
+
     while attempts < max_attempts {
         match collection
             .create_search_index(
@@ -115,13 +115,19 @@ async fn create_search_index(collection: &Collection<bson::Document>) {
         {
             Ok(_) => return,
             Err(_) => {
-                println!("Waiting for MongoDB to be ready... Attempts remaining: {}", max_attempts - attempts - 1);
+                println!(
+                    "Waiting for MongoDB to be ready... Attempts remaining: {}",
+                    max_attempts - attempts - 1
+                );
                 sleep(Duration::from_secs(5)).await;
                 attempts += 1;
             }
         }
     }
-    panic!("Failed to create search index after {} attempts", max_attempts);
+    panic!(
+        "Failed to create search index after {} attempts",
+        max_attempts
+    );
 }
 
 async fn bootstrap_collection(host: String, port: u16) -> Collection<bson::Document> {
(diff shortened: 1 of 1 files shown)


cvauclair:
Hey @cygaar thanks for the help in fixing this very annoying integration test! Retrying until the search index has been created is indeed the right way of fixing this. However, `create_search_index` doesn't actually return an error if the search index is not ready (so your loop will always exit on the first iteration). 

Instead, I think the way to do this is to poll the `list_search_indices` function until the search index has `status == READY`. Something like:
```rust
while attempts < max_attempts {
    match collection
        .list_search_indexes()
        .name(index_name)
        .await?
        .with_type::<rig_mongodb::SearchIndex>()
        .next()
     {
         Some(index) if index.status == "READY" => return,
         Some(_) => continue
         None => panic!() 
     }
}
```

cygaar:
I'll fix shortly!

cygaar:
@cvauclair I'm pretty sure the logic I have here works - I added a log to check if the index was created and it seems to work.
I checked the definition of `create_search_index` and it does return a value.

Screenshot 2024-12-17 at 12 27 40 PM

cvauclair:
FYI I had increased the delay further to 30s in another PR that has since been merged, but feel free to reduce it back to 5s (or remove it entirely) once the polling for READY status is done!

cygaar:
I will note that the creation of the search index is what's causing the original error - we would essentially need to keep the logic i currently have and add an additional check to make sure the index was created

cygaar:
@cvauclair i think the last test failure is related to missing api keys - not entirely sure how you guys have github actions setup to get that to work

rig-mongodb/tests/integration_tests.rs
```diff
@@ -92,6 +93,65 @@ async fn vector_search_test() {
...
+        {
+            Ok(_) => {
+                // Wait for index to be available
+                for _ in 0..max_attempts {
+                    let indexes = collection
+                        .list_search_indexes()
+                        .await
+                        .unwrap()
+                        .collect::<Vec<_>>()
+                        .await;
```
cvauclair, line 125: Tip: Instead of fetching all indices, you can filter by name and get only one:
```rust
let maybe_index = collection
    .list_search_indexes()
    .name(VECTOR_SEARCH_INDEX_NAME)
    .await
    .unwrap()
    .next()
    .await
```

rig-mongodb/tests/integration_tests.rs
```diff
@@ -92,6 +93,65 @@ async fn vector_search_test() {
...
+
+                    if indexes.iter().any(|idx| {
+                        idx.as_ref()
+                            .ok()
+                            .and_then(|i| i.get_str("name").ok())
+                            .map_or(false, |name| name == VECTOR_SEARCH_INDEX_NAME)
```
cvauclair, line 131: You need to check that the "status" field is "READY", otherwise this will succeed even if the index is in an invalid state

cvauclair (changes requested):

PR #158, opened 2024-12-19
fix: tokio-rusqlite to crates.io instead of github

edisontim:


0xMochan (approved):
Great catch! @tarrencev do you remember why this was a github dependency in the first place?

cvauclair (approved):

PR #168, opened 2024-12-26
fix: anthropic tool use

edisontim:
anthropic currently sends back a vec on `Content` structs. Before the fix we were matching on the first element of that. The first element of a ToolUse is rarely ever the tool but more often text so we were never actually getting the tool use as a CompletionResponse

The PR first tries to find a tool use in the vec then some text

### DIFF
--- a/rig-core/src/agent.rs
+++ b/rig-core/src/agent.rs
@@ -265,7 +265,7 @@ impl<M: CompletionModel> Chat for Agent<M> {
                 ..
             } => Ok(msg),
             CompletionResponse {
-                choice: ModelChoice::ToolCall(toolname, args),
+                choice: ModelChoice::ToolCall(toolname, _, args),
                 ..
             } => Ok(self.tools.call(&toolname, args.to_string()).await?),
         }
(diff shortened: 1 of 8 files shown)


0xMochan:
Thanks for this PR! I think this is something that plagues all providers atm, but it seems like it affects Anthropic a bit more thoroughly. In general, we try to encourage tracking issues to be made before a PR is written so that we can properly understand the bug (it also helps in future maintenance). In this situation, I'd wonder if this bug fix should be applied for every provider.

edisontim:
I've only every used anthropic for my project so haven't noticed it happening in other providers but can change that yeah

kloudsamurai (commented):
Thank you very much for tackling this. I really like Rig but was trying to figure out tools and got snagged here as well. is there a reason `tool_use` only supports 3 items? would it be more applicable to support n number of tools?

edisontim:
> Thank you very much for tackling this. I really like Rig but was trying to figure out tools and got snagged here as well. is there a reason `tool_use` only supports 3 items? would it be more applicable to support n number of tools?

Maybe I didn't understand this correctly, it supports n tools currently.

Also the ID I've added is for anthropic `tool_result` response. See [here](https://docs.anthropic.com/en/docs/build-with-claude/tool-use#how-tool-use-works)

cvauclair (approved):

cvauclair:
Hey @edisontim sorry for the delay, the new year has been super busy for us! 

We're ready to merge your PR! Just need to address the failing CI checks and rebase on main

cvauclair (approved):

PR #169, opened 2024-12-26
feature - update gemini client to support proxy

nihao:
- update gemini client to support proxy
```rust
    pub fn with_proxy(mut self, proxy_url: &str) -> Self {
        let proxy = reqwest::Proxy::all(proxy_url).expect("Proxy URL should be valid");
        self.http_client = reqwest::Client::builder()
            .proxy(proxy)
            .build()
            .expect("Gemini reqwest client should build");
        self
    }
```

### DIFF
--- a/rig-core/src/providers/gemini/client.rs
+++ b/rig-core/src/providers/gemini/client.rs
@@ -43,6 +43,15 @@ impl Client {
         }
     }
 
+    pub fn with_proxy(mut self, proxy_url: &str) -> Self {
+        let proxy = reqwest::Proxy::all(proxy_url).expect("Proxy URL should be valid");
+        self.http_client = reqwest::Client::builder()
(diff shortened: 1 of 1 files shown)


0xMochan (commented):
Thanks for this PR. Is there a specific reason proxy URLs should be used only with Gemini, shouldn't this be available globally? We generally try to avoid PRs without associated issues so it would be best if there was a feature request issue made that describes this feature and why it's neccessary!

cvauclair:
Hey @nihao, thanks for the contribution! 

Forgive me if this is an obvious question, but is there a reason to use `with_proxy` as opposed to just setting a different base URL using `gemini::Client::from_url("API_KEY", "PROXY_URL")`?

nihao:
ok. pls reject this commit

PR #170, opened 2024-12-27
chore: fix types

qwdsds:
Fix spelling of words


0xMochan (approved):
Good catches in the codebase! btw, this would be a `docs: ` prefix for the squashed message name!

PR #171, opened 2024-12-27
feat: Decentralized Inference for rig

genesis-0000:
Add onchain-verifiable, decentralized inference for rig with onchain Llama, onchain Hermes, and onchain Intellect.


0xMochan (changes requested):
Hey, we appreciate this integration in this PR! Wanna quickly note that the failing test case is not due to this project, but due to the API keys to test other providers are not available on your github repo. This is an oversight in our intergration testing on our end that we'll address Q1 '25.

You will have to address the failing `fmt` CI, checkout our [CONTRIBUTING.md](https://github.com/0xPlaygrounds/rig/blob/main/CONTRIBUTING.md) for more instructions!

---

Make sure you include an example in the `examples` repo. A good one to copy is the `agent_with_grok.rs` one as it tests every major feature. I tested this locally and got basic completion working, but could not get any tools to work, might need to check that.

It seems like a lot of this API sits on top of the openai integration. It might be suitable to actually reuse a lot of the openai models and implementations. Theoretically, you could add extra params similar to how `additional_params` works for the `chain_id` and the response `onchain_data`. I'm not entirely certain if the goal of this platform is for direct openai compatibility or if they plan deviating in the future, but at the very least, some set of openai models could be used to simplify this integration.

tl;dr
- Address the changes I've suggested
- Consider some amount of OpenAI reuse
- Add an example (based on `agent_with_grok`)
- Test all features (basically just have `agent_with_grok` but run with eternalai).

Let me know if you have any specific questions are if you need help in any step, I'd be happy to help! Also feel free to reach out to me on our Discord as well!

0xMochan:
Also, quick notes:
- Our next release will be after the holidays, when the whole team is active again so there won't be a release until the first week of January.
- In the future, please make an [issue](https://github.com/0xPlaygrounds/rig/issues/new/choose) before creating a PR so we can better keep track of the feature, thanks!

0xMochan:
@genesis-0000 Hey, I've seen a lot of updates here, is this ready for re-review?

0xMochan (approved):

0xMochan:
Hey @genesis-0000 , could you rebase to main. Once we do that, we can merge this

PR #175, opened 2024-12-31
feat: add send sync for CF worker

ronaldslc:
This enables rig to work on CF workers which opens up a whole new dimension of use case to integrate with web based UI such as leptos.

The simple change is put behind a feature flag.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ target/
 .DS_Store
 .idea/
 .vscode/
+.devcontainer/
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4824,6 +4824,7 @@ dependencies = [
  "tokio-test",
  "tracing",
  "tracing-subscriber",
+ "worker",
 ]
 
 [[package]]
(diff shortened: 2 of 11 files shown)


cvauclair:
Hey @ronaldslc, thanks for the contribution! 

Quick question to make sure I understand correctly: Does that new feature flag enable Rig to be compiled to WASM in any context? Or only within the context of CF workers?

ronaldslc:
Hi @cvauclair, that's a good question, the feature flag is intended to allow rig to work with CF workers only using axum support for CF workers.

More specifically the macro was added in CF worker-rs crate due to this issue https://github.com/cloudflare/workers-rs/issues/485

My understanding is JsFuture is `!send` in wasm_bindgen, so when rig.rs async functions are converted to wasm, the structure is not send, but axum and many frameworks requires send futures, which creates an intermediate compilation issue. So I would say this is a compatibility layer to have certain wasm compatible Rust framework, work with rig.rs. Not necessarily worker specific.

Best regards

cvauclair:
Thanks for the reply @ronaldslc, makes a lot of sense! 

We'll go ahead and merge this, but we'll do some testing on our end to see how general that feature is. Be aware, if it does end up being a general WASM compatibility layer we'll probably change the name of the feature flag to something like `wasm` to better reflect the capabilities.

Cheers!

cvauclair (approved):

omarabid:
This adds a whole foreign dependency (worker) just to add a `!Send` wrapper. First, `worker` is specific to Cloudflare workers and thus might break other Wasm builds. Second, Wasm might introduce multi-threading in the future and the !Send implementation is `unsafe`.

Library consumers can wrap the types/futures themselves and make it !Send. `workers` already provide the tooling for that.

PR #178, opened 2025-01-01
Fix 404 POST response message from Ollama due to wrong endpoint url

wale-e:
The current ollama URL throws an 404 (not found) error message. This PR fixes the bug


0xMochan (approved):
Looks good!

PR #184, opened 2025-01-06
fix: anthropic documents in prompt

### LINKED ISSUE #183: bug: document formatting when using Anthropic

## Bug Report

Anthropic expects documents to be structured in a [particular XML](https://docs.anthropic.com/en/docs/build-with-claude/prompt-engineering/long-context-tips#example-multi-document-structure) format. This is currently not the case - the documents are wrapped in `<attachments>` tags.

## Expected behavior

When using an Anthropic model, I expect the request from Rig to be formatted correctly according to what the Anthropic API expects and recommends.

## Additional context

The issue might be resolved once #149 is addressed.

milancermak:
Resolves #183

### DIFF
--- a/rig-core/CHANGELOG.md
+++ b/rig-core/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - *(openai)* Make integration more general (#156)
+- *(anthropic)* Properly format context documents in prompt (#183)
 
 ### Other
 
--- a/rig-core/src/providers/anthropic/completion.rs
+++ b/rig-core/src/providers/anthropic/completion.rs
@@ -197,7 +197,20 @@ impl completion::CompletionModel for CompletionModel {
         // specific requirements of each provider. For now, we just manually check while
         // building the request as a raw JSON document.
 
-        let prompt_with_context = completion_request.prompt_with_context();
+        let prompt_with_context = if completion_request.documents.is_empty() {
+            completion_request.prompt.clone()
+        } else {
+            let documents = completion_request.documents.into_iter().enumerate().map(|(i, doc)| {
+                format!(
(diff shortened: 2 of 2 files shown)


milancermak:
Rebased with main

0xMochan:
This form of document serialization sorta bypasses the general pattern that is used via the `Display` trait for all `Document`s, tbh, it might make sense to use anthropic's patterns as the way to serialize for all platforms. I chose the current method arbitrarily.

Specifically for [anthropic](https://docs.anthropic.com/en/api/messages), it seems like you can (_now_?) pass documents as a type of content in a user message. It might be more suitable to use this method (albeit, contradicting with it's own user guides?) for passing documents to anthropic? Curious if you saw this and had thoughts?

milancermak:
> pass documents as a type of content in a user message

It seems that the API only supports PDFs in this case - the API docs only show `application/pdf` as the available option for `messages.content.source.media_type` 

Screenshot 2025-01-09 at 12 19 27

BTW I watch [this video on prompting Claude](https://www.youtube.com/watch?v=hkhDdcM5V94) yesterday. The guy mentions that Claude understand XML better (because apparently it has been trained on a lot of XML 🤷 ) and it doesn't necessarily matter what tags you use as long as they are meaningful and separate stuff in the prompt (so it's also a good idea to use sth like <instructions>). I'm writing all that just to say that the original <attachments> you used isn't bad at all 😄

joshua-mo-143:
It looks like the test failed because we're missing an API key.

cc @0xMochan

0xMochan:
> It looks like the test failed because we're missing an API key.
> 
> cc @0xMochan

This PR needs to rebase / merge w/ main for the CI/CD to update.

joshua-mo-143:
Closing as there has been no update from the contributor.

@milancermak If this PR is still something you'd like to contribute we'd be happy to take it!

PR #188, opened 2025-01-08
feat: add Galadriel API integration

kristjanpeterson1:
Add a basic [Galadriel](https://docs.galadriel.com) LLM integration

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,16 @@
 version = 4
 
 [[package]]
+name = "Inflector"
+version = "0.11.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fe438c63458706e03479442743baae6c88256498e6431708f6dfc520a26515d3"
+dependencies = [
+ "lazy_static",
+ "regex",
+]
+
+[[package]]
 name = "addr2line"
 version = "0.24.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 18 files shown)


0xMochan:
Hello! Welcome to the rig community, we appreciate your PR a ton! It looks like this integration is nearly identifical to the openai provider, if the API is based on the OpenAI spec it might be more fitting to just use the openai provider client instead of having a duplicate provider. I do see that an extra client method was added so it might be alright to have a singular method for instantiating an openai client w/ a custom reqwest client.

Also, diving into the documentation for Galadriel, it seems like the crux of this platform is to provide verifiable / on-chain attestations of LLM output. This seems like a much deeper integration in order to provide that if we were to match how the `sentience` library works on their end, atleast in implementing the TEE.

If it's a simple provider integration, it's likely not needed to maintain an entire provider if you can reuse the openai client. With a larger TEE implementation, it might require a deeper look on how that platform works. I think it would be best if an issue was made so that we can explore this in more detail, perhaps discussing with the proper teams, etc. Let me know if this makes sense, or I can follow up with any clarifications!

kristjanpeterson1:
@0xMochan Yep it is very similar to the OpenAI provider, but the Galadriel one does not support embedding models, making it kind of incorrect to use the openai provider? 

The main part of the `sentience` library is still just LLM inference, users can also fetch their inferences and prove their LLM inputs/outputs, but that would usually happen on some other platform, eg some frontend to showcase these, so I dont think it makes sense to think about that part here. 

So overall I think it should be just a provider integration, but I am not sure if it could be simplified to remove duplications. Do you have any suggestions here?

kristjanpeterson1:
@0xMochan Thats kind of what I did, or do you have something else in mind?

cvauclair (changes requested):
Looks good! 

Only thing required so we can merge this would be to pull the latest updates from the `main` branch, which includes fixes to our testing suite. Wanna make sure this passes CI before merging.

Sorry for the delay, it's been a challenge to keep up with all the contributions. Let's get this merged! Cheers!

kristjanpeterson1:
@cvauclair Merged and fixed a small issue! Should be ok now :)

joshua-mo-143:
Lgtm - thanks for the PR @kristjanpeterson1!

PR #195, opened 2025-01-11
feat: Agent debug trait

augustin-v:
Implement the `Debug` trait for `Agent`. Omitted the `model` field due to potential Debug implementation limitations for `CompletionModel`.


cvauclair:
Hey @augustin-v thanks for the PR! Could you provide an example output of what this Debug trait impl looks like for a fully customized Agent? Thanks!

PR #200, opened 2025-01-14
Add conditional op

AdrielC:
This PR introduces two macros, `conditional!` and `try_conditional!`, for dispatching to different pipeline operations based on which variant of a single-parameter enum is given. It’s similar to an “ArrowChoice” concept for Rust: we have a branching operator that calls different sub-ops depending on the variant. This makes it easier to handle “fan-in” logic in a single pipeline.

### Changes

- Added conditional! macro, generating an Op impl that switches on enum variants.
- Added try_conditional! macro, generating a TryOp impl that does the same but returns a Result.
- Documented requirements:
  - The enum must be a single-type-parameter wrapper (e.g., enum MyEnum<T> { VariantA(T), VariantB(T) }).
  - All sub-ops must share the same input and same output.


cvauclair (changes requested):
Nice one @AdrielC ! 

If you could merge main into your branch (to get the latest non-broken tests) as well as run `cargo fmt` (to pass the CI style checker errors) then we can merge this. 

Cheers!

cvauclair (approved):

PR #205, opened 2025-01-16
feat(model): add support for EternalAI onchain toolset

eternal-ai-org:
Many agents are currently offchain programs. The Eternal AI team is implementing an Onchain Agent Transformer that takes any agent and turn it into a fully onchain EVM smart contract.

After being tranformed, agents will live forever on the blockchain. Currently, Eternal AI is supporting 14+ different blockchains.

Because transformed agents are onchain, they are unstoppable, censorship-resistant, tamper-proof, and permissionless accessible by anyone.

Base on #206


rig-core/src/providers/eternalai.rs
```diff
@@ -470,6 +538,37 @@ impl completion::CompletionModel for CompletionModel {
             vec![]
         };
 
+        println!("Try to get on-chain system prompt");
```
cvauclair, line 541: Should use `tracing::info!` or `tracing::debug!` macro to log things

rig-core/src/providers/eternalai.rs
```diff
@@ -470,6 +538,37 @@ impl completion::CompletionModel for CompletionModel {
...
+        let eternal_ai_agent_id =
+            std::env::var("ETERNALAI_AGENT_ID").unwrap_or_else(|_| "".to_string());
+        if !eternal_ai_rpc.is_empty()
+            && !eternal_ai_contract.is_empty()
+            && !eternal_ai_agent_id.is_empty()
+        {
+            println!(
+                "get on-chain system prompt with {}, {}, {}",
+                eternal_ai_rpc, eternal_ai_contract, eternal_ai_agent_id
+            );
```
cvauclair, line 554: Use `tracing` macros

rig-core/src/providers/eternalai.rs
```diff
@@ -470,6 +538,37 @@ impl completion::CompletionModel for CompletionModel {
...
+            let c_value: c_uint = eternal_ai_agent_id.parse::<u32>().unwrap_or(0);
+            let prompt =
+                get_on_chain_system_prompt(&eternal_ai_rpc, &eternal_ai_contract, c_value).await;
+            match prompt {
+                None => {
+                    println!("on-chain sytem prompt is none")
```
cvauclair, line 560: Use `tracing` macros

rig-core/src/providers/eternalai.rs
```diff
@@ -505,6 +604,8 @@ impl completion::CompletionModel for CompletionModel {
             })
         };
 
+        println!("request: {:?}", request.to_string());
```
cvauclair, line 607: Use `tracing` macro (or remove entirely if not needed anymore)

rig-core/src/providers/eternalai.rs
```diff
@@ -331,6 +337,68 @@ pub fn get_chain_id(key: &str) -> Option<&str> {
...
+    let decoded_strings: Vec<String> = system_prompts
+        .iter()
+        .map(|bytes| {
+            String::from_utf8(bytes.to_vec()).unwrap_or_else(|_| "[Invalid UTF-8]".to_string())
+        })
+        .collect();
+
+    if !decoded_strings.is_empty() {
+        let prompt = decoded_strings[0].clone();
+        println!("system prompt : {}", prompt);
```
cvauclair, line 371: Should use `tracing::info!` or `tracing::debug!` macros to log things

rig-core/src/providers/eternalai.rs
```diff
@@ -331,6 +337,68 @@ pub fn get_chain_id(key: &str) -> Option<&str> {
...
+}
+
+pub async fn fetch_on_chain_system_prompt(content: &str) -> Option<String> {
+    if content.contains(IPFS) {
+        let light_house = content.replace(IPFS, LIGHTHOUSE_IPFS);
+        println!("light_house : {}", light_house);
+        let mut response = get(light_house).await.unwrap();
+        if response.status().is_success() {
+            let body = response.text().await.unwrap();
+            println!("light_house body: {}", body);
```
cvauclair, line 384: Use `tracing` macros

rig-core/src/providers/eternalai.rs
```diff
@@ -331,6 +337,68 @@ pub fn get_chain_id(key: &str) -> Option<&str> {
...
+            return Some(body);
+        } else {
+            let gcs = content.replace(IPFS, GCS_ETERNAL_AI_BASE_URL);
+            println!("gcs: {}", gcs);
```
cvauclair, line 388: Use `tracing` macros

rig-core/src/providers/eternalai.rs
```diff
@@ -331,6 +337,68 @@ pub fn get_chain_id(key: &str) -> Option<&str> {
...
+            response = get(gcs).await.unwrap();
+            if response.status().is_success() {
+                let body = response.text().await.unwrap();
+                println!("gcs body: {}", body);
```
cvauclair, line 392: Use `tracing` macros

rig-core/src/providers/eternalai.rs
```diff
@@ -331,6 +337,68 @@ pub fn get_chain_id(key: &str) -> Option<&str> {
...
+        SystemPromptManagementContract,
+        r#"
+        [{"inputs": [{"internalType": "uint256", "name": "_agentId", "type": "uint256"}], "name": "getAgentSystemPrompt", "outputs": [{"internalType": "bytes[]", "name": "","type": "bytes[]"}], "stateMutability": "view", "type": "function"}]
+        "#
+    );
+    // Connect to an Ethereum node
+    let provider = Provider::<Http>::try_from(rpc_url).expect("Failed to parse url");
+    let client = Arc::new(provider);
+    let contract_address: Address = contract_addr.parse().expect("invalid contract address");
+    let contract = SystemPromptManagementContract::new(contract_address, client);
```
cvauclair, line 355: Would be nice to store all those values in some sort of `OnChainClient` struct so we don't have to initialize them every time the function is called

rig-core/src/providers/eternalai.rs
```diff
@@ -331,6 +337,68 @@ pub fn get_chain_id(key: &str) -> Option<&str> {
...
+    agent_id: c_uint,
+) -> Option<String> {
+    abigen!(
+        SystemPromptManagementContract,
+        r#"
+        [{"inputs": [{"internalType": "uint256", "name": "_agentId", "type": "uint256"}], "name": "getAgentSystemPrompt", "outputs": [{"internalType": "bytes[]", "name": "","type": "bytes[]"}], "stateMutability": "view", "type": "function"}]
+        "#
+    );
+    // Connect to an Ethereum node
+    let provider = Provider::<Http>::try_from(rpc_url).expect("Failed to parse url");
```
cvauclair, line 352: The function should return a `Result::Err` instead of using `.expect` which can panic

rig-core/src/providers/eternalai.rs
```diff
@@ -331,6 +337,68 @@ pub fn get_chain_id(key: &str) -> Option<&str> {
...
+    let client = Arc::new(provider);
+    let contract_address: Address = contract_addr.parse().expect("invalid contract address");
```
cvauclair, line 354: Return a `Result::Err` instead of panicking

rig-core/src/providers/eternalai.rs
```diff
@@ -331,6 +337,68 @@ pub fn get_chain_id(key: &str) -> Option<&str> {
...
+    let contract = SystemPromptManagementContract::new(contract_address, client);
+    let system_prompts: Vec<Bytes> = contract
+        .get_agent_system_prompt(U256::from(agent_id))
+        .call()
+        .await
+        .expect("invalid agent system prompt");
```
cvauclair, line 360: Return a `Result::Err` instead of panicking

rig-core/src/providers/eternalai.rs
```diff
@@ -331,6 +337,68 @@ pub fn get_chain_id(key: &str) -> Option<&str> {
...
+        println!("system prompt : {}", prompt);
+        return fetch_on_chain_system_prompt(&prompt).await;
+    }
+    None
+}
+
+pub async fn fetch_on_chain_system_prompt(content: &str) -> Option<String> {
+    if content.contains(IPFS) {
+        let light_house = content.replace(IPFS, LIGHTHOUSE_IPFS);
+        println!("light_house : {}", light_house);
```
cvauclair, line 380: Use` tracing` macros

rig-core/src/providers/eternalai.rs
```diff
@@ -331,6 +337,68 @@ pub fn get_chain_id(key: &str) -> Option<&str> {
...
+
+    if !decoded_strings.is_empty() {
+        let prompt = decoded_strings[0].clone();
+        println!("system prompt : {}", prompt);
+        return fetch_on_chain_system_prompt(&prompt).await;
+    }
+    None
+}
+
+pub async fn fetch_on_chain_system_prompt(content: &str) -> Option<String> {
```
cvauclair, line 377: Maybe rename to `fetch_from_ipfs`? IMO this would be both more accurate (we're specifically fetching from IPFS, not any "onchain") and more general (we can fetch anything, not just system prompts)

cvauclair (changes requested):
Nice work! Please take a look at my comments. 

Additionally, since this PR adds `ethers` as a dependency, it would make sense to split out this integration as a companion crate `rig-eternalai` so as not to aadd more dependencies to the main `rig-core` crate.

Cheers!

cvauclair (changes requested):
Would it be possible to also move the `providers/eternalai.rs` to `rig-eternalai`? Right now using `rig-core` automatically depends on `rig-eternalai` which depends on `ethers` so it doesnt really solve the "keep `rig-core` lean" problem.

The example should also be moved to the sub crate. Thanks!

cvauclair:
Hey @eternal-ai-org! The `prompt_with_context` method is getting removed in the next big Rig update and replaced by a method for each provider (since different providers might expect context documents in different formats). So what I would suggest is to implement it in the provider module for `eternalai` (if you want the same behavior you can simply copy the method as is).

As for the rest, just to make sure I'm clear, this is ideally what I would like to see in the sub-crate:
```bash
rig-eternalai/
  examples/
    agent_with_eternalai.rs
  src/
    lib.rs
    eternalai_system_prompt_manager_toolset.rs
    client.rs (moved from rig-core/src/providers/eternalai.rs)

```
Also, `rig-core` should **not** depend on `rig-eternalai`.

Finally, some docstrings could be useful for the onchain functions ;)

Cheers, let's get this merged!

rig-eternalai/src/providers/eternalai.rs
```diff
@@ -471,6 +471,47 @@ impl completion::CompletionModel for CompletionModel {
             vec![]
         };
 
+        println!("Try to get on-chain system prompt");
```
cvauclair, line 474: Change to `tracing::info!` or `tracing::debug!`

cvauclair (changes requested):
Hey @eternal-ai-org , sorry for the delay, there's a very small thing to change but after that this should be good to! Cheers!

cvauclair (approved):

PR #215, opened 2025-01-18
docs: agent recipes

joshua-mo-143:
Adds recipe agent examples from [the agent recipes website](https://www.agentrecipes.com/) inspired by Anthropic, but in Rust.

Some of the examples are not directly from the recipes website.

### DIFF
--- a/rig-core/examples/agent_autonomous.rs
+++ b/rig-core/examples/agent_autonomous.rs
@@ -1,13 +1,6 @@
-use std::env;
-use rig::{
-    embeddings::EmbeddingsBuilder,
-    parallel,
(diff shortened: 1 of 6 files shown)


0xMochan (approved):
These are awesome! Could go alongside a blog post or something to talk about them in more detail!

In the future, I'd like to refactor our examples a bit to clean them up, maybe add some folders or something

joshua-mo-143:
Yeah for sure, I think a blog post would be the ideal way to present these. Can scope this in probably sometime during w/c Feb 10 as I am highly likely to be quite busy

PR #217, opened 2025-01-19
fix: mongodb vector search example

### LINKED ISSUE #216: bug: vector_search_mongodb fails with `invalid type: map, expected a string`

## Bug Report
In `vector_search_mongodb`, `index.top_n` (line 92),  fails with `invalid type: map, expected a string`.
That is because the doc contains `"_id": Object {"$oid": String("678ccaea8019e2068d869375")}`, which is an object instead of a String as expected.

## Reproduction
```
cd rig-mongodb
cargo run --example vector_search_mongodb --features=rig-core/derive
```

## Expected behavior
The example should run.

dandagow3t:
Resolves #216


joshua-mo-143 (approved):
lgtm

PR #220, opened 2025-01-19
feat: Add support for DeepSeek

wale-e:
feat: Add support for DeepSeek

### DIFF
--- a/rig-core/src/providers/deepseek.rs
+++ b/rig-core/src/providers/deepseek.rs
@@ -1,0 +1,216 @@
+//! DeepSeek API client and Rig integration
+//!
+//! # Example
+//! ```
+//! use rig::providers::deepseek;
+//!
+//! let client = deepseek::Client::new("DEEPSEEK_API_KEY");
+//!
+//! let deepseek_chat = client.completion_model(deepseek::DEEPSEEK_CHAT);
+//! ```
+use crate::{
+    completion::{
+        CompletionModel, CompletionRequest, CompletionResponse
+    },
+    json_utils,
+};
+use reqwest::Client as HttpClient;
+use serde::Deserialize;
+use serde_json::json;
+
+// ================================================================
+// Main DeepSeek Client
+// ================================================================
+const DEEPSEEK_API_BASE_URL: &str = "https://api.deepseek.com";
+
+
+#[derive(Clone)]
+pub struct Client {
+    pub base_url: String,
+    pub api_key: String,
+    http_client: HttpClient,
+}
+
+impl Client {
+    // Create a new DeepSeek client from an API key.
+    pub fn new(api_key: &str) -> Self {
+        Self {
+            base_url: DEEPSEEK_API_BASE_URL.to_string(),
+            api_key: api_key.to_string(),
+            http_client: HttpClient::new(),
+        }
+    }
+
+    // If you prefer the environment variable approach:
+    pub fn from_env() -> Self {
+        let api_key = std::env::var("DEEPSEEK_API_KEY").expect("DEEPSEEK_API_KEY not set");
+        Self::new(&api_key)
+    }
+
+    // Handy for advanced usage, e.g. letting user override base_url or set timeouts:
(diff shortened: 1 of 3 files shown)


kloudsamurai (commented):
Firstly, thank you! I also worked on this and have a different cut but i think yours super easy to mod if you agree. I think we want the `/beta` endpoint right? Two of my favorites features are in their  `Chat Prefix Completion` + `FIM` ftw:

https://api-docs.deepseek.com/guides/fim_completion
https://api-docs.deepseek.com/guides/chat_prefix_completion

Also recommend we set default temp to 0. 90% of avg using for code gen and that's their specific recommendation. (super minor and opinionated, just seeing what you think).

I don't see `json_object` which is awesome. I only use that version. 
https://api-docs.deepseek.com/guides/json_mode

And they do support tools / function calling:
https://api-docs.deepseek.com/guides/function_calling

Happy to clean up my branch or branch off of yours if the team wants those features. I feel like it's worth for this model which huge for me ATM.

wale-e:
Thanks, site says function calling is currently unstable. Don't want to added that at this point. 
image

wale-e:
This is a first cut that integrates the LLM. If approved, then enhancements can be suggested to the team and a PR raised to address. It is always best to keep each PR focused and small in my experience. Thanks for the review and feedback.

rig-core/src/providers/deepseek.rs
```diff
@@ -0,0 +1,210 @@
...
+            let role = match msg.role.as_str() {
+                "system" => "system",
+                "assistant" => "assistant",
+                _ => "user",
+            };
+            messages_json.push(json!({
+                "role": role,
+                "content": msg.content,
+            }));
+        }
```
cvauclair, line 136: This currently places the chat history _after_ the current prompt. It should be the opposite, i.e.: the current prompt should be the last message of the array (see this [example](https://api-docs.deepseek.com/guides/multi_round_chat) from the DeepSeek docs).

In general, the order is always: `[system_message, ..chat_history, new_prompt]`

cvauclair (changes requested):
Thanks for the contribution! There is a small bug with the order of messages in the request but apart from that this looks very good!

rig-core/src/providers/deepseek.rs
wale-e: Thanks for the review and finding the bug, I have pushed a fix for the message order.

kloudsamurai:
> This is a first cut that integrates the LLM. If approved, then enhancements can be suggested to the team and a PR raised to address. It is always best to keep each PR focused and small in my experience. Thanks for the review and feedback.

I have experienced that loop or sure. Had to add some protection for it .

cvauclair (approved):

rig-core/src/providers/deepseek.rs
```diff
@@ -0,0 +1,210 @@
...
+            raw_response: json_resp,
+        })
+    }
+}
+
+// ================================================================
+// DeepSeek Completion API
+// ================================================================
+/// `deepseek-chat` completion model
+pub const DEEPSEEK_CHAT: &str = "deepseek-chat";
```
flyq, line 210: how about add deepseek-reasoner completion model, it use the R1 model:

```rust
pub const DEEPSEEK_REASONER: &str = "deepseek-reasoner";
```

flyq (commented):
great work

kloudsamurai:
@wale-e :: thank you, very much!

@cvauclair :: thanks for the feedback and pointer to `additional_params`. I tend to prefer a trait in Rust since Rust is awesome like that. If we add some provider specific to the Deepseek provider I'd use that implementation detail behind a more intuitive trait -- that match your thinking?

re: FIM -- did you see it as an internal detail the Deepseek client would implement behind a completion transparently to the caller? (FIM superior to completion given context of inference) or as an explicit only operation  scoped to this provider? It would seem to make sense to have a trait for FIM we could use for Codestral etc. I'd be happy to follow up with PR once I see what you're thinking -- if this is a desirable enhancement.

PR #225, opened 2025-01-21
feat: Fix broken link to Conventional Commits doc

mdqst:
I noticed a small in the docwhere the link to the Conventional Commits website was missing the `https://` protocol. This caused the link to be broken. I’ve added the missing `https://` to ensure the link works correctly.  

Here’s the corrected line:  
```markdown  
Commit messages should follow the [Conventional Commit](https://conventionalcommits.org/en/v1.0.0) format (prefixing with `feat`, `fix`, etc.) as this integrates into our auto-releases via a [release-plz](https://github.com/MarcoIeni/release-plz) Github action.  
```  

LFG!


cvauclair (approved):
Good catch @mdqst , thanks for the fix!

PR #231, opened 2025-01-22
feat: postgres vector store

### LINKED ISSUE #5: feat: Add support for PostgreSQL vector store
## Feature Request
Add vector store support for PostgreSQL (with `pg_vector` extension).

### Motivation
PostgreSQL is one of the most widely used open source databases and supports vector search via the [`pg_vector`](https://github.com/pgvector/pgvector) extension. This makes the database a good integration candidate for Rig.

### Proposal
- Create a new companion crate in the repository called `rig-postgres`.
- Implement the `VectorStore` and `VectorStoreIndex` traits for the selected PostgreSQL client.

Note: The integration should be similar to the MongoDB vector store integration (see [rig-mongodb](https://github.com/0xPlaygrounds/rig/blob/main/rig-mongodb/src/lib.rs)).

carlos-verdes:
Resolves #5 based on PR #157

@milancermak mentioned not having time to finish his PR so I created this one based on his implementation.

Current code has just the library code but I added an integration test based on rig-qdrant module, it starts a Docker container with Postgres + PgVector and it simulate calls to OpenAI using a mocked API.

Difference from previous PR:
- removed static lifetime
- the documents are stored as json in the database so you can use any type that implements `Serializable`, `Deserializable`
- support for more than one embedding per document (this also is not supported in rig-qdrant that only takes first vector), in the case more than one result hit the same document only nearest one is returned.
- use sqlx instead of tokio-postgres
- creation of database is outside the code and I provide example of setup in integration test (using sqlx migrations)
- you can use any distance filter supported by PgVector (is not hardcoded to cosine)
- Created example that can be launch using `make run` (from rig-postgres folder), load environmental variables from `.env` file and handles documents with more than one embedding. It also runs migrations automatically on the database to make sure Postgres tables are ready for the test.

Pending:
- add documentation


carlos-verdes:
@0xMochan @cvauclair do you mind taking a look into this?

cvauclair:
Hey @carlos-verdes just saw your ping, wasn't sure if the PR was ready for review since it's still a draft but I'll take a look later today!

Thanks a lot for the PR, let's get this `postgres` integration done :mechanical_arm:

carlos-verdes:
It's in draft because documentation is not finished, but the code + example + integration test is ready.

I can add later an example using Streams if that would be useful!

rig-postgres/README.md
```diff
@@ -0,0 +1,127 @@
...
+    let products: Vec<Product> = ...;
+
+    let documents = EmbeddingsBuilder::new(model.clone())
+        .documents(products)
+        .unwrap()
+        .build()
+        .await?;
+
+    // Create your index
+    let vector_store = PostgresVectorStore::default(model, pool);
```
cvauclair, line 117: `pool` definition is missing

rig-postgres/examples/vector_search_postgres.rs
```diff
@@ -0,0 +1,99 @@
...
+
+    let documents = EmbeddingsBuilder::new(model.clone())
+        .documents(words)
+        .unwrap()
+        .build()
+        .await
+        .expect("Failed to create embeddings");
+
+    // init vector store
+    sqlx::query("TRUNCATE documents").execute(&pool).await?;
```
cvauclair, line 80: Perhaps indicate that this query deletes existing documents in the DB, wouldn't want someone to run this thinking it's required

rig-postgres/src/lib.rs
```diff
@@ -0,0 +1,178 @@
...
+        query: &str,
+        n: usize,
+    ) -> Result<Vec<(f64, String)>, VectorStoreError> {
+        self.top_n::<Value>(query, n).await.map(|result| {
+            result
+                .into_iter()
+                .map(|(distance, id, _)| (distance, id))
+                .collect()
+        })
+    }
```
cvauclair, line 177: This still fetches the documents over the wire (which may or may not be a lot), what do you think of using an entirely different query that only selects the scores and ids? Might be premature optimization, so not sure if that's worth the added complexity

rig-postgres/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+name = "rig-postgres"
+version = "0.1.0"
+edition = "2021"
+description = "PostgreSQL-based vector store implementation for the rig framework"
+license = "MIT"
+
+[dependencies]
+futures = "0.3.31"
+tokio = { version = "1.43.0", features = ["macros", "rt-multi-thread"] }
+tokio-stream = "0.1.6"
```
cvauclair, line 11: Are those deps needed outside of tests/examples? I don't see them being used in `lib.rs` so unless I'm missing something, they should be moved to dev dependencies

rig-postgres/src/lib.rs
```diff
@@ -0,0 +1,178 @@
...
+            model,
+            pg_pool,
+            documents_table: documents_table.unwrap_or(String::from("documents")),
+            distance_function,
+        }
+    }
+
+    pub fn default(model: Model, pg_pool: PgPool) -> Self {
+        Self::new(model, pg_pool, None, PgVectorDistanceFunction::Cosine)
+    }
```
cvauclair, line 83: Perhaps rename to `with_defaults`? `default` is a method name usually reserved for the `Default` trait implementation

rig-postgres/Cargo.toml
```diff
@@ -0,0 +1,35 @@
+[package]
+name = "rig-postgres"
+version = "0.1.0"
+edition = "2021"
+description = "PostgreSQL-based vector store implementation for the rig framework"
+license = "MIT"
```
cvauclair, line 6: Need to add `readme` and `repository` fields or else crates.io complains

cvauclair (commented):
Looks solid! I added a couple suggestions/comments, please take a look when you have a moment.

Cheers!

rig-postgres/README.md
carlos-verdes: I wanted to keep the README small as you can go to examples or integration test for full code, but I'll add it

rig-postgres/examples/vector_search_postgres.rs
carlos-verdes: Added a better comment, good point

rig-postgres/src/lib.rs
carlos-verdes: It has sense and it also remove the deserialization step (not needed).
I added the new logic and updated integration test.

rig-postgres/Cargo.toml
carlos-verdes: updated dependencies

rig-postgres/src/lib.rs
carlos-verdes: true, done

rig-postgres/Cargo.toml
carlos-verdes: updated

carlos-verdes:
@cvauclair thanks for the review, I updated code based on your comments, take a look and let me know if you want to squash the commits.

cvauclair (approved):
Awesome work, thanks for the contribution @carlos-verdes !

carlos-verdes:
Thanks for the kind words, I just realized that main README is not updated :)

PR #232, opened 2025-01-22
feat: streaming API implementation for Anthropic provider

piotrostr:
- Added streaming traits and types in a new `src/streaming.rs` module
- Implemented streaming support for Anthropic models with tool call handling
- Added examples demonstrating streaming usage with and without tools
- Updated dependencies to support streaming functionality
- Extended `Agent` impls with streaming traits

Dependencies added:
- tokio-stream = "0.1.17"
- tokio = { version = "1.34.0", features = ["full"] } (currently dev dep)
- bytes = "1.9.0"
- Added "stream" feature to reqwest

Testing:
- `cargo run --example anthropic_streaming` 
- `cargo run --example anthropic_streaming_with_tools`


rig-core/src/agent.rs
```diff
@@ -425,3 +430,41 @@ impl<M: CompletionModel> AgentBuilder<M> {
         }
     }
 }
+
+impl<M: StreamingCompletionModel> StreamingCompletion<M> for Agent<M> {
+    async fn streaming_completion(
+        &self,
+        request: CompletionRequest,
+    ) -> Result<
+        Pin<Box<dyn Stream<Item = Result<StreamingChoice, CompletionError>> + Send>>,
```
carlos-verdes, line 439: Have you tried with any of these options?
```rust
BoxStream<Result<StreamChoice, CompletionError>>
```
or
```rust
impl TryStream<Ok=StreamChoice, Error=CompletionError>
```

Also normally with effectful streams you don't need to return `Result`, you can just return `impl  TryStream` (easier to consume)

rig-core/src/providers/anthropic/streaming.rs
```diff
@@ -0,0 +1,264 @@
...
+        let prompt_with_context = completion_request.prompt_with_context();
+
+        let max_tokens = if let Some(tokens) = completion_request.max_tokens {
+            tokens
+        } else if let Some(tokens) = self.default_max_tokens {
+            tokens
+        } else {
+            return Err(CompletionError::RequestError(
+                "`max_tokens` must be set for Anthropic".into(),
+            ));
```
carlos-verdes, line 100: Here instead of returning `Err` you return a `Stream` with an error inside:
```rust
use async_stream::stream;
stream! {
    let error = CompletionError::RequestError("`max_tokens` must be set for Anthropic".into());
    yield(Err(error));
}
```


This way your function will be like this:
```rust
async fn stream(
    &self,
    completion_request: CompletionRequest,
) ->impl TryStream<Ok=StreamingChoice, Err=CompletionError>> + Send {
```

rig-core/src/providers/anthropic/streaming.rs
```diff
@@ -0,0 +1,264 @@
...
+        }
+
+        let chunk = chunk_result.unwrap();
+        if let Ok(text) = String::from_utf8(chunk.to_vec()) {
+            for line in text.lines() {
+                if let Some(data) = line.strip_prefix("data: ") {
+                    if data.trim() == "[DONE]" {
+                        emit_final_tool_call(&mut current_tool_call, &tx).await;
+                        break;
+                    }
```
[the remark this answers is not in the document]
carlos-verdes: check my PR on Postgres please :)

rig-core/src/providers/anthropic/streaming.rs
```diff
@@ -0,0 +1,264 @@
...
+        }
+
+        let stream = response.bytes_stream();
+        let (tx, rx) = mpsc::channel(100);
+
+        tokio::spawn(async move {
+            process_stream(stream, tx).await;
+        });
+
+        Ok(Box::pin(ReceiverStream::new(rx)))
```
carlos-verdes, line 163: I don't think is needed to pin the stream and create a channel to consume it, normally you return the stream here and whenever is needed to be consumed you use `tokio::pin(stream)`.

Example, this function would be like this:
```rust
 async fn stream(
        &self,
        completion_request: CompletionRequest,
    ) -> impl TryStream<Ok=StreamingChoice, Err=CompletionError>> + Send {
. . .
    response.bytes_stream()
}
```

Then when I need to consume this stream:
```rust
#[tokio::main]
    let my_stream_model: StreamingCompletionModel  = ...;

    tokio::pin!(my_stream_model);

    while let Some(val) = my_stream_model.try_next().await? {
        print!(val);
    }
```

carlos-verdes:
I honestly think we should change the approach and avoid constraints like returning a pinned Stream, we should try to return `impl Stream` or `impl TryStream` unless we need trait objects.

Couple of good articles: 
- [Tokio Streams](https://tokio.rs/tokio/tutorial/streams)
- [A guided tour of streams in Rust](https://www.qovery.com/blog/a-guided-tour-of-streams-in-rust/)

rig-core/src/providers/anthropic/streaming.rs
```diff
@@ -0,0 +1,264 @@
...
+pub enum ContentBlock {
+    Text {
+        text: String,
+    },
+    ToolUse {
+        id: String,
+        name: String,
+        input: serde_json::Value,
+    },
+}
```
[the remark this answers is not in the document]
piotrostr: b1a7a4cb636de0b42c04e9d2255f56e78e4426ce

rig-core/src/providers/anthropic/streaming.rs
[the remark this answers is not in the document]
piotrostr: image

😅 06b734b449dcb451c3f06622050943e83537e2f5

rig-core/src/providers/anthropic/streaming.rs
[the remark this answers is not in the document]
piotrostr: a good deal, with some tweaks, will confirm exactly how much when implementing another provider

the core logic and trait for sure, hopefully some of the types too, APIs are quite similar on first glance

rig-core/src/providers/anthropic/completion.rs
```diff
@@ -151,9 +151,9 @@ impl From<completion::Message> for Message {
 
 #[derive(Clone)]
 pub struct CompletionModel {
-    client: Client,
+    pub client: Client,
```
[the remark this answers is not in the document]
piotrostr: something to consider, connected to the previous comment, i wanted to use the request builder, in this case we need to call

```rust
let response = self
    .client
    .post("/v1/messages")
    .json(&request)
    .send()
    .await?;
```

in the `StreamingCompletionModel::stream` method

gonna ponder on this a bit, must be an elegant way..

rig-core/src/agent.rs
piotrostr: I like this! way cleaner, gonna give it a go for sure

piotrostr:
@cvauclair, I agree we can benefit from generics here, since there will be probably 4-5 providers already and more to come in the future, going to implement a more elegant interface and possibly keep the `client` prop private

@carlos-verdes thank you for the suggestion with `tokio-stream`, while both approaches provide similar performance, the simpler API might be worth the minimal change

sorted the simple changes, going to push the update tomoz

thank you!

rig-core/src/providers/anthropic/streaming.rs
piotrostr: @carlos-verdes there would be some additional issues:
* the building of completion request is async, so we would not be able to maintain the async interface and re-use the `Agent::completion` for generating a prompt + chat_history requests
* rather than returning the pin (current impl), we pin it afterwards to have the retval `await: <TryNext<'_, Pin<&mut impl TryStream<Ok = StreamingChoice, Error = CompletionError>>> as IntoFuture>::Output`

while I gave it a go (https://github.com/piotrostr/rig/commit/58c31dcd707700673043f378a200e6b57e69ba1c), going to sort the interfaces first, if you know exactly how to sort it please let me know, might be missing something here

rig-core/src/agent.rs
piotrostr: 2a57cc8128ff8b153128ef1df39435207e9e4df4

piotrostr: and b39593135aedd009b6053e43baa63917f886f395

piotrostr: and 5cd7fb3a88b4d6bf3cda5546de3b338c29609109

rig-core/src/streaming.rs
```diff
@@ -0,0 +1,108 @@
...
+}
+
+/// Trait for low-level streaming completion interface
+pub trait StreamingCompletion<M: StreamingCompletionModel>: Send + Sync {
+    /// Generate a streaming completion from a request
+    fn streaming_completion(
+        &self,
+        request: CompletionRequest,
+    ) -> impl Future<Output = Result<StreamingResult, CompletionError>> + Send;
+}
```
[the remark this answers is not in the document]
piotrostr: yeah, bit of explanation - i wanted to re-use the existing builder for completion requests, the `stream` is the only method that requires implementing

by making the client public, we can implement just this single method and satisfy the same interface with prompt and chat, only thing that changes is we add the `stream: true` in the request, same as each completion implementation adjusts to match the API of the given provider

rig-core/src/providers/anthropic/completion.rs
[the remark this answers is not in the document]
piotrostr: oh yeah sorry misread that, if `pub(crate)` works then 560a0e545eb1884f74884201aea1986ede21c5f9

rig-core/src/agent.rs
carlos-verdes: Looks much cleaner right?

Some comments on your last code:
- the static lifetime I think is ok on this scenario but in general I would avoid
- BoxStream is a good abstraction but I would even try to remove using impl StreamTry
- instead of return Result<Stream<Result ... you can flatten to Stream<Result

I created an utility in one of my projects to accomplish these points (using reqwest):
```rust
use futures::{FutureExt, TryFutureExt, TryStreamExt};
use reqwest::{RequestBuilder, Response};
use reqwest_streams::{error::StreamBodyError, JsonStreamResponse};
use serde::Deserialize;

pub trait SendStream: Sized {
    fn send_stream(self) -> impl futures::TryStream<Ok = Response, Error = super::StreamError>;
    fn send_stream_then_parse<Output>(
        self,
    ) -> impl futures::TryStream<Ok = Output, Error = super::StreamError>
    where
        Output: for<'de> Deserialize<'de> + Send + 'static,
    {
        let result = self
            .send_stream()
            .map_ok(|response| {
                response
                    .json_array_stream::<Output>(64 * 1024)
                    .map_err(|e| {
                        super::StreamError::new(
                            "Error creating json stream from reqwest request",
                            e.into(),
                        )
                    })
            })
            .try_flatten();

        result
    }
}

impl SendStream for RequestBuilder {
    fn send_stream(self) -> impl futures::TryStream<Ok = Response, Error = super::StreamError> {
        self.send()
            .and_then(|response| async { response.error_for_status() })
            .map_err(|err| {
                println!("(send_stream) Error sending reqwest request: {:?}", err);
                err
            })
            .err_into()
            .into_stream()
    }

    fn send_stream_then_parse<Output>(
        self,
    ) -> impl futures::TryStream<Ok = Output, Error = super::StreamError>
    where
        Output: for<'de> Deserialize<'de> + Send + 'static,
    {
        let result = self
            .send_stream()
            .map_ok(|response| {
                response
                    .json_array_stream::<Output>(64 * 1024)
                    .map_err(|e| {
                        println!(
                            "(send_stream) Error creating json stream from reqwest request: {:?}",
                            e
                        );
                        super::StreamError::new(
                            "Error creating json stream from reqwest request",
                            e.into(),
                        )
                    })
            })
            .try_flatten();

        result
    }
}

impl From<reqwest::Error> for super::StreamError {
    fn from(err: reqwest::Error) -> Self {
        super::StreamError::new("Reqwest API error", err.into())
    }
}

impl From<StreamBodyError> for super::StreamError {
    fn from(err: StreamBodyError) -> Self {
        super::StreamError::new("Reqwest stream error", err.into())
    }
}
```

Once you have this utility you can use it like this:
```rust
// option with BoxStream (look we don't need async anymore
fn stream_prompt(&self, prompt: &str) -> BoxStream<Result<StreamingChoice, CompletionError>> {

    //example calling API using reqwest
        reqwest_client.
            .get(your_url)
            .send_stream_then_parse::<StreamingChoice>()
            .err_into()
            .boxed()
    }

// option without boxing
fn stream_prompt(&self, prompt: &str) -> impl StreamTry<Ok=StreamingChoice, Error=CompletionError> {

    //example calling API using reqwest
        reqwest_client.
            .get(your_url)
            .send_stream_then_parse::<StreamingChoice>()
            .err_into()
    }
```

rig-core/src/providers/anthropic/streaming.rs
carlos-verdes: I sent full example with one of my libraries (it's code from a working project) where I return impl Stream on functions that actually are not async, you can simply wrap the side effect inside the stream.

More or less the idea is the effect call (for example a http request) is wrapped inside the Stream, something like:
- if request is ok, then emit one Stream element per response result
- if request returns error, then emit one Stream error event 

The only point where you need async/pin is when you consume the stream, this mean that all the logic around streams is fully lazy, there is non blocking until you actually need to consume the stream.

rig-core/src/agent.rs
piotrostr: not sure if this is cleaner tbh

i implemented sse as part of another thing, with this interface as well, works like a charm

here we are working with async functions (rig completion compatible interface) here and we **do** wanna return a future

while using the boxed stream is an improvement as the type there was kinda tricky, for the TryStream trait - check out the solana pubsub client, they do something similar, without the TryStream as well

but feel free to pick up the WIP branch, i posted a link earlier and started on this implementation, was a headache so prioritized other stuff

cvauclair:
@piotrostr just tried to compile your branch to WASM (`cargo build --features worker --target wasm32-unknown-unknown`) and it looks like `tokio` is not compilable to wasm (makes sense since the WASM environment would have it's own async runtime).

Since we do want to support compiling Rig to WASM, there are two options:
1. We ditch the `tokio` and `tokio-stream` and use `futures` instead (basically @carlos-verdes 's suggestion)
2. We put all of this behind a `stream` feature flag

Personally I favor the first option since (I think) it would make the streaming API available when compiling to WASM, but for the sake of time we could go for option 2. now and implement option 1. in a subsequent PR.

Also if you could rebase on main and fix the conflicts that would be appreciated, thank you :grin:

piotrostr:
dropped the tokio dep 244645980e6d3c70338dc580aff606e11a82fac6, `cargo build --features worker` build clean

i could not get `cargo build --features worker --target wasm32-unknown-unknown` to build, but also on the main branch (tag: 896774b3255332025284c9c4bc57ef4232550976)

cvauclair:
@piotrostr not all rig crates are currently wasm compatible, only `rig-core` for now! And FYI the `main` branch does compile to wasm (`cargo build --package rig-core --features worker --target wasm32-unknown-unknown`).

I'll go back to my initial idea of putting the streaming API behind a `stream` (or `streaming`) feature flag for now until we figure out how to make it wasm compatible!

cvauclair (approved):
Good work @piotrostr :muscle:

PR #234, opened 2025-01-23
feat: Add support for Azure OpenAI

jimmydottech:
## Description

This PR adds support for **Azure OpenAI** to the AI system, enabling seamless integration with Azure's LLM services alongside existing providers.

While Azure OpenAI is similar to OpenAI, it requires unique configurations such as API version, custom endpoints, and deployment IDs as model names. For example:

```
POST https://{endpoint}/openai/deployments/{deployment-id}/embeddings?api-version=2024-10-21
```

URI Parameters:
Name | In | Required | Type | Description
-- | -- | -- | -- | --
endpoint | path | Yes | stringurl | Supported Azure OpenAI endpoints (protocol and hostname, for example: https://aoairesource.openai.azure.com. Replace "aoairesource" with your Azure OpenAI resource name). https://{your-resource-name}.openai.azure.com
deployment-id | path | Yes | string | Deployment ID of the model which was deployed.
api-version | query | Yes | string | API version

Request Header:
Name | Required | Type | Description
-- | -- | -- | --
api-key | True | string | Provide Azure OpenAI API key here

Latest API version:
Latest GA API | Latest Preview API
-- | --
2024-10-21 | 2024-10-01-preview

Reference:
https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/models
https://learn.microsoft.com/en-us/azure/ai-services/openai/reference
https://learn.microsoft.com/en-us/azure/ai-services/openai/supported-languages
https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/switching-endpoints


rig-core/src/providers/azure.rs
```diff
@@ -0,0 +1,593 @@
...
+    azure_endpoint: String,
+    http_client: reqwest::Client,
+}
+
+impl Client {
+    /// Create a new Azure OpenAI client with the given API key.
+    pub fn new(api_key: &str, api_version: &str, azure_endpoint: &str) -> Self {
+        let azure_endpoint = azure_endpoint.trim().trim_end_matches("/");
+        // Validate the endpoint format
+        if !azure_endpoint.starts_with("https://") || !azure_endpoint.ends_with(".openai.azure.com")
```
joshua-mo-143, line 38: We shouldn't enforce a URL endpoint here - the reason being that if we were to try and use a mock server with our endpoint, it would fail (because this only accepts live production URLs). It is also possible that a user may want to use a proxy endpoint instead of just pointing to the Azure OpenAI resource.

joshua-mo-143 (changes requested):
Looks good! Just need to fix the one change request and we'll be good to go.

joshua-mo-143 (approved):
Lgtm

PR #251, opened 2025-01-29
feat: support tools for DeepSeek provider

### LINKED ISSUE #250: DeepSeek Completion support for tools
As you can see on the code DeepSeek integration is ignoring tools configuration:
```rust
let body = json!({
            "model": self.model,
            "messages": messages_json,
            "frequency_penalty": 0,
            "max_tokens": request.max_tokens.unwrap_or(2048),
            "presence_penalty": 0,
            "temperature": request.temperature.unwrap_or(1.0),
            "top_p": 1,
            "tool_choice": "none",
            "logprobs": false,
            "stream": false,
        });
```

From their [documentation](https://api-docs.deepseek.com/api/create-chat-completion):

> **tool_choice** (object, nullable)
> 
> Controls which (if any) tool is called by the model.
> 
> **none:** means the model will not call any tool and instead generates a message.

I expect an implementation similar to openAI:
```rust
        let request = if completion_request.tools.is_empty() {
            json!({
                "model": self.model,
                "messages": full_history,
                "temperature": completion_request.temperature,
            })
        } else {
            json!({
                "model": self.model,
                "messages": full_history,
                "temperature": completion_request.temperature,
                "tools": completion_request.tools.into_iter().map(ToolDefinition::from).collect::<Vec<_>>(),
                "tool_choice": "auto",
            })
        };
```

carlos-verdes:
Fixes #250

### DIFF
--- a/rig-core/Cargo.toml
+++ b/rig-core/Cargo.toml
@@ -37,6 +37,7 @@ assert_fs = "1.1.2"
 tokio = { version = "1.34.0", features = ["full"] }
 tracing-subscriber = "0.3.18"
 tokio-test = "0.4.4"
+dotenvy = "0.15.7"
 
 [features]
 all = ["derive", "pdf", "rayon"]
--- a/rig-core/examples/agent_with_deepseek.rs
+++ b/rig-core/examples/agent_with_deepseek.rs
@@ -2,6 +2,9 @@ use rig::{completion::Prompt, providers};
 
 #[tokio::main]
 async fn main() -> Result<(), anyhow::Error> {
+    // load env from .env file
+    dotenvy::dotenv().ok();
+
     let client = providers::deepseek::Client::from_env();
     let agent = client
         .agent("deepseek-chat")
--- a/rig-core/examples/agent_with_tools.rs
+++ b/rig-core/examples/agent_with_tools.rs
@@ -47,7 +47,7 @@ impl Tool for Adder {
     }
 
     async fn call(&self, args: Self::Args) -> Result<Self::Output, Self::Error> {
-        println!("Adding {} and {}", args.x, args.y);
+        println!("[tool-call] Adding {} and {}", args.x, args.y);
         let result = args.x + args.y;
         Ok(result)
     }
@@ -84,7 +84,7 @@ impl Tool for Subtract {
     }
 
     async fn call(&self, args: Self::Args) -> Result<Self::Output, Self::Error> {
-        println!("Subtracting {} from {}", args.y, args.x);
+        println!("[tool-call] Subtracting {} from {}", args.y, args.x);
         let result = args.x - args.y;
         Ok(result)
     }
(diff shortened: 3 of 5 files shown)


carlos-verdes:
@cvauclair take a look into this one, I think it's an important one

carlos-verdes:
@maxsorto can you take a look into this?

joshua-mo-143:
Hi Carlos, thanks for this PR! It is currently quite late but when I am available tomorrow I'll be reviewing this.

rig-core/examples/agent_with_deepseek.rs
```diff
@@ -2,6 +2,9 @@ use rig::{completion::Prompt, providers};
 
 #[tokio::main]
 async fn main() -> Result<(), anyhow::Error> {
+    // load env from .env file
+    dotenvy::dotenv().ok();
```
joshua-mo-143, line 6: Not required - users can just use environment variables instead

rig-core/examples/agent_with_tools.rs
```diff
@@ -83,13 +84,17 @@ impl Tool for Subtract {
...
+        println!("[tool-call] Subtracting {} from {}", args.y, args.x);
         let result = args.x - args.y;
         Ok(result)
     }
 }
 
 #[tokio::main]
 async fn main() -> Result<(), anyhow::Error> {
+    // load env from .env file
+    dotenvy::dotenv().ok();
```
joshua-mo-143, line 96: Not required - users can just use environment variables instead

rig-core/examples/agent_with_tools.rs
```diff
@@ -103,9 +108,34 @@ async fn main() -> Result<(), anyhow::Error> {
...
+    let calculator_agent = deepseek_client
+            .agent(providers::deepseek::DEEPSEEK_CHAT)
+            .preamble("You are a calculator here to help the user perform arithmetic operations. Use the tools provided to answer the user's question.")
+            .max_tokens(1024)
+            .tool(Adder)
+            .tool(Subtract)
+            .build();
+
+    // Prompt the agent and print the response
+    println!("\n\n####################################################");
```
joshua-mo-143, line 133: Nit: It might be better to add the deepseek tools example to `agent_with_deepseek.rs` here. Having it here isn't necessarily wrong, but a user might expect to find a deepseek example with tools to be in the deepseek example rather than in the tools example

joshua-mo-143 (changes requested):
Thanks for creating this PR Carlos! Just need to fix the pending issues then we should be good to go.

rig-core/Cargo.toml
```diff
@@ -37,6 +37,7 @@ assert_fs = "1.1.2"
 tokio = { version = "1.34.0", features = ["full"] }
 tracing-subscriber = "0.3.18"
 tokio-test = "0.4.4"
+dotenvy = "0.15.7"
```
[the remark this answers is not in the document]
carlos-verdes: I always use .env files and I think many developer too as it's simple to setup.
I only added as dev-dependencies but if this is still an issue I'll remove.

rig-core/examples/agent_with_tools.rs
carlos-verdes: I thought it would be good idea to test the call to tools using different providers, I'll move into deepseek file.

carlos-verdes:
@joshua-mo-143 changes applied

PR #253, opened 2025-01-29
feat!: new field for Message struct to support tool chain calls

carlos-verdes:
it breaks current usage of Message api as you need a new field

Related with discussion #252


0xMochan:
Hey, this is handled in #199 which is in the middle of our review process!

carlos-verdes:
I think I'll close this one and focus on #199 

One thing we should notify that PR is a breaking change to get proper semver

PR #261, opened 2025-02-03
fix: Fixed incorrect article usage before "ancient"

defiberrys:
I noticed a issue where "a ancient" was used instead of "an ancient".
Since "ancient" starts with a vowel sound, the correct article is "an".

This has been updated for clarity and correctness.


cvauclair:
Duplicate of #260

PR #262, opened 2025-02-03
chore: Improved test assertion style for better readability

defiberrys:
I replaced `assert!(expected == actual)` with `assert_eq!(expected, actual)` in the test. While both work, `assert_eq!` is more idiomatic in Rust as it provides better output for debugging by automatically displaying the values when the assertion fails.

p.s. Change improves the readability and overall quality of the tests.


joshua-mo-143:
Closing because there has been no response from the PR author.

If you're still interested in opening another PR, you are free to do so!

PR #265, opened 2025-02-03
feat: galadriel api integration (redux)

joshua-mo-143:
Adds Galadriel API provider integration.

Note that this was originally implemented in #188 but then had to be reverted due to a missing test that *should've* been ran not being run which mistakenly led to a merge. This has been reverted to ensure CI reliability in other PRs while we figure out exactly why the test isn't running. Otherwise, it's exactly the same file contents.

cc @kristjanpeterson1 (just as an FYI in case this doesn't get merged in time for Rig 0.8)

### DIFF
--- a/rig-core/examples/agent_with_galadriel.rs
+++ b/rig-core/examples/agent_with_galadriel.rs
@@ -1,0 +1,23 @@
+use rig::{completion::Prompt, providers};
+use std::env;
+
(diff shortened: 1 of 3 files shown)


joshua-mo-143:
For future reference: Adding the `cfg-attr` for the worker feature fixed it :facepalm: d'oh

PR #266, opened 2025-02-03
fix: o3-mini doesn't support temperature

### LINKED ISSUE #258: bug: o3-mini temperature parameter

Nothing in Issues mentioning "o3-mini"

## Bug Report
`CompletionError(ProviderError("{\n  \"error\": {\n    \"message\": \"Unsupported parameter: 'temperature' is not supported with this model.\",\n    \"type\": \"invalid_request_error\",\n    \"param\": \"temperature\",\n    \"code\": \"unsupported_parameter\"\n  }\n}"))`

## Reproduction
```rust
use rig::{providers::openai, agent::AgentBuilder};

// Initialize the OpenAI client
let openai = openai::Client::new("your-openai-api-key");

// Create a model and initialize an agent
let o3-mini = openai.completion_model("o3-mini");

let agent = AgentBuilder::new(o3-mini)
    .preamble("\
        You are Gandalf the white and you will be conversing with other \
        powerful beings to discuss the fate of Middle Earth.\
    ")
    .build();

// Alternatively, you can initialize an agent directly
let agent = openai.agent("gpt-4o")
    .preamble("\
        You are Gandalf the white and you will be conversing with other \
        powerful beings to discuss the fate of Middle Earth.\
    ")
    .build();
```

## Expected behavior
The code executes without error.

## Screenshots
N.A.

## Additional context
N.A.

joshua-mo-143:
Fixes #258 

Admittedly this is a bit of a crude fix, but should work for now until we can find a better way to do this (which will probably take more time).


rig-core/src/providers/openai.rs
```diff
@@ -510,18 +510,29 @@ impl completion::CompletionModel for CompletionModel {
...
                 "model": self.model,
                 "messages": full_history,
-                "temperature": completion_request.temperature,
                 "tools": completion_request.tools.into_iter().map(ToolDefinition::from).collect::<Vec<_>>(),
                 "tool_choice": "auto",
             })
         };
 
+        // o3-mini doesn't support temperature value
+        let request = if !self.model.starts_with("o3-mini") {
```
0xMochan, line 525: tbh, `temperature` is `Option` for a reason. imo, it's better to just not include `temperature` if it's `None`. In the future, I want to improve some of these request models via some structure (so we can use `skip_serializing_if: "Option::is_none"` with serde models).

0xMochan (changes requested):
While we do check for model strings in Anthropic, ideally we avoid is as much as possible...

I think it be simpler to refactor the code snippet to not include `temperature` in the json blob unless defined. Future work will involve defining a proper `CompletionRequest` model rather than raw json!

0xMochan (approved):
Looks good

PR #267, opened 2025-02-04
feat: View static agent context

### LINKED ISSUE #239: feat: View agent's static context

## Feature Request

ref Discord conversation https://discord.com/channels/896944341598208070/1330217496988352656

Users should be able to see what exactly gets sent to the API (as the prompt).

### Motivation

Better observability for LLM calls

### Proposal

I propose that we could add a new method to the `AgentBuilder` (or `Agent` struct) to be able to see what the underlying prompt is.

### Alternatives

Not really sure what the alternative is here

joshua-mo-143:
Fixes #239 to some degree by allowing a new `PromptData` struct to be pulled out of a completion request, as well as allowing a new `TracePrompt` struct that allows users to get traces of the prompt they've sent to the LLM.

We should probably make it more obvious that `CompletionBuilder` is a thing as it feels like there is a bit of a discoverability issue since all of the examples basically prioritise using the easiest way possible to do everything.

### DIFF
--- a/rig-core/src/completion.rs
+++ b/rig-core/src/completion.rs
@@ -246,6 +246,52 @@ pub trait CompletionModel: Clone + Send + Sync {
     }
 }
 
+/// A way to get the final prompt of a completion request.
+///
+/// Example usage:
+/// ```rust
+/// use rig::{
+///     providers::openai::{Client, self},
+///     completion::CompletionRequestBuilder,
+/// };
+///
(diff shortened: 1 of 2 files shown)


0xMochan:
Likely will require some smart merging since `completions` exploded into `completions/request` and `completions/message`, we can pair program this at some point!

joshua-mo-143:
Closing because there's probably a better way to do this as previously discussed in eng sync.

PR #268, opened 2025-02-04
feat: fastembed integration

### LINKED ISSUE #228: feat: `fastembed-rs` integration
It would be nice if we could use https://github.com/Anush008/fastembed-rs for more cost-effective computation of embeddings locally, instead of having to call the cloud APIs🙂

joshua-mo-143:
Closes #228 

We may want to add other things like `SparseModel` in the future. However, this PR aims to serve as a starting point for the integration so this is somewhat out of scope (for the PR).


0xMochan (approved):
Still need to run locally, looks pretty solid and straightforward
- imo `rig-fastembed` as the example name is confusing, can it be renamed to like `local_embeddings` or `vector_search`, etc.
- update README.md w/ new logo

joshua-mo-143:
Fastembed doesn't actually have a logo, do you want me to just add the text into the table row?

0xMochan (approved):
Works like a charm, pretty awesome crate 🔥.

joshua-mo-143:
Merging as the tests all pass

PR #270, opened 2025-02-05
fix(galadriel): missed fixes from messages pr

0xMochan:
I had some commits that got missed in a `git stash` 😅.


joshua-mo-143:
Lgtm, looks like it runs locally

PR #276, opened 2025-02-06
feature: manage tool responses

### LINKED ISSUE #274: DeepSeek auth header is missing
After #199 DeepSeek api call doesn't bear the auth header.

I'll fix as part of my next PR with next changes:
- first we need to remove api_key from the Client struct
- second when we build the Client we need to add the auth header

carlos-verdes:
It implements the next flow:
- human prompt
- agent respond with tool call
- tool is called
- tool response is sent back to agent
- assistant response is sent back to human

It works with OpenAI but I found an issue with DeepSeek and the serialization of content messages, when you send the tool call back to DeepSeek the content is empty and DeepSeek expects `content: ""`, however current serialization (based on OpenAI) is sending `content: []` which makes DeepSeek fail.

 

also fixes #274

### DIFF
--- a/rig-core/examples/agent_with_deepseek.rs
+++ b/rig-core/examples/agent_with_deepseek.rs
@@ -8,9 +8,14 @@ use serde_json::json;
 
 #[tokio::main]
 async fn main() -> Result<(), anyhow::Error> {
+    tracing_subscriber::fmt()
+        .with_max_level(tracing::Level::DEBUG)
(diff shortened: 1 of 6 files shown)


carlos-verdes:
Created smaller PR covering only agent changes + fixing OpenAI issues (another PR fixing DeepSeek Message model is open which will make tool calls work)

PR #280, opened 2025-02-07
feat: surrealdb integration

### LINKED ISSUE #269: feat: Add support for SurrealDB

SurrealDB for RAG.

There are multiple components to this:
- [x] RAG
- [ ] Graph RAG

### Resources

[Our blog post](https://dev.to/josh_mo_91f294fcef0333006/building-rag-using-rig-surrealdb-deepseek-on-cloudflare-workers-484)
[SurrealQL docs](https://surrealdb.com/docs/surrealql)

joshua-mo-143:
Stage one of #269 

I couldn't find a way to store the documents better, so for now it's raw JSON strings.


rig-surrealdb/examples/vector_search_surreal.rs
```diff
@@ -0,0 +1,78 @@
...
+impl std::fmt::Display for WordDefinition {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.word)
+    }
+}
+
+#[tokio::main]
+async fn main() -> Result<(), anyhow::Error> {
+    // load environment variables from .env file
+    dotenvy::dotenv().ok();
```
0xMochan, line 26: Do we use `dotenv` in other examples in rig?

rig-surrealdb/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+            distance_function,
+            documents_table,
+            ..
+        } = self;
+        format!(
+            "
+               SELECT id {document} {embedded_text}, {distance_function}($vec, embedding) as distance \
+              from {documents_table} order by distance desc \
+            LIMIT $limit",
+        )
```
0xMochan, line 115: Is the best way to compose sql queries via `format`. Is there no prepared statement or some sort of protection for sql injection, wondering what's the defacto way to do things like this!

rig-surrealdb/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+                let embedded_text = embedding.document;
+                let embedding: Vec<f64> = embedding.vec;
+
+                let record = CreateRecord {
+                    document: json_document_as_string.clone(),
+                    embedded_text,
+                    embedding,
+                };
+
+                let _thing: Option<CreateRecord> = self
```
0xMochan, line 136: If `_thing` is unused, why capture the value? I'm guessing you were debugging this so maybe the `trace` call should be left here?

rig-surrealdb/README.md
0xMochan, line 1: This explanation should be more detailed. Add a `## Setup` or `## Installation` with links to surreal on how to setup a local or cloud instance.

0xMochan (approved):
Looks fantastic, I need to test locally still but i made a couple of comments

rig-surrealdb/src/lib.rs
joshua-mo-143: So, this is actually how it's done in [the Postgres module.](https://github.com/0xPlaygrounds/rig/blob/main/rig-postgres/src/lib.rs#L102) If wanted to refactor this into several methods (one for getting a query that only has the fields related to the `top_n_ids` method and one that also retrieves the text or document), I think that would probably be the best way to solve this. At the very least, it'd probably make it extremely unambiguous 

When we're actually executing the query, we bind variables (see L161) which is [the suggested SurrealDB way to do this.](https://surrealdb.com/docs/surrealdb/reference-guide/security-best-practices#query-safety)

rig-surrealdb/examples/vector_search_surreal.rs
joshua-mo-143: Some of them, yeah. This was mostly an artifact from forgetting to delete the dotenv file - I'll get around to deleting it in the next iter

rig-surrealdb/src/lib.rs
joshua-mo-143: Updated this for now so it should be fully bulletproof from SQL injection. We should probably endeavour to do the same in the Postgres module as soon as is feasible to do so.

rig-surrealdb/src/lib.rs
joshua-mo-143: Amended in next commit, I couldn't figure out how to get the type to work properly so I ended up declaring the variable and type.

rig-surrealdb/src/lib.rs
0xMochan: w.r.t postgres, I had the same conclusion there but it seemed like a proper / generalized, `sqlx` integration would have been appropriate with general sql injection protection.

It seems like most of the injection protection should be done on the actual variable side with word look up (so you can't embed like `Bobby; DROP TABLE`) but table names and other dev stuff is more naturally like an environment variable (aka, developer responsibility).

I appreciate you looking into it though, it's good for us to be vigilant during the PR!

0xMochan (approved):
LGTM

PR #283, opened 2025-02-07
refactor: deepseek message to remove dependencies with openai

carlos-verdes:


carlos-verdes:
@cvauclair @0xMochan do you mind taking a look into this one?

rig-core/src/completion/message.rs
```diff
@@ -71,6 +71,8 @@ pub enum ToolResultContent {
 #[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
 pub struct ToolCall {
     pub id: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub index: Option<usize>,
```
0xMochan, line 75: I'm not sure I understand what this index is referencing, is this an important field? It seems to me if there's an id to the tool call, there's no need for an index if there are multiple tool calls.

I think it shouldn't be in this PR as it would require updates to other providers like anthropic, etc.

rig-core/src/providers/deepseek.rs
```diff
@@ -62,7 +58,7 @@ impl Client {
                     headers
                 })
                 .build()
-                .expect("OpenAI reqwest client should build"),
+                .expect("DeeSeek reqwest client should build"),
```
0xMochan, line 61:
```suggestion
                .expect("DeepSeek reqwest client should build"),
```

0xMochan (changes requested):
Looks good, but wanted to check up on the `index` field you are adding.

rig-core/src/completion/message.rs
0xMochan: JSON has ordering in lists, so i'm not entirely sure if it's necessary. Maybe i'll check up with the docs, sometimes these providers do really odd things with their modelling.

carlos-verdes: I removed everything that is not related with DeepSeek, if we get into issues with the index in future it will be an easy PR

carlos-verdes:
@0xMochan applied changes so current PR only affects DeepSeek code

0xMochan (approved):
Looks good, will test and merge

PR #286, opened 2025-02-08
fix(openai): correct some fields for tools

s6nqou:
Three small fixes.
Check out the commits for more details.


carlos-verdes:
Applied part similar fixes + callback to model with ToolResult in #290

0xMochan (approved):
Good catch!

joshua-mo-143:
Just needs a `cargo fmt` and this is ready to merge :smile:

Issue #227, opened 2025-01-21
feat: `strict` parameter in ToolDefinition


klntsky:
- [x] I have looked for existing issues (including closed) about this

## Feature Request

ToolDefinition does not contain a `strict` option

https://platform.openai.com/docs/assistants/tools/function-calling 

### Alternatives

The only alternative is to validate the returned JSON manually

Tameflame:
I'm definitely in need for this feature, or at least a workaround. Is there a way to add strict to a tooldef so that the jsonschema passed to the llm is adhered to more consistently? I'm getting non-adherence to my schema in tool responses and it may be because of this

nikhil-pandey:
@joshua-mo-143 Any plans on supporting structured outputs?

0xMochan:
@nikhil-pandey Looking into getting this working. The strict argument isn't consistent with all providers (and some might be default strict)!

joshua-mo-143:
@klntsky @Tameflame @nikhil-pandey 

Hi folks, we're migrating our OpenAI integration to use the Responses API which makes it trivial to implement `strict` parameters on structured outputs (via the `text` config). Once #503 has a PR opened for it and it's been closed, that will then automatically close this issue.

(in terms of accessing the old Chat Completions API: you'll still be able to do it through a method on the adjusted CompletionModel struct)

Issue #149, opened 2024-12-13
feat: client versioning


0xMochan:
- [x] I have looked for existing issues (including closed) about this

## Feature Request
As model providers start getting more competitive, the provider API routes might adjust and evolve over time. We should consider specifying a standardized versioning with clients (maybe via a builder or something) that helps in specifying how requests should be made to the API.

### Motivation
Cohere recently released a [v2](https://docs.cohere.com/docs/migrating-v1-to-v2) of it's chat endpoint.

### Proposal
- A generic `ClientBuilder` can be defined to build out specific clients. That way, we can make the `ClientBuilder` as the singular interface for building all clients and agents.
- A `Provider` trait that helps simplify current and future provider implementation (specifically for top-level clients).
  - This should make it much easier for having different versioned clients since they'd share the same set of methods to make `agent`s, `extractor`s, etc.

### Alternatives
Only support the latest version. Against this suggestion primarily because LLM and their providers are moving so quickly that many people might be building agents that could become dated in only a month. The providers leave older version endpoints operational so it would be a shame if someone's Agent stopped working due to a breaking change.

The only positive with supporting a latest version is that it simplifies the number of clients per providers but making a provider generic over a `Version` might be suitable for this.

milancermak:
When I was looking through the source code, I've noticed that the way a completion request is created [for Anthropic](https://github.com/0xPlaygrounds/rig/blob/main/rig-core/src/providers/anthropic/completion.rs#L200) is not correct. It's modeled after the [OpenAI API](https://platform.openai.com/docs/api-reference/messages/createMessage#messages-createmessage-attachments) not the [Anthropic](https://docs.anthropic.com/en/docs/build-with-claude/prompt-engineering/long-context-tips#essential-tips-for-long-context-prompts) one.

IIUC tackling this issue would resolve that, right? If so, could you provide a little bit more detail on how you envision the refactor? I could work on it if that's ok.

PR #292, opened 2025-02-10
feat: allow malformed json parsing

TwistingTwists:
addresses #227 

This PR aims at more robust parsing of LLM text to yield Json. and hence helps in parsing the struct. 

There are many ways json could be malformed when returned by LLM. 

1. ` ``` ` backticks

```
let response = r#"
    ```
    {
        "name": "Bob",
        "age": 42,
        "address": {
            "street": "789 Pine Rd",
            "city": "Metropolis",
            "country": "USA",
            "coordinates": {"lat": 40.7128, "lng": -74.0060}
        },
        "hobbies": [
            {"name": "Cooking", "years_active": 5, "proficiency": "intermediate"},
            {"name": "Cycling", "years_active": 10, "proficiency": "expert"}
        ]
    }
    ```
    "#;

```

2. ` ``` json ` backticks with tag `json`
3.  introductory string 

```
let response = r#"
    Here is your json response:
    
    {
        "name": "Bob",
        "age": 42,
        "address": {
            "street": "789 Pine Rd",
            "city": "Metropolis",
            "country": "USA",
            "coordinates": {"lat": 40.7128, "lng": -74.0060}
        },
        "hobbies": [
            {"name": "Cooking", "years_active": 5, "proficiency": "intermediate"},
            {"name": "Cycling", "years_active": 10, "proficiency": "expert"}
        ]
    }
    
    You can use this now.
    "#;
```

---

[Current approach ](https://github.com/0xPlaygrounds/rig/blob/ef1e1aff4a1727f53002c44c9ed4747dcc48c6c3/rig-core/src/extractor.rs#L72) using serde_json::from_str does not cover these cases. When parsing with [json_partial](https://crates.io/crates/json_partial), all these fixes are handled to give a struct.

---
P.S. I am the author of json_partial

more tests related to parsing json using json_partial can be seen at https://github.com/TwistingTwists/aiworker/blob/36eedcc280a624c3f30cfd5fc712a522d8dbb938/examples/structured-jsonish/src/main.rs#L117-L261

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -89,10 +89,54 @@ dependencies = [
 ]
 
 [[package]]
+name = "anstream"
+version = "0.6.18"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8acc5369981196006228e28809f761875c0327210a891e941f4c683b3a99529b"
+dependencies = [
+ "anstyle",
+ "anstyle-parse",
+ "anstyle-query",
+ "anstyle-wincon",
+ "colorchoice",
+ "is_terminal_polyfill",
+ "utf8parse",
+]
+
+[[package]]
 name = "anstyle"
 version = "1.0.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "55cc3b69f167a1ef2e161439aa98aed94e6028e5f9a59be9a6ffb47aef1651f9"
+
+[[package]]
+name = "anstyle-parse"
+version = "0.2.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3b2d16507662817a6a20a9ea92df6652ee4f94f914589377d69f3b21bc5798a9"
+dependencies = [
+ "utf8parse",
+]
+
+[[package]]
+name = "anstyle-query"
+version = "1.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "79947af37f4177cfead1110013d678905c37501914fba0efea834c3fe9a8d60c"
+dependencies = [
+ "windows-sys 0.59.0",
+]
+
+[[package]]
+name = "anstyle-wincon"
+version = "3.0.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ca3534e77181a9cc07539ad51f2141fe32f6c3ffd4df76db8ad92346b003ae4e"
+dependencies = [
+ "anstyle",
+ "once_cell",
+ "windows-sys 0.59.0",
+]
 
 [[package]]
 name = "anyhow"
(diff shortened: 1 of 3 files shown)


0xMochan (commented):
Hey! We really appreciate your PR, your crate looks very useful for LLM applications! For rig specifically, it's important that we are very careful with which dependencies we are using for our projects as we strive to be lean (and are continuing auditing our current set of deps!)

For this PR specifically, I think the idea of having more flexible json parsing for LLMs is a great idea but it's something we want to logic and handle within the rig repo. We also are looking on adding better support for structured outputs as several providers now have direct ways of ensuring proper json and even schema'd output!

TwistingTwists:
Thanks for the feedback @0xMochan 

I appreciate the 'low dependency ' scenario. 

Here are my thoughts: 
1. I could raise another PR to include the code of jsonish crate in rig itself. That way, you don't have another dependency. 
2.  larger models claim to give structured outputs, smaller and cheaper models cannot follow the instruction accurately. Having worked on LLMs for the past year or so, this is a repeated feedback I've had for libraries i have used. 

Here is the scenario that bites often Enough, in my experience. 

I have a pipeline using large model and everything works fine. The moment I switch to smaller model, things go haywire. 

3. Would it be possible for you to make extractor a trait ? That way, I can write a `rig-json-partial-extractor` and it doesn't have to be included in the rig itself. 

People can use the default extractor or any other crate. 

Let me know your thoughts.

0xMochan:
> 1. I could raise another PR to include the code of jsonish crate in rig itself. That way, you don't have another dependency.

This is interesting, I'll bring it up with the team. I don't want the norm to be us usurping crates from the community but there might be some way we can co-maintain or something. From our POV, it's a big ask for rig users to add a new dep to their tree and it's our responsibilities as maintainers to ensure we are diligent with every required dep we add (and optional ones too!)

> 2. larger models claim to give structured outputs, smaller and cheaper models cannot follow the instruction accurately. Having worked on LLMs for the past year or so, this is a repeated feedback I've had for libraries i have used.

Correct, and i also agree that smaller models are going to be a big deal in larger, multi-agent systems, the type where rig could be extremely proficient in!

> 3. Would it be possible for you to make extractor a trait ? That way, I can write a `rig-json-partial-extractor` and it doesn't have to be included in the rig itself.

There actually was a sizable attempt to revamp extracts to be better integrated into the agent stack. I had a `StructuredPrompt` and `StructuredChat` trait defined to allow you to dynamically cast the output of an agent. You would do:
```rs
// Roughly
let agent = client.agent().cast::<Vec<MyType>>().build();
let output: Vec<MyType> = agent.prompt("...")
```

There was some issues we didn't resolve before I started work on the more urgent #199, and now I'd like to come back to that so extractors feel a lot more at "home". there's also the case of structured outputs not being able to be configured and for that I'm inclined to look into #149 Client traits as a way for providers to super-power agents with extra, defined behavior.

PR #297, opened 2025-02-11
fix(gemini): fixed tool calling + tool extractor demo

0xMochan:
Gemini tool calling was broken, this fixes it


joshua-mo-143 (approved):
lgtm, tested locally

PR #298, opened 2025-02-11
fix: make StreamingResult with Send trait

Jarekkkkk:
1. replace Pin<Bx> type with futures::stream::BoxStream for StreamingResult type

### DIFF
--- a/rig-core/src/streaming.rs
+++ b/rig-core/src/streaming.rs
@@ -13,11 +13,10 @@ use crate::agent::Agent;
 use crate::completion::{
     CompletionError, CompletionModel, CompletionRequest, CompletionRequestBuilder, Message,
 };
-use futures::{Stream, StreamExt};
-use std::boxed::Box;
+use futures::stream::BoxStream;
+use futures::StreamExt;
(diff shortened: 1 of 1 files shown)


0xMochan:
Hey, thanks for this PR! I noticed there's not an attached issue, can you expand on the problem you are solving here so we can best understand how the solution impacts this feature, thanks!

Jarekkkkk:
> Hey, thanks for this PR! I noticed there's not an attached issue, can you expand on the problem you are solving here so we can best understand how the solution impacts this feature, thanks!

hi sir, thanks for the response

so my problem here is when i want to spawn the task for stream related functions, i'll be blocked by Send trait from tokio


```
    let agent = create_suilend_agent().await.unwrap();
    let handle = tokio::spawn(async {
        let mut stream = agent
            .stream_chat("Who are you", vec![])
            .await?;

        stream.next();

        Ok(())
    });
```

0xMochan:
Hey, I think #302 solves this!

Jarekkkkk:
> Hey, I think #302 solves this!

i see, thanks for reply

PR #304, opened 2025-02-15
feat: support ollama provider

lispking:


0xMochan:
#285 implements the ollama client via the direct ollama APIs rather than the pass-through openai ones! We'll likely prioritize that PR over this one.

PR #307, opened 2025-02-18
fix: Resolved the issue of failed PDF loading.

lispking:
This morning, when writing an [article](https://mp.weixin.qq.com/s/Z2lX0gopCtgQpjFaKQBnwQ), I found that the loading of some PDF files of papers failed. 

After investigation, the cause of the failure was: 

> Error reading PDF content: IO error: ToUnicode CMap error: Could not parse ToUnicodeCMap: Error

I tried to upgrade `lopdf` from version `0.34.0` to `0.35.0`, and the problem was successfully resolved. 

You can reproduce this issue by running the example code in `pdf_agent.rs`. 

After upgrading the version, the problem can be resolved.


0xMochan:
Wow, would not have guessed it was a dependency problem, really thought we were doing somethig wrong, great find!

joshua-mo-143 (approved):
lgtm :) might wait to merge this until next monday (when we release next, probably) since merging this one will likely cause merge conflicts on a number of PRs

joshua-mo-143:
Looks like this one just needs a quick merge conflict resolution on the Cargo.toml file then it's ready to go!

(if you don't have time for this, I can do so on Monday if we're releasing at that time)

lispking:
> Looks like this one just needs a quick merge conflict resolution on the Cargo.toml file then it's ready to go!
> 
> (if you don't have time for this, I can do so on Monday if we're releasing at that time)

done

0xMochan (approved):
lgtm

Issue #310, opened 2025-02-19
bug:  reqwest client can't use socks proxy


nihao:
- [ x] I have looked for existing issues (including closed) about this

## Bug Report

The reqwest client supports using proxies that are set in the environment, including HTTP and SOCKS. However, since the socks feature is not enabled, the environment configuration is not effective.

## Reproduction

set in environment, but not effective

## Expected behavior

reqwest client using proxies that are set in the environment, including socks

joshua-mo-143:
Closing because the PR to close this issue has been merged.

PR #311, opened 2025-02-19
fix: reqwest can not use SOCKS proxy

nihao:
https://github.com/0xPlaygrounds/rig/issues/310

* enable feature of reqwest to support socks proxy

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6880,6 +6880,7 @@ dependencies = [
  "tokio",
  "tokio-native-tls",
  "tokio-rustls 0.24.1",
+ "tokio-socks",
  "tokio-util",
  "tower-service",
  "url",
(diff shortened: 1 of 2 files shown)


0xMochan:
Hey, thanks for this PR! Is there more details you can describe about your use case with SOCK proxies? I could see this being used for an optional feature or a default feature based on the use case!

nihao:
Inside our datacenter, the servers cannot access the internet directly. Instead, we provide a SOCKS proxy by default because it offers better speed and stability. 
--
Maybe you can set it as an optional feature, but it will require a convenient configuration method

joshua-mo-143:
I think the most ideal way to do this would be to add it as a feature since otherwise, it'll be a breaking change and it would be ideal to keep core dependencies low.

nihao:
hi @0xMochan , when will you merge this contribute to the project

joshua-mo-143:
Hi, Mochan's currently OOO at the moment.

If you can set up the socks feature as a separate feature for `rig-core`, I'd be happy to merge it.

nihao:
@joshua-mo-143 pls check now

0xMochan (approved):
Optional is good. it seems like `multipart` is also added but ik #336 also uses that?

joshua-mo-143:
Just had a quick look, is there any specific reason the multipart feature has been added?

Just checking since technically that's out of scope for this PR and would be better suited in a new PR if you have a specific use case for it as there are no other code additions :sweat_smile:

nihao:
> Had a quick look, is there any specific reason the multipart feature has been added?
> 
> Just checking since technically that's out of scope for this PR and would be better suited in a new PR if you have a specific use case for it as there are no other code additions 😅

-----

In fact, I only added the socks feature. However, since the multipart feature was added to the main branch before this PR, I included it to resolve the conflict with the main branch.

joshua-mo-143:
Ah yes, sorry you're right! 

Ok this looks good to me then

Issue #317, opened 2025-02-24
feat: Add support for AWS Bedrock


mkranjac:
## Model Provider Integration Request

Amazon Bedrock is a fully managed service that offers a choice of high-performing foundation models (FMs) from leading AI companies like AI21 Labs, Anthropic, Cohere, Luma, Meta, Mistral AI, poolside (coming soon), Stability AI, and Amazon 
List of supported [models](https://docs.aws.amazon.com/bedrock/latest/userguide/models-supported.html) 

### Resources
[AWS Bedrock](https://aws.amazon.com/bedrock/)

PR #318, opened 2025-02-24
feat: AWS Bedrock provider

mkranjac:
Added new model provider as a separate crate because it depends on AWS SDK. 
Relates to [issue](https://github.com/0xPlaygrounds/rig/issues/317)

### Important:
Access to Amazon Bedrock foundation models **isn't granted by default.** You can request access, or modify access, to foundation models only by using the Amazon Bedrock console.

This is what you would get if you ran all example programs:

[agent_with_bedrock.txt](https://github.com/user-attachments/files/18949238/agent_with_bedrock.txt)
[document_with_bedrock.txt](https://github.com/user-attachments/files/18949239/document_with_bedrock.txt)
[embedding_with_bedrock.txt](https://github.com/user-attachments/files/18949240/embedding_with_bedrock.txt)
[extractor_with_bedrock.txt](https://github.com/user-attachments/files/18949241/extractor_with_bedrock.txt)
[image_with_bedrock.txt](https://github.com/user-attachments/files/18949242/image_with_bedrock.txt)
[rag_with_bedrock.txt](https://github.com/user-attachments/files/18949243/rag_with_bedrock.txt)

### DIFF
--- a/rig-bedrock/src/completion.rs
+++ b/rig-bedrock/src/completion.rs
@@ -135,8 +135,7 @@ impl completion::CompletionModel for CompletionModel {
         let response = converse_builder
             .send()
             .await
-            .map_err(|sdk_error| AwsSdkConverseError(sdk_error).into())
-            .map_err(|e: CompletionError| e)?;
+            .map_err(|sdk_error| Into::<CompletionError>::into(AwsSdkConverseError(sdk_error)))?;
 
         AwsConverseOutput(response).try_into()
     }
--- a/rig-bedrock/src/completion.rs
+++ b/rig-bedrock/src/completion.rs
@@ -1,0 +1,143 @@
+//! All supported models <https://docs.aws.amazon.com/bedrock/latest/userguide/models-supported.html>
+use crate::{
+    client::Client,
+    types::{
+        assistant_content::AwsConverseOutput, completion_request::AwsCompletionRequest,
+        errors::AwsSdkConverseError,
+    },
+};
+
+use rig::completion::{self, CompletionError};
+
+/// `amazon.nova-canvas-v1:0`
+pub const AMAZON_NOVA_CANVAS: &str = "amazon.nova-canvas-v1:0";
+/// `amazon.nova-lite-v1:0`
+pub const AMAZON_NOVA_LITE: &str = "amazon.nova-lite-v1:0";
+/// `amazon.nova-micro-v1:0`
+pub const AMAZON_NOVA_MICRO: &str = "amazon.nova-micro-v1:0";
+/// `amazon.nova-pro-v1:0`
+pub const AMAZON_NOVA_PRO: &str = "amazon.nova-pro-v1:0";
+/// `amazon.rerank-v1:0`
+pub const AMAZON_RERANK_1_0: &str = "amazon.rerank-v1:0";
+/// `amazon.titan-image-generator-v1`
+pub const AMAZON_TITAN_IMAGE_GENERATOR_V1: &str = "amazon.titan-image-generator-v1";
+/// `amazon.titan-image-generator-v2:0`
+pub const AMAZON_TITAN_IMAGE_GENERATOR_V2_0: &str = "amazon.titan-image-generator-v2:0";
+/// `amazon.titan-text-express-v1`
+pub const AMAZON_TITAN_TEXT_EXPRESS_V1: &str = "amazon.titan-text-express-v1";
+/// `amazon.titan-text-lite-v1`
+pub const AMAZON_TITAN_TEXT_LITE_V1: &str = "amazon.titan-text-lite-v1";
+/// `amazon.titan-text-premier-v1:0`
+pub const AMAZON_TITAN_TEXT_PREMIER_V1_0: &str = "amazon.titan-text-premier-v1:0";
+
+/// `anthropic.claude-3-haiku-20240307-v1:0`
+pub const ANTHROPIC_CLAUDE_3_HAIKU: &str = "anthropic.claude-3-haiku-20240307-v1:0";
+/// `anthropic.claude-3-opus-20240229-v1:0`
+pub const ANTHROPIC_CLAUDE_3_OPUS: &str = "anthropic.claude-3-opus-20240229-v1:0";
+/// `anthropic.claude-3-sonnet-20240229-v1:0`
+pub const ANTHROPIC_CLAUDE_3_SONNET: &str = "anthropic.claude-3-sonnet-20240229-v1:0";
+/// `anthropic.claude-3-5-haiku-20241022-v1:0`
+pub const ANTHROPIC_CLAUDE_3_5_HAIKU: &str = "anthropic.claude-3-5-haiku-20241022-v1:0";
+/// `anthropic.claude-3-5-sonnet-20241022-v2:0`
+pub const ANTHROPIC_CLAUDE_3_5_SONNET_V2: &str = "anthropic.claude-3-5-sonnet-20241022-v2:0";
+/// `anthropic.claude-3-5-sonnet-20240620-v1:0`
+pub const ANTHROPIC_CLAUDE_3_5_SONNET: &str = "anthropic.claude-3-5-sonnet-20240620-v1:0";
+/// `anthropic.claude-3-7-sonnet-20250219-v1:0`
+pub const ANTHROPIC_CLAUDE_3_7_SONNET: &str = "anthropic.claude-3-7-sonnet-20250219-v1:0";
+/// `cohere.command-light-text-v14`
+pub const COHERE_COMMAND_LIGHT_TEXT: &str = "cohere.command-light-text-v14";
+/// `cohere.command-r-plus-v1:0`
+pub const COHERE_COMMAND_R_PLUS: &str = "cohere.command-r-plus-v1:0";
+/// `cohere.command-r-v1:0`
+pub const COHERE_COMMAND_R: &str = "cohere.command-r-v1:0";
+/// `cohere.command-text-v14`
+pub const COHERE_COMMAND: &str = "cohere.command-text-v14";
+/// `cohere.rerank-v3-5:0`
+pub const COHERE_RERANK_V3_5: &str = "cohere.rerank-v3-5:0";
+/// `luma.ray-v2:0`
+pub const LUMA_RAY_V2_0: &str = "luma.ray-v2:0";
+/// `meta.llama3-8b-instruct-v1:0`
+pub const LLAMA_3_8B_INSTRUCT: &str = "meta.llama3-8b-instruct-v1:0";
+/// `meta.llama3-70b-instruct-v1:0`
+pub const LLAMA_3_70B_INSTRUCT: &str = "meta.llama3-70b-instruct-v1:0";
+/// `meta.llama3-1-8b-instruct-v1:0`
+pub const LLAMA_3_1_8B_INSTRUCT: &str = "meta.llama3-1-8b-instruct-v1:0";
+/// `meta.llama3-1-70b-instruct-v1:0`
+pub const LLAMA_3_1_70B_INSTRUCT: &str = "meta.llama3-1-70b-instruct-v1:0";
+/// `meta.llama3-1-405b-instruct-v1:0`
+pub const LLAMA_3_1_405B_INSTRUCT: &str = "meta.llama3-1-405b-instruct-v1:0";
+/// `meta.llama3-2-1b-instruct-v1:0`
+pub const LLAMA_3_2_1B_INSTRUCT: &str = "meta.llama3-2-1b-instruct-v1:0";
+/// `meta.llama3-2-3b-instruct-v1:0`
+pub const LLAMA_3_2_3B_INSTRUCT: &str = "meta.llama3-2-3b-instruct-v1:0";
+/// `meta.llama3-2-11b-instruct-v1:0`
+pub const LLAMA_3_2_11B_INSTRUCT: &str = "meta.llama3-2-11b-instruct-v1:0";
+/// `meta.llama3-2-90b-instruct-v1:0`
+pub const LLAMA_3_2_90B_INSTRUCT: &str = "meta.llama3-2-90b-instruct-v1:0";
+/// `meta.llama3-3-70b-instruct-v1:0`
+pub const LLAMA_3_2_70B_INSTRUCT: &str = "meta.llama3-3-70b-instruct-v1:0";
+/// `mistral.mistral-7b-instruct-v0:2`
+pub const MISTRAL_7B_INSTRUCT: &str = "mistral.mistral-7b-instruct-v0:2";
+/// `mistral.mistral-large-2402-v1:0`
+pub const MISTRAL_LARGE_24_02: &str = "mistral.mistral-large-2402-v1:0";
+/// `mistral.mistral-large-2407-v1:0`
+pub const MISTRAL_LARGE_24_07: &str = "mistral.mistral-large-2407-v1:0";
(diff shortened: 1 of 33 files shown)


0xMochan:
Hey, thanks for this PR that tackles #317 :tada:, we'll get a review going on this soon but in the meantime, do compare to the other model providers to ensure you've hit all the right traits and patterns!

rig-bedrock/src/types/mod.rs
0xMochan, line 1: The methods here are great, but we prefer to leverage the `From` and `TryFrom` traits to perform all of these type conversions. top level is provider `Message` -> `completion::Message` and then you can also implement the downstream trait impls from there!

Here, we would like to use that pattern, as see in the other providers.

rig-bedrock/src/completion.rs
0xMochan, line 1: Don't forget tests! These are usually just mocking the response and testing the serializsation/deserialization aptly the Message <-> Message conversions!

0xMochan (changes requested):
Hey, great work on this PR. Ik working with some of the AWS libraries and tech requires some boilerplate with their libraries so it's pretty cool to have some integrations with. Pointed out some things we do with our other providers that should be matched here.

Feel free to join our discord if you'd like to chat with us more conversationally about these things otherwise you can leave questions or other things here!

rig-bedrock/src/completion.rs
mkranjac: Thanks for the review! 
I changed things as mentioned in PR comments (had to do newtype wrapping because I ran into orphan rule problems) 

Will add more tests soon.

rig-bedrock/src/types/assistent_content.rs
```diff
@@ -0,0 +1,113 @@
...
+            .try_into()?;
+
+        let choice = match message.0 {
+            completion::Message::Assistant { content } => Ok(content),
+            _ => Err(CompletionError::ResponseError(
+                "Response contained no message or tool call (empty)".to_owned(),
+            )),
+        }?;
+
+        if let Some(tool_use) = choice.iter().find_map(|content| match content {
```
mkranjac, line 45: @0xMochan can you please explain why I have to "pick" tool call here in order to get model to execute tool ?

I am aware things will change around tool calls soon in order to support tool results get back to model so I didn't want to poke too much things around here...

edisontim:
Hey! Really nice feature here! Really needed on my end so it's nice to see a PR already open for it :)

rig-bedrock/src/types/assistent_content.rs
0xMochan: So you don't have to pick a tool call here, you can process all of the tool calls and convert them to generic forms of tool calls -- the main model can handle it. Something like:

```rs
let mut content = content
    .iter()
    .map(|c| match c {
        AssistantContent::Text { text } => completion::AssistantContent::text(text),
        AssistantContent::Refusal { refusal } => {
            completion::AssistantContent::text(refusal)
        }
    })
    .collect::<Vec<_>>();

content.extend(
    tool_calls
        .iter()
        .map(|call| {
            completion::AssistantContent::tool_call(
                &call.function.name,
                &call.function.name,
                call.function.arguments.clone(),
            )
        })
        .collect::<Vec<_>>(),
);
Ok(content)
```

Currently, rig does not actually call multiple tools, but I have a branch that I'm currently working on to enable that behavior. Returning multiple tools in the general completion response will pass that to the agent `completion` code which will then dispatch tool-calling.

edisontim:
I think there might be a better way of implementing this btw. The Rust types of the bedrock library are very strict and limiting for a lot of features (e.g. prompt caching) when using the Converse API, but if this crate can instead just act as a provider over an existing agent, the wrapper object can just use the `Invoke` api which takes a json body and allows any field there

edisontim:
I've made the changes on this PR which also has some MCP tools integration: https://github.com/0xPlaygrounds/rig/pull/328. It seems cleaner that way, what do you guys think?

mkranjac:
> I think there might be a better way of implementing this btw. The Rust types of the bedrock library are very strict and limiting for a lot of features (e.g. prompt caching) when using the Converse API, but if this crate can instead just act as a provider over an existing agent, the wrapper object can just use the `Invoke` api which takes a json body and allows any field there

Hi, thank you for helping with this PR!
I agree with you that using directly “Invoke API” has maximum flexibility over Converse API, however with invoke API changing model requires changes on both request and response and that is something that I wanted to avoid. 
Given the fact that AWS has huge numbers of models, that could be tiresome  to write manually and more error prone. 
Even AWS docs strongly suggest using Converse API over Invoke API for that reasons.

I would leave this decision for core maintainers to decide. @0xMochan

mkranjac:
> I've made the changes on this PR which also has some MCP tools integration: #328. It seems cleaner that way, what do you guys think?

This PR include some breaking changes to core package, I think that deserves it's own issue and PR

edisontim:
> however with invoke API changing model requires changes on both request and response and that is something that I wanted to avoid.

What do you mean by this? With the wrapper I made you can use one of the already existing `CompletionModel`s of the already implemented providers, `build_completion` then `try_into` the response and the return type will be the same as the base provider

mkranjac:
> > however with invoke API changing model requires changes on both request and response and that is something that I wanted to avoid.
> 
> What do you mean by this? With the wrapper I made you can use one of the already existing `CompletionModel`s of the already implemented providers, `build_completion` then `try_into` the response and the return type will be the same as the base provider

Yes, using models already implemented in Rig works, but I wanted to use Amazon Titan models and flexibility to swap them with any other Bedrock FM later on.

[`prompt-caching` ](https://aws.amazon.com/bedrock/prompt-caching/) is still in preview mode but will have converse-api support once it's released

rig-bedrock/src/completion.rs
mkranjac: @0xMochan I added some tests and refactored code a bit, also added streaming support.

mkranjac: fixed broken docs

rig-bedrock/src/streaming.rs
```diff
@@ -0,0 +1,108 @@
...
+            .set_inference_config(request.inference_config())
+            .set_tool_config(tool_config)
+            .set_system(request.system_prompt())
+            .set_messages(Some(prompt_with_history));
+
+        let response = converse_builder
+            .send()
+            .await
+            .map_err(|sdk_error| AwsSdkConverseStreamError(sdk_error).into())
+            .map_err(|e: CompletionError| e)?;
```
0xMochan, line 43: Is this secondary `.map_err` necessary, seems redundant?

rig-bedrock/src/streaming.rs
```diff
@@ -0,0 +1,108 @@
...
+                                Err(e) => {
+                                    yield Err(CompletionError::from(e));
+                                }
+                            }
+
+                        }
+                    },
+                    // aws_bedrock::ConverseStreamOutput::MessageStart(message_start_event) => todo!(),
+                    // aws_bedrock::ConverseStreamOutput::MessageStop(message_stop_event) => todo!(),
+                    // aws_bedrock::ConverseStreamOutput::Metadata(converse_stream_metadata_event) => todo!(),
```
0xMochan, line 102: We try not to leave commented blocks of code in final releases. Either leave a `TODO` comment explaining or cut these out!

rig-bedrock/src/types/document.rs
```diff
@@ -0,0 +1,168 @@
...
+impl TryFrom<aws_bedrock::DocumentBlock> for RigDocument {
+    type Error = CompletionError;
+
+    fn try_from(value: aws_bedrock::DocumentBlock) -> Result<Self, Self::Error> {
+        let media_type = match value.format {
+            aws_bedrock::DocumentFormat::Csv => Ok(DocumentMediaType::CSV),
+            aws_bedrock::DocumentFormat::Html => Ok(DocumentMediaType::HTML),
+            aws_bedrock::DocumentFormat::Md => Ok(DocumentMediaType::MARKDOWN),
+            aws_bedrock::DocumentFormat::Pdf => Ok(DocumentMediaType::PDF),
+            aws_bedrock::DocumentFormat::Txt => Ok(DocumentMediaType::TXT),
```
0xMochan, line 60: you could have this as a specific `TryFrom` or `From` implementation though I realize how annoying that can be to write for every single little struct / enum used. Up to you!

rig-bedrock/src/types/message.rs
```diff
@@ -0,0 +1,130 @@
...
+
+use super::{
+    assistant_content::RigAssistantContent,
+    user_content::{RigUserContent, UserContentWithPrompt},
+};
+
+pub struct MessageWithPrompt {
+    pub message: Message,
+    pub prompt: Option<String>,
+}
```
0xMochan, line 17: I'm a bit confused on why `MessageWithPrompt` is required for conversions here. Can you expand on the reasoning here? For some context, there's a bit of a refactor in an upcoming branch that actually removes `prompt` from `CompletionRequest` (since completion endpoints just add it to message history). Is AWS different here or is there significance with how prompt is handling. some insight would be helpful!

The main usecase of rig's Message <-> provider Message conversion is to provide ergonomics when maintaining a provider agnostic message history (which we'll expand on in the future). This struct might make that reallly awkward for end-users when using `aws-bedrock`.

rig-bedrock/src/types/message.rs
```diff
@@ -0,0 +1,130 @@
...
+                        .collect::<Result<Vec<aws_bedrock::ContentBlock>, _>>()?,
+                ))
+                .build()
+                .map_err(|e| CompletionError::RequestError(Box::new(e)))?,
+        };
+        Ok(result)
+    }
+}
+
+pub struct RigMessage(pub Message);
```
0xMochan, line 59: Ditto with above in terms of extra boxing here.

0xMochan (changes requested):
I left a couple of comments. The big one is namely the `MessageWithPrompt` structure that is the basis of the aws-bedrock Message <-> Rig Message conversion. I'm requesting changes but primarily need some insight on the decision making there.

Also, I realize you've had this branch out for a long time and I truely appreciate your patience and throughput in keeping this rebased on main frequently. I hope we can support your amazing work by getting this PR out swiftly now so I'm attaching my personal priority on this!

rig-bedrock/src/types/completion_request.rs
```diff
@@ -0,0 +1,90 @@
...
+        let mut full_history: Vec<Message> = Vec::new();
+        full_history.append(&mut chat_history);
+        full_history.push(prompt_with_context);
+
+        full_history
+            .into_iter()
+            .map(|message| {
+                MessageWithPrompt {
+                    message,
+                    prompt: self.0.preamble.to_owned(),
```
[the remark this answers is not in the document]
mkranjac: this was a mistake, I removed this

rig-bedrock/src/streaming.rs
mkranjac: because of type inference, one liner was not possible

I replaced it  now with 
```.map_err(|sdk_error| { Into::<CompletionError>::into(AwsSdkConverseStreamError(sdk_error)) })?;```

not sure how to make it better....

File: rig-bedrock/src/types/message.rs (hunk 5 above)
Comment (mkranjac) [reply rig-bedrock/src/types/message.rs]: I removed this one, 

Reason why this prompt fields was here in the first place was one weird AWS limitation with Document content block, where prompt is required to be passed next to document block which I [hacked](https://github.com/0xPlaygrounds/rig/pull/318/files#diff-0b0789ad00b80802729396fc06889f65105013c430aa5d5fbbcbc475c0e411eeR83) now

File: rig-bedrock/src/types/message.rs (hunk 6 above)
Comment (mkranjac) [reply rig-bedrock/src/types/message.rs]: extra boxing was unfortunately required everywhere because in this crate both aws and rig structs are foreign types and I ran into orphan rule

Comment (0xMochan) [reply rig-bedrock/src/types/message.rs]: ah yea, this one makes sense

File: rig-bedrock/src/streaming.rs (hunk 2 above)
Comment (0xMochan) [reply rig-bedrock/src/streaming.rs]: I see, the `e: CompletionError` was used in the `.into`. yea, you can also use the `From::` form as well. not as nice looking but it's atleast directly showcasing the conversion.

File: rig-bedrock/src/types/completion_request.rs [hunk 8]
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    pub fn system_prompt(&self) -> Option<Vec<SystemContentBlock>> {
+        self.0
+            .preamble
+            .to_owned()
+            .map(|system_prompt| vec![SystemContentBlock::Text(system_prompt)])
+    }
+
+    pub fn prompt_with_history(&self) -> Result<Vec<aws_bedrock::Message>, CompletionError> {
```
0xMochan, line 72: I'll likely be changing this in my upcoming branch but this will be merged prior.

0xMochan (approved):
Looks great, my comment is merely a .. comment :)
I appreciate all the hard work put in this branch, and I noticed even the latest image generation stuff, pretty nifty.

For now, I still need to acquire an AWS bedrock key to test but i'll ask if we can feature freeze this branch. If there are new features you want to do, I'll request if you can spin a new branch from this, thanks!

---

There _is_ a merge conflict, with `cargo.lock` which might be due to the latest shenanigans with chrono and arrow dependencies, i'll see if i can recommend a good solution.

mkranjac:
> Looks great, my comment is merely a .. comment :) I appreciate all the hard work put in this branch, and I noticed even the latest image generation stuff, pretty nifty.
> 
> For now, I still need to acquire an AWS bedrock key to test but i'll ask if we can feature freeze this branch. If there are new features you want to do, I'll request if you can spin a new branch from this, thanks!
> 
> There _is_ a merge conflict, with `cargo.lock` which might be due to the latest shenanigans with chrono and arrow dependencies, ~i'll see if i can recommend a good solution.~ I think it's mostly just some updates from main that conflict with the new packages added here but if there is a commit that changes a version of anything `arrow`, `chrono`, or `deranged` -- it's probably needs to be manually adjusted (like if u delete the `cargo.lock` file, things might get wacky).

Fixed merge conflict.

mkranjac:
@0xMochan I wanted to remove relative path for `rig-core` on `rig-bedrock` crate so that it stays on 0.10.0 and not to slowdown development of rig-core with breaking changes, but it seems feature flag `image` was not available on [rig-core](https://crates.io/crates/rig-core/versions?sort=date) even though it's the same version

0xMochan:
> @0xMochan I wanted to remove relative path for `rig-core` on `rig-bedrock` crate so that it stays on 0.10.0 and not to slowdown development of rig-core with breaking changes, but it seems feature flag `image` was not available on [rig-core](https://crates.io/crates/rig-core/versions?sort=date) even though it's the same version

The `image` feature flag doesn't exist in `rig-core`

mkranjac:
> > @0xMochan I wanted to remove relative path for `rig-core` on `rig-bedrock` crate so that it stays on 0.10.0 and not to slowdown development of rig-core with breaking changes, but it seems feature flag `image` was not available on [rig-core](https://crates.io/crates/rig-core/versions?sort=date) even though it's the same version
> 
> The `image` feature flag doesn't exist in `rig-core`

This one:
https://github.com/0xPlaygrounds/rig/blob/main/rig-core/Cargo.toml#L59

0xMochan:
All examples confirmed tested with `amazon.nova-lite-v1:0`, `amazon.nova-canvas-v1:0`, and `anthropic.claude-3-7-sonnet-20250219-v1:0` @cvauclair !

cvauclair:
@mkranjac @0xMochan are we waiting for a final commit or can I merge as-is?

0xMochan (approved):

PR #324, opened 2025-02-27
Add reasoning content for deepseek R1 API and tackle problems of tool_calls

VanderBieu:
Add reasoning content for deepseek R1 API and tackle problems of tool_calls may cause chat agent crash


VanderBieu:
Also I made it possible to stream response of deepseek models

joshua-mo-143:
Superceded by #557 #370

PR #327, opened 2025-02-28
fix(openai): skip serializing empty tool_calls vector

GustavoWidman:
OpenAI was returning an error in completion requests because outgoing "tool_calls" vector was being serialized as "[]" (empty vec) if there were no tool calls. This should fix it :)

Error that was being received:

```
 |  ProviderError: {
 |    "error": {
 |      "message": "Invalid 'messages[2].tool_calls': empty array. Expected an array with minimum length 1, but got an empty array instead.",
 |      "type": "invalid_request_error",
 |      "param": "messages[2].tool_calls",
 |      "code": "empty_array"
 |    }
 |  }
 ```

### DIFF
--- a/rig-core/src/providers/openai.rs
+++ b/rig-core/src/providers/openai.rs
@@ -491,7 +491,11 @@ pub enum Message {
         audio: Option<AudioAssistant>,
         #[serde(skip_serializing_if = "Option::is_none")]
         name: Option<String>,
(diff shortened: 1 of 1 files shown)


0xMochan:
Ah, great catch. Not sure how this slipped through, will get this out ASAP! These message models from the providers are fairly tricky to get perfect 😅

M4n5ter:
deepseek provider also has this issue

PR #329, opened 2025-02-28
feat: Azure OpenAI Token Authentication

Verticaleap:
# Overview

This pull requests extends the Azure OpenAI Client to provide authentication via token.

### References:
https://learn.microsoft.com/en-us/azure/ai-services/openai/reference#authentication

### DIFF
--- a/rig-core/src/providers/azure.rs
+++ b/rig-core/src/providers/azure.rs
@@ -32,6 +32,12 @@ pub struct Client {
     http_client: reqwest::Client,
 }
 
+#[derive(Clone)]
+pub enum AzureOpenAIAuth {
+    ApiKey(String),
+    Token(String),
+}
+
 impl Client {
     /// Creates a new Azure OpenAI client.
     ///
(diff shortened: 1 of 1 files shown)


0xMochan:
Thanks for the PR, most of the team is traveling but we'll get this reviewed as soon as we are back!

rig-core/src/providers/azure.rs
```diff
@@ -32,36 +32,60 @@ pub struct Client {
...
     /// Creates a new Azure OpenAI client.
     ///
     /// # Arguments
     ///
-    /// * `api_key` - Azure OpenAI API key required for authentication
+    /// * `auth` - Azure OpenAI API key or token required for authentication
     /// * `api_version` - API version to use (e.g., "2024-10-21" for GA, "2024-10-01-preview" for preview)
     /// * `azure_endpoint` - Azure OpenAI endpoint URL, for example: https://{your-resource-name}.openai.azure.com
-    pub fn new(api_key: &str, api_version: &str, azure_endpoint: &str) -> Self {
+    pub fn new(auth: AzureOpenAIAuth, api_version: &str, azure_endpoint: &str) -> Self {
```
0xMochan, line 49: If there's two forms of auth, I think there should be 2 additional methods that take in either `api_key: &str` or `token: &str`.

`new` would be the same, taking in `auth` (could also consider changing into `impl Into<AzureOpenAIAuth>` choosing a reasonable default for strings too for backwards compat).

Then `from_api_key` and `from_token` would define specific versions of the enum directly which is self-documenting and also user-friendly (no need to deal with the enum directly).

0xMochan: All docs with `Client::new` should either be reverted if `impl Into<...>` is added OR using one of the new constructors being added.

0xMochan (changes requested):
Mostly some ergonomic suggesitions

rig-core/src/providers/azure.rs
Verticaleap: I like these ideas! I just pushed your recommended changes.

joshua-mo-143 (approved):
Hi Verticaleap, thanks for submitting this PR! 

LGTM now that the required amendments have been made :slightly_smiling_face:

Issue #312, opened 2025-02-19
feat: Is it possible to split out a rig-providers as a new crate? This seems like it would be very useful for community needs


Cupnfish:
- [x] I have looked for existing issues (including closed) about this

## Feature Request
I think if we provide a rig-providers as a new crate, other projects in the community can integrate it well, which would definitely be beneficial for project maintenance. I looked into existing projects in the community, such as [llm](https://github.com/graniet/llm), which appears to have many features, but in reality, a lot of them are not implemented.

### Motivation
I want to build a chat GUI software llm by myself for fun, but I haven't found a suitable package. Rig is too big for me, I don't need so many proxy functions. The current core doesn't seem streamlined enough, and some parts could be developed as a separate crate.

### Proposal
It feels like splitting the providers module into a separate crate would solve the problem

### Alternatives
This is just a wish I have out of a desire for a streamlined providers package, hoping that rig can separate this module as a high-quality providers library that the community can directly use

0xMochan:
This is something that came up internally. It's sorta tricky bc ideally, we want something straightforward for people to install `cargo add rig-core tokio` and then u can run an example. it would end up changing to `cargo add rig-core rig-providers tokio` for things to continue.

I _think_ something like keeping atleast `openai` in `rig-core` and perhaps `anthropic` + `gemini` in core and we can throw the other providers in an side-crate would be suitable. I will say, I don't see how a `rig-providers` crate would be usable outside the community unless we spun out the `completion` method as it's own thing and added some stuff for `rig-core` to be an extra feature which would add some maintenance burden to the team for some unknown benefit. The main benefit of rig's shared traits and structs is to generalize method with each of the providers.

I definitely can see how the inclusion of abstraction layers like agents w/ tool calling sorta can be seen as a heavier integration so something where the providers can be used standalone is enticing. From our POV, if you desire a closer connection to the LLM interface, the `CompletionModel` trait via `.completion` method is a _direct_ way to talk to the LLM with no interpretation of embeddings, RAGing, tool calling, etc. The only transformation that occurs is the direct generalization from the provider's message model to the generalized one.

---

TL;DR: It's def. on our radar with having a `rig-providers`. In terms of having `rig-core` being an optional features for said side-crate, it's likely not our highest priority.

Issue #236, opened 2025-01-23
epic: Multimodal (audio and image generation) APIs


cvauclair:
## Feature Request
Define audio and image generation API to enable multi-modal use cases.

### Motivation
While text completion is a major use case of generative AIs, it is far from the only one. Audio and image generation are among the top other use cases of generative AIs and having the ability to combine all three types of generation to create complex AI pipelines is a major design goal of Rig.

Rig already provides a generalized text completion API (i.e.: completion request and response API which can be converted to and from provider-specific API), but provides no equivalent for audio and image generation models. 

This issue will be used to track all multi-modal related discussions, issues and PRs.

### Proposal
Create a common audio and image generation API and integrate a first batch of providers for each. This can (and should) be done across multiple PRs and will be implemented progressively into the library.

❗ IMPORTANT: Although we are making this plan public to encourage discussions and feedback for the sake of transparency, we are not expecting these features to come entirely from the community nor for them to be available in the short term. We want to get the design of these new APIs right (well mostly right) and will prioritize quality over speed. If you do want to work on these APIs, please comment below or reach out to us on our Discord! Also expect a thorough reviewing process for any PRs submitted in the context of this epic.

#### Step 1: Define common API
The first step in integrating other types of generation is to define common APIs for both audio and image generation. These APIs should roughly follow the same structure as the existing text completion API (to some extent). For example, for the audio generation API:
```rust
// in src/audio_generation.rs
pub enum AudioGenerationError {...}

pub struct AudioGenerationRequest {...}

pub struct AudioGenerationResponse {...}

pub trait AudioGenerationModel: Clone + Send + Sync {
    /// The raw response type returned by the underlying audio generation model.
    type Response: Send + Sync;

    /// Generates an audio generation response for the given audio generation request.
    fn generate_audio(
        &self,
        request: AudioGenerationRequest,
    ) -> impl std::future::Future<Output = Result<AudioGenerationResponse<Self::Response>, AudioGenerationError>>
           + Send;
}

pub struct AudioGenerationRequest<M: AudioGenerationModel> {...}
```

The same for the image generation API in `image_generation.rs`.

❗ IMPORTANT: The data types above should be **general** and contain the minimal attributes that can be converted to _any_ audio generation model provider API.

❗ IMPORTANT: Unless these new APIs require no additional dependencies (which would be surprising), they should be locked behind `audio` and `image` feature flags respectively so as to keep `rig-core` as lean as possible.  

[...]

PR #331, opened 2025-03-03
refactor: Split OpenAI into multiple files

yavens:
This PR breaks up OpenAI into multiple smaller files, similar to some of the other providers. Most important types are reexported at the top of the module so it shouldn't be a difficult migration for most users.

As progress is made on #236 and #312, it would be nice to make this change for all existing provider implementations.


0xMochan:
Is this designed to go before or after the #336 ?

yavens:
It doesn't particularly matter since OAI transcription was merged in the initial feature PR.

joshua-mo-143:
Is this something we are still wanting to prioritise at some point this week?

yavens:
I think it should just be rolled into the image generation PR now.

joshua-mo-143:
Ok no worries - let's close this after the image gen PR is done then

0xMochan:
Superseded by #357

Issue #320, opened 2025-02-24
feat: standardized streaming support


0xMochan:
- [x] I have looked for existing issues (including closed) about this

## Feature Request
We currently have [limited streaming support](https://github.com/0xPlaygrounds/rig/pull/232) for Anthropic and much of the code and traits are duplicated to support chunked messages even though the processing and majority of the internal code can be shared.

There is likely a natural way to promote a streaming-styled set of traits that can reuse partial components from the normal generation traits that can be implement across the rest of the providers.

### Motivation
Streaming is the most natural way to incorporate LLM responses in client applications. While it's application in RAG can be tricky, not providing ubiquitous streaming support throughout our ecosystem leaves us missing an entire category of LLM projects!

### Proposal
- Study streaming support across the ecosystem and refactor the `completion` modules to reuse aspects of `Message` etc.
  - This includes being able to produce `Message` types from streaming outputs so that it can be easily piped into message histories, etc.
  - Also includes generalizing aspects of `completion` into other methods so that `streaming_completion` can reuse`.
- Implement streaming across all of our providers (openai, gemini, etc.)
  - Likely should be split up PR by PR (esp after the #312 issue is resolved).

### Alternatives
Continue with how Anthropic has it implemented with silo'd support.

0xMochan:
## Update
Streaming is supported across all providers that support it.

- [ ] Cohere is still missing streaming due to PR merge order

In terms of standardized traits, due to the work in #347 , most providers spun duplicate behavior into internal methods that create provider specific completion requests.

PR #332, opened 2025-03-03
feat: SSE/JSONL decoders ported from Anthropic TS SDK

piotrostr:
This fixes the issue with trailing whitespace JSONL and improves the flow of accumulating/streaming chunked outputs

The logic can likely be reused with minimal change for other providers

### DIFF
--- a/rig-core/src/providers/anthropic/decoders/mod.rs
+++ b/rig-core/src/providers/anthropic/decoders/mod.rs
@@ -2,7 +2,7 @@
  * The code from this module is a Rust port of the
  * https://github.com/anthropics/anthropic-sdk-typescript/tree/main decoders
  *
- * The original code is licensed under PBT license
+ * The original code is licensed under MIT license
  * https://github.com/anthropics/anthropic-sdk-typescript/blob/main/LICENSE
  */
 pub mod jsonl;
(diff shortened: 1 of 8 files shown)


rig-core/src/providers/anthropic/decoders/mod.rs
```diff
@@ -0,0 +1,10 @@
+/*
+ * The code from this module is a Rust port of the
+ * https://github.com/anthropics/anthropic-sdk-typescript/tree/main decoders
+ *
+ * The original code is licensed under PBT license
```
0xMochan, line 5: This looks like MIT license

0xMochan (approved):
Good adaptation from the anthropic typescript library. I think we'll take a closer look on how decoding is handled here when we lift it to beyond anthropic APIs in an upcoming #320 PR.

Minor nitpick on PBT -> MIT license in comment.

rig-core/src/providers/anthropic/decoders/mod.rs
piotrostr: updated!

0xMochan (approved):
Good use of `impl Into<...>`

joshua-mo-143:
lgtm, thanks Piotrek!

PR #335, opened 2025-03-05
added mcp_tool + Example

stevohuncho:
Minimal implementation of MCP tooling via [mcp-core](https://crates.io/crates/mcp-core). Includes an example of starting an MCP Server, creating an MCP client, and adding those MCP tools to a rig agent and executing a tool call.

Currently using `mcp-core` dep while the official `mcp/rust-sdk` is in development as it is not currently on crates.io. `mcp-core` is contributing its efforts to the official repo in the meantime.


rig-core/src/tool.rs
```diff
@@ -185,6 +185,117 @@ impl<T: Tool> ToolDyn for T {
...
+        Box::pin(async move {
+            let result = self
+                .client
+                .call_tool(&name, Some(args))
+                .await
+                .map_err(|e| {
+                    ToolError::ToolCallError(Box::new(McpToolError(format!(
+                        "Tool returned an error: {}",
+                        e
+                    ))))
```
0xMochan, line 249: Since `McpToolError` will commonly be converted into a `ToolCallError`, i would just draft a `impl ToolCallError` for `McpToolError` that lets u write `?` directly.

0xMochan (changes requested):
Looks pretty straightforward. I have other plans on making working with tools in Rig much easier but I think my main suggestion here is using optional features for rig itself.

I think separately, `mcp-core` currently aligns itself to actix and other async web stuff but theoretically, it might be worthwhile to make it framework agnostic (similar to how `rig` doesn't force `tokio`). this could be a future goal!

rig-core/examples/mcp_tool.rs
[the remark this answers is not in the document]
stevohuncho: So you can create a Client that connects to the server but you do not need to spin up a server in the same program that will be using rig. This client can choose a transport layer either SSE or STDIO.

rig-core/src/tool.rs
stevohuncho: just added this now

0xMochan (approved):
looks pretty good!

cvauclair:
Hey @stevohuncho big thanks for the contribution! Would like to merge this asap, could you resolve the merge conflicts so we can push this forward? Thanks!

stevohuncho:
@cvauclair fixed the conflicts and updated the deps for a more recent version of the package allowing for a cleaner example

cvauclair:
@stevohuncho awesome! Now just need to get CI passing :sweat_smile:

0xMochan:
I did some research, this is due to a recent unintentional breaking change made with a chrono patch release in one of our deps.

We'll have to mitigate this with a pinned dep ver but this pr is unblocked, we'll fix it on main.

https://github.com/apache/arrow-rs/issues/7196

0xMochan:
@stevohuncho So we have two options:
1) we pin `arrow=54.2.1` (which has the fix for chrono compat) in rig
2) or we pin in mcp-core

Currently, this wasn't an issue with rig bc our chrono / arrow is before this but somehow, mcp-core's is after.

PR #336, opened 2025-03-05
feat: add transcription to all providers that support it

yavens:
Adds Transcription support to:
- Gemini
- Azure
- Groq

### DIFF
--- a/rig-core/Cargo.toml
+++ b/rig-core/Cargo.toml
@@ -32,8 +32,8 @@ rayon = { version = "1.10.0", optional = true }
 worker = { version = "0.5", optional = true }
 bytes = "1.9.0"
 async-stream = "0.3.6"
-mime_guess = { version = "2.0.5", optional = true }
-base64 = { version = "0.22.1", optional = true}
+mime_guess = { version = "2.0.5"}
+base64 = { version = "0.22.1"}
 
 
 
@@ -52,7 +52,6 @@ pdf = ["dep:lopdf"]
 epub = ["dep:epub", "dep:quick-xml"]
 rayon = ["dep:rayon"]
 worker = ["dep:worker"]
-blob = ["dep:base64", "dep:mime_guess"]
 socks = ["reqwest/socks"]
 
 [[test]]
@@ -98,8 +97,3 @@ required-features = ["derive"]
 [[example]]
 name = "together_embeddings"
 required-features = ["derive"]
-
-[[example]]
-name = "transcription"
-required-features = ["blob_data"]
-
--- a/rig-core/Cargo.toml
+++ b/rig-core/Cargo.toml
@@ -32,6 +32,9 @@ rayon = { version = "1.10.0", optional = true }
 worker = { version = "0.5", optional = true }
 bytes = "1.9.0"
 async-stream = "0.3.6"
+mime_guess = { version = "2.0.5", optional = true }
+base64 = { version = "0.22.1", optional = true}
+
 
 
 [dev-dependencies]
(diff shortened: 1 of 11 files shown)


yavens:
Not sure how to fix the wasm issue, it seems to be an issue with the feature flag I added for transcription since Gemini required a mime-type and base64 content for audio data.

rig-core/Cargo.toml
```diff
@@ -50,6 +52,7 @@ pdf = ["dep:lopdf"]
 epub = ["dep:epub", "dep:quick-xml"]
 rayon = ["dep:rayon"]
 worker = ["dep:worker"]
+blob_data = ["dep:base64", "dep:mime_guess"]
```
0xMochan, line 55: `blob_data` is a weird feature name? maybe just `blob`? i'm not sure, any ideas @cvauclair ?

rig-core/src/providers/gemini/transcription.rs
```diff
@@ -0,0 +1,144 @@
...
+        let mut generation_config = serde_json::from_value::<GenerationConfig>(additional_params)?;
+
+        // Set temperature from completion_request or additional_params
+        if let Some(temp) = request.temperature {
+            generation_config.temperature = Some(temp);
+        }
+
+        let system_instruction = Some(Content {
+            parts: OneOrMany::one(
+                "Translate the provided audio exactly. Do not add additional information.".into(),
```
0xMochan, line 61: Gemini requires system instructions for transcribing?!?
I would 100% store this in a const at the top of the file w/ a name. raw english being apart of code working is weird but it's best to highlight it.

rig-core/src/providers/gemini/transcription.rs
```diff
@@ -0,0 +1,144 @@
...
+            ),
+            role: Some(Role::Model),
+        });
+
+        let mime_type =
+            if let Some(mime) = mime_guess::from_path(Path::new(&request.filename)).first() {
```
0xMochan, line 67: `mime_guess` looks neat, it might be worthwhile for us to switch to using `mime` throughout the entire codebase in general instead of my custom set in the `completion/messages.rs`

0xMochan (changes requested):
Minor tweaks, main thought is renaming `blob_data` -> `blob` for simplicity or maybe there's a better name..

in terms of WASM, i'll ping @joshua-mo-143 who's fixed this in the past!

rig-core/src/providers/gemini/transcription.rs
yavens: Yeah, Gemini is natively multi-modal, so it'll be conversational in the response if you don't prompt it otherwise. Whisper is single purpose, so it's not an issue there.

joshua-mo-143:
Got a feeling the WASM issue not passing is because of the feature gating. You might need to create a second version of all affected structs/etc without `blob` enabled. Otherwise, you could just add it as a default feature for the `rig-core` crate.

yavens:
I figured it out, I just had some imports that weren't gated behind the feature flag.

rig-core/Cargo.toml
```diff
@@ -32,6 +32,9 @@ rayon = { version = "1.10.0", optional = true }
 worker = { version = "0.5", optional = true }
 bytes = "1.9.0"
 async-stream = "0.3.6"
+mime_guess = { version = "2.0.5"}
```
joshua-mo-143, line 35: Is there any chance some context can be provided on why we've removed the feature flag? (just need to clarify as I know Stopher commented specifically on this)

joshua-mo-143 (commented):
Pretty much just one thing that needs to be addressed and that's it :slightly_smiling_face:

rig-core/Cargo.toml
yavens: We talked about just removing the feature flag since `mime_guess` is already in the dependency tree from `reqwest/multipart` and `base64` is completely self contained.

joshua-mo-143: Lgtm then :slightly_smiling_face:

PR #347, opened 2025-03-12
feat: Add streaming to all model providers

yavens:
This PR adds streaming support to all providers with OpenAI-compatible ones using the same streaming logic. 

The first step towards getting #320 done.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7812,6 +7812,7 @@ dependencies = [
  "tokio",
  "tokio-native-tls",
  "tokio-rustls 0.24.1",
+ "tokio-socks",
  "tokio-util",
  "tower-service",
  "url",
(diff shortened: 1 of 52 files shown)


yavens:
It could potentially be standardized further between the OAI-compatible ones; let me know if you have any ideas.

joshua-mo-143:
Maybe something like this?

```rust
use rig::completion::{CompletionModel, CompletionError};

fn send_completion_request<M: CompletionModel>(model: M, request: CompletionRequest) -> Result<reqwest::Response, CompletionError> {
        let mut request = self.create_completion_request(request)?;

        request = merge(request, json!({"stream": true}));

        let response = self
            .client
            .post("/chat/completions")
            .json(&request)
            .send()
            .await?;

        if !response.status().is_success() {
            return Err(CompletionError::ProviderError(format!(
                "{}: {}",
                response.status(),
                response.text().await?
            )));
        }
        
        Ok(res)
}
```

joshua-mo-143 (commented):
lgtm - i've added a suggestion for OAI standardization if you want to add it in

yavens:
Sweet, I made a similar change to the one you suggested; adding streaming is now ~4 lines for most OAI-compatible providers.

PR #348, opened 2025-03-13
Fix wrong handling for Qwen's responses when using tools

qhliang:
Fix empty content field handling when using Qwen2.5 with tools via OpenAI-compatible API. This PR skips processing empty content strings to ensure program stability, as valid tool calls may return empty content in the API response.


joshua-mo-143:
You just need to run `cargo fmt` on this and then it's mergeable :slightly_smiling_face:

qhliang:
> You just need to run `cargo fmt` on this and then it's mergeable 🙂您只需在此运行 `cargo fmt` ，然后就可以合并了 🙂

I'm sorry, I'll execute the required modifications and resubmit.

Issue #315, opened 2025-02-24
feat: Integrate Cohere V2 API


cvauclair:
- [ ] I have looked for existing issues (including closed) about this

## Feature Request
Integrate new version of Cohere API (v2)

### Motivation
The new API is better designed to handle things like tool calling loops and other important reasoning features.

### Proposal
Integrate Cohere V2 API and drop support for their V1 API.

### Alternatives
Keep supporting both API versions

0xMochan:
closed by #350

PR #350, opened 2025-03-14
feat(provider): cohere-v2

0xMochan:
Solves #315 .

## Versioning
So originally, the plan was to support both v1 and v2 but it became clear that v2 is a pure improvement over a badly designed v1 API. In my next run of PRs, I refactor `prompt` being an argument to a provider's implementation by combining it with `message_history` (since that's what the API needs). This is necessary for proper ordering with full tool loops but cohere v1 accepts the prompt as a separate API parameter which was incompatible with this change hence the needed update.

## Breaking
Cohere connectors are no more so a web-search based tool has to be created from scratch akin to other providers. There are also other model changes but nothing on the exterior interface.


joshua-mo-143:
Needs lints fixed, otherwise lgtm.

Have you tested this locally?

joshua-mo-143 (commented):
Needs lints fixed, otherwise lgtm.

Have you tested this locally?

Issue #352, opened 2025-03-16
feat: proc macro for implementing Tool


cbrit:
- [x] I have looked for existing issues (including closed) about this

## Feature Request

Add a proc macro for easily implementing the Tool trait with automatic parameter validation and error handling.

## Motivation

Currently, implementing the Tool trait requires a lot of boilerplate code. Obfuscation can be bad, but the macro can act as a shorthand that has all the same info as a function signature, so nothing is really obfuscated except struct creation. Also, by adding this feature in a separate crate it can be optional and not affect core functionality.

## Proposal

Add a #[rig_tool] proc macro that:
- Automatically generates a Tool implementation from an async function
- Provides attribute-based configuration for:
  - Tool description
  - Parameter descriptions
(tool name is function name automatically)

Example usage:

```rust
#[rig_tool(
    description = "Perform basic arithmetic operations",
    params(
        x = "First number in the calculation",
        y = "Second number in the calculation",
        operation = "The operation to perform"
    )
)]
async fn calculator(x: i32, y: i32, operation: String) -> Result<i32, rig::tool::ToolError> {
    match operation.as_str() {
        "add" => Ok(x + y),
        "subtract" => Ok(x - y),
        "multiply" => Ok(x * y),
        "divide" => {
             if y == 0 {
                 Err(rig::tool::ToolError::ToolCallError(
                     "Division by zero".into(),
                 ))
             } else {
                 Ok(x / y)
             }
         }
         _ => Err(rig::tool::ToolError::ToolCallError(
             format!("Unknown operation: {}", operation).into(),
         )),
     }
}
```

## Alternatives

Manual Implementation: Continue requiring developers to implement the Tool trait manually.

PR #353, opened 2025-03-16
feat: rig-macros crate providing a rig_tool proc macro

cbrit:
Based heavily on the [model context protocol tool generation API](https://github.com/modelcontextprotocol/rust-sdk/blob/main/crates/mcp-macros/src/lib.rs)

Relates to issue #352. Adds the rig-macros crate which contains the `rig_tool` proc_macro for designating a function as a tool with a simple macro.

A struct is generated with a name matching the target function name, and the `rig::tool::Tool` trait is implemented for that struct. The tool name is the function name and isn't currently configurable to something different, though it's a small fix to change that.

The macro parameters are descriptions for the tool and its parameters. 

Requires the return type to be a Result with error type `ToolError`.

Example usage 

```rust
#[rig_tool(
    description = "Perform basic arithmetic operations",
    params(
        x = "First number in the calculation",
        y = "Second number in the calculation",
        operation = "The operation to perform"
    )
)]
async fn calculator(x: i32, y: i32, operation: String) -> Result<i32, rig::tool::ToolError> {
    match operation.as_str() {
        "add" => Ok(x + y),
        "subtract" => Ok(x - y),
        "multiply" => Ok(x * y),
        "divide" => {
             if y == 0 {
                 Err(rig::tool::ToolError::ToolCallError(
                     "Division by zero".into(),
                 ))
             } else {
                 Ok(x / y)
             }
         }
         _ => Err(rig::tool::ToolError::ToolCallError(
             format!("Unknown operation: {}", operation).into(),
         )),
     }
}
```


joshua-mo-143:
Hey @cbrit, thanks for your PR!

I am wondering if we should combine this with the `rig-derive` crate as it would make it much easier to have all macros under one crate? cc @0xMochan @cvauclair

cbrit:
@joshua-mo-143 I can do that! Will wait for feedback from those tagged.

cbrit:
Sounds great thanks @cvauclair

cvauclair:
Hey @cbrit! So we've talked with the developer of the `rig-tool-macro` crate and he has given us the greenlight to merge his work into Rig! 

What we'd like to do is bring your macro to parity with his, is that something you would be open to do in your PR? If not, we can merge your PR as-is and handle the rest ourselves.

Note that in both cases, my comments above still need to be addressed (i.e.: moving your macro to the `rig-core-derive` crate and re-export deps from `rig-core` instead of including them in the macro).

Cheers!

cbrit:
@cvauclair Fantiastic!

I'll move it and work on bringing the PR to parity with his crate. I should be able to get to it in the next couple of days.

cbrit:
Sorry for the delay, work got busy. Will get on this.

joshua-mo-143:
Re-requested review due to aforementioned proc_macro issues

cbrit:
@cvauclair You're so right. I think maybe what I was seeing was an error due to placing it in a module file instead of the root `lib.rs` and I didn't read closely.

I moved everything over and added a doc comment!

rig-core/src/lib.rs
```diff
@@ -98,6 +98,7 @@ pub mod streaming;
 pub mod tool;
 pub mod transcription;
 pub mod vector_store;
+pub use serde_json;
```
cvauclair, line 101: I'd rather we don't re-export `serde_json` here since that could create a whole lot of issues down the line

cbrit: Would you mind clarifying? Earlier we discussed re-exporting from `rig-core` so that the generated code would not require the user to have `serde_json` in their Cargo.toml. I removed schemars entirely, and I simlified the `call` impl to not need serde_json, but unfortunately `ToolDefinition` requires that its parameters field be a serde_json::value::Value. Should I just re-export from `rig-core-derive` instead?

cvauclair: Yeah apologies, you're right that I'm contradicting myself here. But after thinking and reading about it (here is a small [blogpost](https://brokenco.de/2023/07/26/rust-re-export.html) talking about when to re-export and when not to), I'm not convinced that `serde_json` should be re-exported from any of our crates. If the user wants to use the macro, they would just need to also install `serde_json`.

If this becomes too much of a hassle we can always add the re-export later on. But adding it now and removing it would break the API, which we're trying to avoid as much as possible even if the crate is still in pre-1.0.0.

cbrit: Makes sense! `serde_json` is so common that it has a higher likelihood of creating those version mismatches.

cbrit:
@cvauclair Think we're good now

cvauclair (approved):

PR #355, opened 2025-03-17
fix(openai): serde rename for image_url UserContent

### LINKED ISSUE #349: bug:  The type of image uploaded by OpenAI is image_url instead of image

## Bug Report

OpenAI's image type has been changed to a image_url type, which causes an error to always appear when I want to upload an image through the frame, saying that the image type is not supported

Image

However, I found out that image is used as a type in openai's UserContent struct.

Image

## Additional context

All you need to do is add a line of code.

Image

joshua-mo-143:
Fixes #349 

I am wondering if we should change the enum variant to just say `ImageUrl` instead of `Image` as that more accurately describes the variant.

### DIFF
--- a/rig-core/src/providers/openai.rs
+++ b/rig-core/src/providers/openai.rs
@@ -1164,7 +1164,7 @@ mod tests {
                     "text": "What's in this image?"
                 },
                 {
-                    "type": "image",
+                    "type": "image_url",
                     "image_url": {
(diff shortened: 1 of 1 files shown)


0xMochan:
I think it was called `Image` as it matched closer to the generalized version of the model. i could see it either way

0xMochan (approved):

joshua-mo-143:
Ok let's just keep it as is then. It's neither here nor there and would be a breaking change

PR #357, opened 2025-03-18
feat: Add image generation to all providers that support it

yavens:
This PR adds image generation to OpenAI, Azure, Huggingface, and Hyperbolic. 

Locks image generation behind feature flag as per #236.


rig-core/examples/openai_image_generation.rs
```diff
@@ -0,0 +1,27 @@
...
+        .image_generation_request()
+        .prompt("A castle sitting upon a large mountain, overlooking the water.")
+        .size((1024, 1024))
+        .send()
+        .await
+        .expect("Failed to generate image");
+
+    let arguments: Vec<String> = args().collect();
+
+    let path = Path::new(&arguments[1]);
```
joshua-mo-143, line 24: This should be done at the start of the example - we don't want the user to use up tokens generating an image only to find out after that they need to pass a filename in

rig-core/examples/hyperbolic_image_generation.rs
```diff
@@ -0,0 +1,27 @@
...
+        .image_generation_request()
+        .prompt("A castle sitting upon a large mountain, overlooking the water.")
+        .size((1024, 1024))
+        .send()
+        .await
+        .expect("Failed to generate image");
+
+    let arguments: Vec<String> = args().collect();
+
+    let path = Path::new(&arguments[1]);
```
joshua-mo-143, line 24: Same issue as openAI example wrt using up tokens generating the image before trying to find out if the user has passed in a filename or not. We should probably have a default here.

rig-core/examples/huggingface_image_generation.rs
```diff
@@ -0,0 +1,27 @@
...
+        .image_generation_request()
+        .prompt("A castle sitting upon a large mountain, overlooking the water.")
+        .size((1024, 1024))
+        .send()
+        .await
+        .expect("Failed to generate image");
+
+    let arguments: Vec<String> = args().collect();
+
+    let path = Path::new(&arguments[1]);
```
joshua-mo-143, line 24: Same issue as openAI example wrt using up tokens generating the image before trying to find out if the user has passed in a filename or not. We should probably have a default here.

rig-core/src/image_generation.rs
```diff
@@ -0,0 +1,115 @@
...
+}
+pub trait ImageGeneration<M: ImageGenerationModel> {
+    /// Generates a transcription request builder for the given `file`.
+    /// This function is meant to be called by the user to further customize the
+    /// request at transcription time before sending it.
+    ///
+    /// ❗IMPORTANT: The type that implements this trait might have already
+    /// populated fields in the builder (the exact fields depend on the type).
+    /// For fields that have already been set by the model, calling the corresponding
+    /// method on the builder will overwrite the value set by the model.
```
joshua-mo-143, line 34: Is there a way to prevent this from happening or is this due to the way we currently set up completion requests?

rig-core/src/providers/huggingface/client.rs
```diff
@@ -38,11 +41,24 @@ impl SubProvider {
...
         }
     }
 
+    /// Get the image generation endpoint for the SubProvider
+    /// Required because Huggingface Inference requires the model
+    /// in the url and in the request body.
+    pub fn image_generation_endpoint(&self, model: &str) -> String {
+        match self {
+            SubProvider::HFInference => format!("/{}", model),
+            _ => panic!(
```
joshua-mo-143, line 55: I am not sure we should panic here, we should return an error

rig-core/src/providers/huggingface/client.rs
```diff
@@ -164,6 +182,7 @@ impl Client {
 
     pub(crate) fn post(&self, path: &str) -> reqwest::RequestBuilder {
         let url = format!("{}/{}", self.base_url, path).replace("//", "/");
+        println!("{}", url);
```
joshua-mo-143, line 185: I'm wondering if we should use `tracing::info!` here for telemetry instead of `println!` as it encourages good practices and devs who are using Rig in prod already likely have a tracing setup created (and we can also redirect developers to resources on how they can improve their own telemetry)

rig-core/src/image_generation.rs
```diff
@@ -0,0 +1,115 @@
...
+    > + Send;
+
+    fn image_generation_request(&self) -> ImageGenerationRequestBuilder<Self> {
+        ImageGenerationRequestBuilder::new(self.clone())
+    }
+}
+
+pub struct ImageGenerationRequest {
+    pub prompt: String,
+    pub size: (u32, u32),
```
joshua-mo-143, line 66: We should probably split this into width and height, given that most of the APIs that support image generation also have their own input parameters for width and height. Having the two `u32`s in a tuple like this may also be confusing later down the line and we should ensure that there is no ambiguity.

joshua-mo-143 (changes requested):
Looks great! I have left some comments, but overall the implementation is mostly fine.

rig-core/src/image_generation.rs
yavens: Yeah, this is set up the same as completion requests.

rig-core/src/providers/huggingface/client.rs
yavens: Oops, I totatlly meant to remove that 😅

rig-core/examples/openai_image_generation.rs
yavens: Right, didn't think about that.

rig-core/src/providers/huggingface/client.rs
yavens: Fair enough, I swapped it to a Result

rig-core/src/image_generation.rs
yavens: 👍 Fixed

rig-core/examples/huggingface_image_generation.rs
yavens: Moved + added simple default

rig-core/examples/hyperbolic_image_generation.rs
yavens: Moved + added simple default

joshua-mo-143:
lgtm

PR #363, opened 2025-03-26
fix: no params tools definition for Gemini

piotrostr:


0xMochan:
Interesting, I didn't realize this was nested as so. I'm surprised since the gemini example works fine with tools, i presume that's using a tool definition.

It still works with this PR so I'm surprised that this is necessary, maybe I'm missing something

piotrostr:
its only when no params are passed that this occurs! :P

PR #364, opened 2025-03-26
Fix Clippy warnings for doc indentation and Error::other usage

strmfos:
**Summary**
- Cleaned up some doc comments that Clippy flagged for over-indentation.
- Switched from `std::io::Error::new(ErrorKind::Other, ...)` to `Error::other(...)` in a couple of places to satisfy Clippy’s rules.

**Details**
- In `rig-qdrant` and `rig-neo4j`, I reduced extra whitespace in a few doc comments so they line up properly.
- In `rig-core`, changed `Error::new(...)` calls to `Error::other(...)` for a tidier approach.

Everything still builds fine under `cargo clippy --all -- -D warnings`, with no new warnings.

### DIFF
--- a/rig-core/src/streaming.rs
+++ b/rig-core/src/streaming.rs
@@ -102,7 +102,7 @@ pub async fn stream_to_stdout<M: StreamingCompletionModel>(
                     .tools
                     .call(&name, params.to_string())
                     .await
-                    .map_err(|e| std::io::Error::new(std::io::ErrorKind::Other, e.to_string()))?;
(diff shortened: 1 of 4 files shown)


0xMochan:
I appreciate the commit. It's a bit odd that you found some clippy errors that our CI/CD didn't catch since it shouldn't be possible to merge clippy non-compliant code into the branch. Maybe there might be a difference in clippy version there, but appreciate the fixes here!

Issue #346, opened 2025-03-11
feat: replace pipelines with `tower` style middleware for Rig


joshua-mo-143:
- [x] I have looked for existing issues (including closed) about this

## Feature Request

Agentic chains (which we refer to as "pipelines") in Rig as it stands at the moment are currently a bit limited and arguably difficult to extend.

By leveraging recursively stacking types, we can create a pipeline system that is much more extendable and less reliant on macros(!) than our current pipeline system. Once we have reached feature parity with this and the feature has been rolled out, we can then deprecate the old pipeline module.

### Motivation

Better DX, better scalability, better everything. We might need to ensure we implement a macro at some point for better debugging though, if users start trying to implement their own layers.

### Proposal

Use the `tower` crate and create several layers/services that will effectively have parity with the current pipeline system.

### Alternatives

Not really. We could extend the current pipeline system, but it doesn't seem very easy to do so.

joshua-mo-143:
Closing because at some point we want to implement streaming pipelines, which `tower` is bad at.

PR #365, opened 2025-03-28
feat: impl base Tower services/layers 

joshua-mo-143:
One step in a journey towards getting #346 done.

~~I'm keeping this PR relatively small as it's just to ensure that we get _something_ done while figuring out the other stuff.~~ See files changed - we're basically implementing all the services/layers that one would need to not require the Pipeline API anymore.

## Breaking changes from this PR
- `CompletionRequest` to become `Clone`


joshua-mo-143:
Probably awaiting #370 before merge as it is breaking and is extremely likely to break this PR.

rig-core/src/middlewares/tools.rs
```diff
@@ -0,0 +1,86 @@
...
+    tools: Arc<ToolSet>,
+}
+
+impl<S, T> Service<CompletionRequest> for ToolLayerService<S>
+where
+    S: Service<CompletionRequest, Response = CompletionResponse<T>> + Clone + Send + 'static,
+    T: Send + 'static,
+    S::Future: Send,
+{
+    type Response = (Vec<Message>, String, ToolResultContent);
```
cvauclair, line 47: Should this not be `CompletionResponse`?

rig-core/src/middlewares/rag.rs
cvauclair, line 1: IMO we should have a general "RAG" layer/service that only requires access to a function `Fn(String) -> Vec<Document>` (or equivalent). Thus allowing non-vector store based RAGging (e.g.: web search, fulltext-search, etc.).

rig-core/src/middlewares/components.rs
```diff
@@ -0,0 +1,252 @@
...
+    D: Debug,
+{
+    pub fn new(inner: S, integration: T, predicate: R, _t: PhantomData<D>) -> Self {
+        Self {
+            inner,
+            integration,
+            predicate,
+            _t,
+        }
+    }
```
cvauclair, line 155: Why do we need to pass `_t: PhantomData<D>` in the constructor? Can we not just initialize it when constructing the struct?

rig-core/src/middlewares/components.rs
cvauclair, line 1: The code is a bit dense (took me a while to figure out what was happening), but this is pretty :fire:

cvauclair (commented):
Nicely done :muscle:  This is a good starting point for the tower middleware refactor. I've added a couple comments (feel free to address now or in a subsequent PR, not a big deal since this is not going to main).

FYI even though this new API is designed to supersede the pipeline API in most cases, it doesn't have to follow the exact same architecture (speaking in terms of ops <-> layers/services)

rig-core/src/middlewares/rag.rs
joshua-mo-143: Ended up throwing out the old RAG layer and implementing a new RAG layer that uses a trait on the inner generic (that is basically `Fn(String) -> Vec<T: Deserialize + Serialize>`) so that you can pretty much use whatever RAG method you want.

PR #385, opened 2025-04-09
feat: impl From<mcp_core::types::Tool> for ToolDefinition

### LINKED ISSUE #383: feat(core): impl From<mcp_core::types::Tool> for ToolDefinition

## Feature Request

Self-explanatory.

### Motivation

You can technically just do it yourself by manually instantiating `ToolDefinition`, but this is kinda clunky and shouldn't really be required.

### Proposal

Just impl From<T>. The field names are essentially the same, so there isn't much work required here.

joshua-mo-143:
Self explanatory. Fixes #383 

Potential queries:
- The description is actually an Option<String> in `mcp-core`'s version of description. I have set it to produce an empty String if the unwrap fails. Do we want to handle this differently?

### DIFF
--- a/rig-core/src/tool.rs
+++ b/rig-core/src/tool.rs
@@ -205,6 +205,17 @@ where
 }
 
 #[cfg(feature = "mcp")]
+impl From<mcp_core::types::Tool> for ToolDefinition {
+    fn from(val: mcp_core::types::Tool) -> Self {
+        Self {
+            name: val.name,
+            description: val.description.unwrap_or_default(),
+            parameters: val.input_schema,
+        }
+    }
+}
+
+#[cfg(feature = "mcp")]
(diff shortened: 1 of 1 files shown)


rig-core/src/tool.rs
```diff
@@ -204,6 +204,17 @@ where
     }
 }
 
+#[cfg(feature = "mcp")]
+impl From<mcp_core::types::Tool> for ToolDefinition {
```
cvauclair, line 208: Should we also want an `impl From<&mcp_core::types::Tool>` so that we don't necessarily consume the tool?

joshua-mo-143: done

cvauclair (approved):

PR #386, opened 2025-04-09
fix: gemini embeddings does not work for multiple documents

### LINKED ISSUE #376: bug: Gemini Embedding Not Working.
Example for Gemini Edbedding Not Working.  

Panic for "Document should be present" 

link: [https://github.com/0xPlaygrounds/rig/blob/main/rig-core/examples/gemini_embeddings.rs](url)

joshua-mo-143:
Fixes #376 

Has been tested locally and I've added a docstring to link back to the API ref.

EDIT: This solution is quite hacky. We can kick the can down the road, but we're probably gonna have to fix  this at some point.


0xMochan (approved):
Probably can be improved in the future but yea gemini's api here is odd at best

PR #387, opened 2025-04-09
refactor(lancedb): docs, make examples idempotent

### LINKED ISSUE #107: docs(rig-lancedb): Improve instructions and example
The `rig-lancedb` documentation could be improved with the following:
- [ ] Document that `lancedb` requires the user to have installed `protoc` on their machine (the `lancedb` library will not compile otherwise) and add instructions on how to install it for a couple of popular systems.
- [ ] In the examples, when the table is created, add a check to see if the table already exists. The current examples will fail if run twice since the table creation step will error with a "table already exists" error.

### LINKED ISSUE #211: Update contributors guide with dependencies
When I run `cargo test` for first time I found an error running `protoc`.
I learn that is needed to run LanceDB module, however is not in the contributors guide README, it would be nice to add this type of documentation (actually would be nice to not depend on installing external libraries).

joshua-mo-143:
Fixes #107, fixes #211

### DIFF
--- a/rig-lancedb/README.md
+++ b/rig-lancedb/README.md
@@ -18,7 +18,7 @@
 This companion crate implements a Rig vector store based on Lancedb.
 
 ## Pre-requisites
-If you are using `rig-lancedb` locally, you must ensure you have `protoc` (the [Protobuf Compiler](https://protobuf.dev/installation/)).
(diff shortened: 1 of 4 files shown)


cvauclair (approved):
Good stuff!

PR #379, opened 2025-04-03
feat: OpenRouter streaming with Parallel Tool Calls

piotrostr:
- **pass on IDs as well**
- **fix serde**

PR #388, opened 2025-04-09
refactor: Improve Streaming API

yavens:
This PR improves streaming by providing access to a completed `Message` and important data such as usage.


rig-core/src/streaming.rs
```diff
@@ -41,29 +57,98 @@ impl Display for StreamingChoice {
...
-pub type StreamingResult = Pin<Box<dyn Stream<Item = Result<StreamingChoice, CompletionError>>>>;
+pub type StreamingResult<R> =
+    Pin<Box<dyn Stream<Item = Result<RawStreamingChoice<R>, CompletionError>>>>;
+
+pub struct StreamingCompletionResponse<R: Clone + Unpin> {
+    inner: StreamingResult<R>,
+    text: String,
+    tool_calls: Vec<(String, String, serde_json::Value)>,
+    pub message: Message,
+    pub response: Option<R>,
```
cvauclair, line 72: Docstrings please!

rig-core/src/streaming.rs
```diff
@@ -41,29 +57,98 @@ impl Display for StreamingChoice {
...
+                    content.insert(0, AssistantContent::text(stream.text.clone()));
+                }
+
+                stream.message = Message::Assistant {
+                    content: OneOrMany::many(content)
+                        .expect("There should be at least one assistant message"),
+                };
+
+                Poll::Ready(None)
+            }
```
cvauclair, line 112: I would add a comment explaining that this logic is executed at the end of the stream and basically collects everything in the `StreamingCompletionResponse<R>::message` field

cvauclair: I'd also document this behavior on the `StreamingCompletionResponse<R>` type itself

rig-core/src/streaming.rs
```diff
@@ -13,14 +13,30 @@ use crate::agent::Agent;
...
+
+    /// The final response object
+    FinalResponse(R),
+}
+
+/// Enum representing a streaming chunk from the model
+#[derive(Debug, Clone)]
 pub enum StreamingChoice {
     /// A text chunk from a message response
     Message(String),
```
cvauclair, line 42: I believe `StreamingChoice` can be replaced by `completion::message::AssistantContent` here without losing anything. This should be done to unify both APIs

cvauclair (changes requested):
Good work! Not a full review (will leave that to @0xMochan or @joshua-mo-143), but spotted a couple things. Cheers!

rig-core/src/streaming.rs
TheNuclearNexus: I don't see why we couldn't, would mean a major breaking change though

yavens: I don't see why we couldn't, would mean a major breaking change though

rig-core/src/streaming.rs
yavens: 👍

rig-core/src/streaming.rs
yavens: Right, makes sense

rig-core/src/streaming.rs
cvauclair: It's already a breaking PR though, and IMO unifying both APIs would allow for a lot of code reuse across both streaming and non-streaming contexts

joshua-mo-143: We would need to make sure we notify people ahead of time about this and potentially provide a migration path, but I'm fine with it if it means we can unify the APIs.

yavens: I went ahead and made this change for now, its like 2 lines on our end so rolling it back isn't an issue

0xMochan: It might break with @piotrostr stuff

rig-core/src/streaming.rs
```diff
@@ -13,57 +13,136 @@ use crate::agent::Agent;
...
+pub type StreamingResult<R> =
+    Pin<Box<dyn Stream<Item = Result<RawStreamingChoice<R>, CompletionError>>>>;
+
+/// The response from a streaming completion request;
+/// message and response are populated at the end of the
+/// `inner` stream.
+pub struct StreamingCompletionResponse<R: Clone + Unpin> {
+    inner: StreamingResult<R>,
+    text: String,
+    tool_calls: Vec<(String, String, serde_json::Value)>,
```
0xMochan, line 52: I feel like this `ToolCall` tuple here is often used, it might be nice to have a type alias for it? i've seen it in multiple places

0xMochan (approved):
This looks pretty good. I think before this gets merged. it would be good to merge in the work from #379 .

Also, an example for tool call looping would be good, particularly after work from #370 is merged

rig-core/src/streaming.rs
yavens: Realized I could just use the ToolCall struct directly here instead

rig-core/src/providers/openrouter/completion.rs
```diff
@@ -292,16 +156,30 @@ impl completion::CompletionModel for CompletionModel {
             "temperature": completion_request.temperature,
```
kosz78, line 156: To support tools invocation this should be tools declaration too like `"tools": completion_request.tools,`

0xMochan:
I pushed some code related to actually doing some tool calling in the `openai_streaming_with_tools.rs`. I want to think a bit more deeply to make it slightly easier for users to be able to do full tool calling applications while providing the streaming feature set.

It bothers me a bit that `CompletionModel::completion` and `StreamingCompletionModel::stream_completion` work completely differently in feature set (the `streaming` version doesn't even have any built-in RAG). I think a lot of this gets resolved via the #346 PR that we keep mentioning but I think there's deeper thinking to do.

In order to avoid this PR being bike-shed to death, I'm thinking of just promoting the `tool_call_helper` method from `openai_streaming_with_tools.rs` somewhere and approaching this in a follow-up PR that makes streaming and streaming-based reasoning loops more apt!

rig-core/src/providers/openai/streaming.rs
```diff
@@ -121,64 +138,85 @@ pub async fn send_compatible_streaming_request(
...
         }
 
-        for (_, (name, arguments)) in calls {
+        for (_, (id, name, arguments)) in calls {
             let Ok(arguments) = serde_json::from_str(&arguments) else {
                 continue;
             };
 
-            yield Ok(streaming::StreamingChoice::ToolCall(name, "".to_string(), arguments))
+            println!("{id} {name}");
```
kosz78, line 211: Please use `tracing` instead of  `println`

0xMochan: What's the reasoning for this suggestion? this is a `yield` statement being converted into a `println!` call?

yavens: This was debug, I forgot to remove it 😅

0xMochan (approved):

PR #392, opened 2025-04-11
chore(fastembed): update fastembed version

### LINKED ISSUE #390: bug: compiler error for rig-fastembed

## Bug Report
error[E0004]: non-exhaustive patterns: `&fastembed::EmbeddingModel::ModernBertEmbedLarge` not covered                                                                            
   --> C:\Users\longz\.cargo\registry\src\mirrors.tuna.tsinghua.edu.cn-e791a3f93f26854f\rig-fastembed-0.1.4\src\lib.rs:147:11
    |
147 |     match model {
    |           ^^^^^ pattern `&fastembed::EmbeddingModel::ModernBertEmbedLarge` not covered
    |
note: `fastembed::EmbeddingModel` defined here
   --> C:\Users\longz\.cargo\registry\src\mirrors.tuna.tsinghua.edu.cn-e791a3f93f26854f\fastembed-4.6.0\src\models\text_embedding.rs:9:1
    |
9   | pub enum EmbeddingModel {
    | ^^^^^^^^^^^^^^^^^^^^^^^
...
45  |     ModernBertEmbedLarge,
    |     -------------------- not covered
    = note: the matched value is of type `&fastembed::EmbeddingModel`
help: ensure that all possible cases are being handled by adding a match arm with a wildcard pattern or an explicit pattern as shown
    |
174 ~         | FastembedModel::GTELargeENV15Q => 1024,
175 ~         &fastembed::EmbeddingModel::ModernBertEmbedLarge => todo!(),

fastembed::EmbeddingModel::ModernBertEmbedLarge  was not applied to the match block.

joshua-mo-143:
Fixes #390. The source of the issue was a breaking change.

### DIFF
--- a/rig-fastembed/Cargo.toml
+++ b/rig-fastembed/Cargo.toml
@@ -13,7 +13,7 @@ serde = { version = "1.0.193", features = ["derive"] }
 serde_json = "1.0.108"
 tracing = "0.1.40"
 schemars = "0.8.16"
-fastembed = "4.4.0"
+fastembed = "4.6.0"
 
 [dev-dependencies]
 anyhow = "1.0.75"
--- a/rig-fastembed/src/lib.rs
+++ b/rig-fastembed/src/lib.rs
@@ -171,6 +171,7 @@ pub fn fetch_model_ndims(model: &FastembedModel) -> usize {
         | FastembedModel::MxbaiEmbedLargeV1
         | FastembedModel::MxbaiEmbedLargeV1Q
         | FastembedModel::GTELargeENV15
+        | FastembedModel::ModernBertEmbedLarge
         | FastembedModel::GTELargeENV15Q => 1024,
     }
 }


joshua-mo-143:
Potential clarifying Qs before we review this:
- Do we want to create a version of the `EmbeddingsBuilder` for images? It currently supports Strings at the moment and that's it.
- Do we want to extend `ImageEmbeddingModel` to additionally support embedding images straight from files/URLs? This seems like something that would be a relatively easy win and improve DX significantly.

joshua-mo-143:
De-scoping this PR for now, I think there is enough useful work here that we can meaningfully merge it and the image embeddings builder is likely to be a larger piece of work that will significantly delay this PR being able to get merged. Not particularly helpful as this PR was originally just meant to be a bug fix.

cvauclair:
> Do we want to create a version of the EmbeddingsBuilder for images? It currently supports Strings at the moment and that's it.

We could create a different version with a similar API, or use a typestate pattern for it. E.g.:
```rust
trait EmbeddingModel {
    type Document;
}

// text embedding models would have `Document = String` while image embedding models would have `Document = Vec<u8>` or something like that

struct EmbeddingBuilder<M> {
    pub fn document(mut self, document: M::Document) -> Self {...}
}
```
Not sure how feasible this is with the current `EmbeddingBuilder<T: Embed>` trait bound but something to think about.

> Do we want to extend ImageEmbeddingModel to additionally support embedding images straight from files/URLs? This seems like something that would be a relatively easy win and improve DX significantly.

Would that just be a wrapper that loads the file in-memory? Or does the model provider actually accept a URL?

joshua-mo-143:
> > Do we want to extend ImageEmbeddingModel to additionally support embedding images straight from files/URLs? This seems like something that would be a relatively easy win and improve DX significantly.
> 
> Would that just be a wrapper that loads the file in-memory? Or does the model provider actually accept a URL?

That would depend. The tl;dr is that the `rig-fastembed` library currently accepts straight up `Vec<u8>` (or `&[u8]` if we're being pendantic) but I imagine when it comes to managed models, we will need to do some further exploration. We currently don't actually support any other providers that support image embeddings directly.

joshua-mo-143:
Superceded by #480 (although we can re-use this PR for implementing image embeddings)

joshua-mo-143:
Superceded by #480 (although we can re-use this PR for implementing image embeddings)

PR #394, opened 2025-04-11
feat: Inception provider

### LINKED ISSUE #393: feat: Add support for Inception Labs' Mercury Coder Beta dLLM
## Model Provider Integration Request

Inception Labs has started rolling out beta access to its `mercury-coder-small` dLLM. It would be awesome to have support for it in Rig.

### Resources

API docs are at least account-gated, they might be gated by beta access as well. I have a beta key so I can see them, not sure if anyone else can:

https://platform.inceptionlabs.ai/docs

I'd be happy to work on this.

cbrit:
Closes #393


yavens:
It looks like Inception is an OpenAI-compatible provider. Is there a reason you can't reuse the types directly from the OAI module? See [Hyperbolic](https://github.com/0xPlaygrounds/rig/blob/main/rig-core/src/providers/hyperbolic.rs#L357-L405)'s intergration

cbrit:
> It looks like Inception is an OpenAI-compatible provider. Is there a reason you can't reuse the types directly from the OAI module? See [Hyperbolic](https://github.com/0xPlaygrounds/rig/blob/main/rig-core/src/providers/hyperbolic.rs#L357-L405)'s intergration

Oh great catch. That would simplify things for sure.

cbrit:
So since the beta only supports text, it seems they don't yet accept structured content data in the `content` field of the messages:

```bash
// this works
{ ... "messages": [{ "content": "Hello, world!" }] }

// this is what Open AI expects. Does not work with Inception API right now
{ ... "messages": [{ "content": { "type": "text", "text": "Hello, world!" }}]}
```

So I was able to simplify it by using the OpenAI provider's streaming utility, but not by using its `Message` type.

0xMochan:
> So since the beta only supports text, it seems they don't yet accept structured content data in the `content` field of the messages:
> 
> ```bash
> // this works
> { ... "messages": [{ "content": "Hello, world!" }] }
> 
> // this is what OpenAI expects/OpenAI Message type serializes to. Does not work with Inception API right now
> { ... "messages": [{ "content": { "type": "text", "text": "Hello, world!" }}]}
> ```
> 
> So I was able to simplify it by using the OpenAI provider's streaming utility, but not by using its `Message` type.
> 
> The resulting runtime error if using the other `Message` type:
> 
> ```
> ProviderError:  "detail":[{"type":"string_type","loc":["body","messages",0,"content"],"msg":"Input should be a valid string","input":[{"type":"text","text":"You are a helpful AI assistant."}]},{"type":"string_type","loc":["body","messages",1,"content"],"msg":"Input should be a valid string","input":[{"type":"text","text":"Hello, how are you?"}]}]}
> ```

It seems like a lot of "openai compatible" apis are like this which makes me think we might want a custom serializer here..

0xMochan:
> @0xMochan Reverted to the custom message implementation

Is this bc they support the full content type now? otherwise, we can flag this for review?

cbrit:
> > @0xMochan Reverted to the custom message implementation
> 
> Is this bc they support the full content type now? otherwise, we can flag this for review?

We discussed in discord dms about reverting because they don't. I haven't actually checked to see if the API has changed though now that it's been a few weeks.

joshua-mo-143:
@cbrit Don't suppose anything has changed on this? I'm looking into potentially adding more new providers.

joshua-mo-143:
Closing as this is currently not in scope (and the impl is severely outdated).

If you still want to contribute this provider we're more than happy to accept! However you will probably want to open a new PR as the way we create new providers is quite different from the original implementation used here.

PR #396, opened 2025-04-12
feat: trait for embedding images

### LINKED ISSUE #397: feat: trait for image embeddings

## Feature Request

In order to implement #395, we need to first add a trait to `rig-core` itself so that we can implement image embeddings.

### Motivation

From ChatGPT:
Image embeddings are a powerful tool in modern AI systems because they enable complex visual data to be represented as compact, semantically meaningful vectors. These embeddings capture high-level features of an image—such as objects, textures, and spatial relationships—while reducing the dimensionality of raw pixel data. This allows for efficient storage, retrieval, and computation, which is essential for large-scale image processing tasks. By mapping similar images to nearby points in vector space, image embeddings make it easy to perform tasks like image classification, clustering, search, and similarity detection using simple mathematical operations.

For AI developers, embeddings serve as a versatile interface between raw visual content and downstream tasks. Pretrained models like CLIP, ResNet, or Vision Transformers can generate embeddings that generalize well across domains, reducing the need for extensive labeled datasets. These embeddings can be indexed for fast nearest-neighbor search in retrieval-based systems or combined with text embeddings for cross-modal applications like image-captioning, visual question answering, and multimodal agents. In essence, image embeddings provide a language for machines to "understand" and reason about visual content in a way that is both scalable and composable.

### Proposal

See linked PR.

joshua-mo-143:
Fixes #397

### DIFF
--- a/rig-core/src/embeddings/embedding.rs
+++ b/rig-core/src/embeddings/embedding.rs
@@ -60,21 +60,21 @@ pub trait EmbeddingModel: Clone + Sync + Send {
     }
 }
 
-/// Trait for embedding models that can generate embeddings for documents.
+/// Trait for embedding models that can generate embeddings for images.
 pub trait ImageEmbeddingModel: Clone + Sync + Send {
-    /// The maximum number of documents that can be embedded in a single request.
+    /// The maximum number of images that can be embedded in a single request.
     const MAX_DOCUMENTS: usize;
 
     /// The number of dimensions in the embedding vector.
     fn ndims(&self) -> usize;
 
-    /// Embed multiple text documents in a single request
+    /// Embed multiple images in a single request from bytes.
     fn embed_images(
         &self,
         images: impl IntoIterator<Item = Vec<u8>> + Send,
     ) -> impl std::future::Future<Output = Result<Vec<Embedding>, EmbeddingError>> + Send;
 
-    /// Embed a single text document.
+    /// Embed a single image from bytes.
     fn embed_image<'a>(
         &'a self,
         bytes: &'a [u8],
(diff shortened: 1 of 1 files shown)


rig-core/src/embeddings/embedding.rs
```diff
@@ -60,6 +62,39 @@ pub trait EmbeddingModel: Clone + Sync + Send {
...
+
+    /// The number of dimensions in the embedding vector.
+    fn ndims(&self) -> usize;
+
+    /// Embed multiple text documents in a single request
+    fn embed_texts<'a>(
+        &'a self,
+        images: impl IntoIterator<Item = &'a [u8]> + Send,
+    ) -> Pin<
+        Box<dyn std::future::Future<Output = Result<Vec<Embedding>, EmbeddingError>> + Send + 'a>,
```
0xMochan, line 78: for a trait, we should return `impl Future` or `impl IntoFuture` as it's more flexible.

implementations of the trait can use a specific `BoxFuture`. *I also think we should always write `BoxFuture` instead of expanded form as it's easier to update when we get `impl` in assoc types, etc.*

rig-core/src/embeddings/embedding.rs
```diff
@@ -60,6 +62,39 @@ pub trait EmbeddingModel: Clone + Sync + Send {
...
+    >;
+
+    /// Embed a single text document.
+    fn embed_text<'a>(
+        &'a self,
+        bytes: &'a [u8],
+    ) -> Pin<Box<dyn std::future::Future<Output = Result<Embedding, EmbeddingError>> + Send + 'a>>
```
0xMochan, line 85: Ditto

0xMochan (changes requested):

0xMochan:
I forgot to leave a msg. Overall looks clean other than using `impl Trait`s for return position for future flexibility

rig-core/src/embeddings/embedding.rs
joshua-mo-143: Amended for now. I wanted to try and use &[u8] as the iterator item but it looks like that totally breaks the trait.

0xMochan (approved):
Looks good!

PR #398, opened 2025-04-13
feat: semantic routing implementation

### LINKED ISSUE #389: feat: Semantic routing implementation

## Feature Request

We've talked about this briefly internally a while ago, but a semantic routing implementation for users who want that would not go amiss.

[A rough description of what semantic routing is with the semantic-router Python repo](https://github.com/aurelio-labs/semantic-router)

### Motivation

Makes it easier to deny sending a response back if the topic doesn't fit.

### Proposal

It would probably look something like this (not exact as this is mainly pseudocode):

```rust
struct SemanticRouter<T> {
    router: T
}

pub trait SemanticRouting {
    async fn find_route(&self, prompt: &str) -> Option<String>;
    async fn insert_phrase(&self, phrase: &str, category: &str);
}

impl SemanticRouter for InMemoryVectorStore {
    // .. so on and so forth
}
```

### Alternatives

Just not implementing it

joshua-mo-143:
Fixes #389

### DIFF
--- a/rig-core/src/semantic_routing.rs
+++ b/rig-core/src/semantic_routing.rs
@@ -1,0 +1,77 @@
+use std::collections::HashMap;
+
+use serde::{Deserialize, Serialize};
+
+use crate::{embeddings::EmbeddingModel, vector_store::VectorStoreIndex};
+
+pub struct SemanticRouter<V> {
+    store: V,
+    threshold: f64,
+}
+
+impl<V> SemanticRouter<V> {
+    pub fn builder() -> SemanticRouterBuilder<V> {
+        SemanticRouterBuilder::new()
+    }
+}
+
+impl<V> SemanticRouter<V>
+where
+    V: VectorStoreIndex,
+{
+    pub async fn select_route(&self, query: &str) -> Option<String> {
+        let res = self.store.top_n(query, 1).await.ok()?;
+        let (score, _, SemanticRoute { tag }) = res.first()?;
+
+        if *score < self.threshold {
+            return None;
+        }
+
+        Some(tag.to_owned())
+    }
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct SemanticRoute {
+    tag: String,
+}
+
+pub trait Router: VectorStoreIndex {
+    fn retrieve_route() -> impl std::future::Future<Output = Option<String>> + Send;
+}
+
+pub struct SemanticRouterBuilder<V> {
+    store: Option<V>,
+    threshold: Option<f64>,
+}
+
+impl<V> SemanticRouterBuilder<V> {
+    pub fn new() -> Self {
+        Self {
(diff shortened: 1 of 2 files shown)


rig-core/src/semantic_routing.rs
```diff
@@ -0,0 +1,89 @@
...
+        if *score < self.threshold {
+            return None;
+        }
+
+        Some(tag.to_owned())
+    }
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct SemanticRoute {
```
0xMochan, line 35: I want to normalize adding some more default derives, atleast `Debug` (i've ran into situations in debugging where something was missing `Debug`.)

Atleast for new structs, we can go back and fill in some derives in a later PR.

rig-core/src/semantic_routing.rs
```diff
@@ -0,0 +1,89 @@
...
+    tag: String,
+}
+
+pub trait Router: VectorStoreIndex {
+    fn retrieve_route() -> impl std::future::Future<Output = Option<String>> + Send;
+}
+
+pub struct SemanticRouterBuilder<V> {
```
0xMochan, line 43:
```suggestion
#[derive(Default)]
pub struct SemanticRouterBuilder<V> {
```

Should be easier to use the `derive` here instead of an `impl` unless you want to reserve relying on `::new` usage as it could change in the future.

rig-core/src/semantic_routing.rs
```diff
@@ -0,0 +1,89 @@
...
+
+        self
+    }
+
+    pub fn build(self) -> Result<SemanticRouter<V>, SemanticRouterError> {
+        let Some(store) = self.store else {
+            return Err(SemanticRouterError::StoreNotFound);
+        };
+
+        let threshold = self.threshold.unwrap_or(0.9);
```
0xMochan, line 79: If the default is `0.9`, I would place that directly into the struct via `#[default = 0.9]` or make it a constant.
Mostly tho, is there a significance for `0.9` here? using a `const` allows you to place a docstring explaining it's origin which is nice.

0xMochan (changes requested):
Looks straightforward and clean. minor changes but otherwise we can send it

rig-core/src/semantic_routing.rs
joshua-mo-143: Not at the moment - it was mostly just so it doesn't automatically break if the user adds their own value. I might loosen the value though if we're keeping it in as 0.9 is a bit high

joshua-mo-143:
Closing because I'm moving this impl to a new crate which will allow for much more rapid experimentation and I think we can do much better than a tiny wrapper over a vector index.

Once it's been fully implemented and we can see that it's being used, we should then graduate it to `rig-core` if that is still what we want.

PR #403, opened 2025-04-16
Support custom clients for bedrock

marcbowes:
The current ClientBuilder is limited. Instead of adding more options, we
now provide a way to externally configure and inject a client.

### DIFF
--- a/rig-bedrock/Cargo.toml
+++ b/rig-bedrock/Cargo.toml
@@ -7,7 +7,7 @@ readme = "README.md"
 description = "AWS Bedrock model provider for Rig integration."
 
 [dependencies]
-rig-core = { version = "0.11.0", features = ["image"]  }
+rig-core = { version = "0.11.1", path = "../rig-core", features = ["image"]  }
 rig-derive = { path = "../rig-core/rig-core-derive", version = "0.1.1" }
 serde = { version = "1.0.193", features = ["derive"] }
 serde_json = "1.0.108"
--- a/rig-bedrock/src/client.rs
+++ b/rig-bedrock/src/client.rs
@@ -52,6 +52,12 @@ pub struct Client {
     pub(crate) aws_client: aws_sdk_bedrockruntime::Client,
 }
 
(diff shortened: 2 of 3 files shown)


rig-bedrock/Cargo.toml
```diff
@@ -7,7 +7,7 @@ readme = "README.md"
 description = "AWS Bedrock model provider for Rig integration."
 
 [dependencies]
-rig-core = { version = "0.11.0", features = ["image"]  }
+rig-core = { version = "0.11.1", path = "../rig-core", features = ["image"]  }
```
marcbowes, line 10: I'm not sure if this is the right way to do it, but I spot checked some other crates and this is what the pattern seems to be.

0xMochan:
Hey, we appreciate your catch on this feature. I think it's important people can use their own clients, though, it might get tricky if you move it into the rig based client unless we can figure a way to also borrow it or use an `Arc` if necessary.

rig-bedrock/Cargo.toml
0xMochan: I'm fairly certain this is correct as well!

0xMochan (approved):
Looks good though, I want to think a bit about the situation where you want to share a client. It might not be that common to share clients since they likely don't carry state so cloning them might be sufficient in most cases.

rig-bedrock/src/client.rs
```diff
@@ -52,6 +52,12 @@ pub struct Client {
     pub(crate) aws_client: aws_sdk_bedrockruntime::Client,
 }
 
+impl From<aws_sdk_bedrockruntime::Client> for Client {
+    fn from(aws_client: aws_sdk_bedrockruntime::Client) -> Self {
+        Client { aws_client }
+    }
+}
+
```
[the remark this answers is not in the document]
0xMochan: Ah, `Client` is a wrapped of an `Arc`, makes sense!

0xMochan (approved):
I think this is great!

marcbowes:
Thanks for the quick review and merge.

PR #404, opened 2025-04-17
fix: AWS Bedrock documents

mkranjac:
feat: Bedrock has requirement that each document needs unique name so I added fingerprint based on document content
fix: decode document content based on ContentFormat variant

PR #405, opened 2025-04-17
feat: AWS Bedrock document changes

mkranjac:
feat: Bedrock has requirement that each document needs unique name so I added fingerprint based on document content
bugfix: decode document content based on ContentFormat variant
feat: use normalized_documents instead of prompt_with_context


mkranjac:
@0xMochan this PR contains everything from [#404](https://github.com/0xPlaygrounds/rig/pull/404) with changes to `normalized_documents`

rig-bedrock/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "rig-bedrock"
-version = "0.1.1"
+version = "0.1.2"
```
mkranjac, line 3: @0xMochan do we have any policy for bumping crates?

0xMochan: There's no need to, it's automatic by our CI/CD.

PR #409, opened 2025-04-21
fix: rig tool macro struct not public

### LINKED ISSUE #406: bug: Params struct for rig tool macro is private

## Bug Report

The generated args/params struct for the `rig-core-derive` tool macro is private. This causes issues as the generated tool struct itself is public.

## Reproduction

See the image below.

## Expected behavior

It should compile.

## Screenshots

Image

joshua-mo-143:
Fixes #406

### DIFF
--- a/rig-core/rig-core-derive/src/lib.rs
+++ b/rig-core/rig-core-derive/src/lib.rs
@@ -288,7 +288,7 @@ pub fn rig_tool(args: TokenStream, input: TokenStream) -> TokenStream {
 
     let expanded = quote! {
         #[derive(serde::Deserialize)]
-        struct #params_struct_name {
+        pub(crate) struct #params_struct_name {
             #(#param_names: #param_types,)*
         }
 


rig-core/rig-core-derive/src/lib.rs
```diff
@@ -288,7 +288,7 @@ pub fn rig_tool(args: TokenStream, input: TokenStream) -> TokenStream {
 
     let expanded = quote! {
         #[derive(serde::Deserialize)]
-        struct #params_struct_name {
+        pub(crate) struct #params_struct_name {
```
0xMochan, line 291: Does this struct need to be visible outside the crate. when a side-crate is using this macro, it'll create a struct that's visible within the crate it's invoked in and not globally which I think makes sense? but it might be awkward if u want to use that struct in like a side-crate if u have a multi-crate repo.

I think it's fine to do this, we can always make it more public later but i wanted to think about it.

0xMochan (approved):

rig-core/rig-core-derive/src/lib.rs
0xMochan: yea, figured! lets send it!

Issue #381, opened 2025-04-08
feat: improving observability & tracing


0xMochan:
- [x] I have looked for existing issues (including closed) about this

## Feature Request
Standardize and improve tracing calls to a uniform design universal for any provider.

### Motivation
Debugging rig agents especially with complex routing is a pain. Where things are traced and handled is inconsistent amongst most models in formatting and there isn't any standardization.

Observability is an important facet that serious developers will need in order to grow and scale a platform on rig. If a buisness has deployed personalized agents for their clients, it'll be essentially to be able to react and study issues that occur with agents, pulling up logs, and being able to rely on rig for consistent observability for rapid diagnosis.

### Proposal
- all custom tracing logic per provider should be canned unless it's specific (like rig-eternalai's custom prompt logic)
- generic agent tracing should follow a specific pattern and provide extra data in a form that allows for observability
- consider integration with a platform (like [logfire](https://pydantic.dev/logfire))

### Alternatives
- simplified tracing w/ observability at a future date

joshua-mo-143:
Looks like logfire just uses OTel under the hood. I'll have a look and see what potential things we might need to uncover beforehand, as I've worked with `tracing-opentelemetry` before but not entirely sure if we need to create our own subscriber.

joshua-mo-143:
This appears to be complete now that the following have been implemented: #496 (and the many tracing-related PRs that have been added recently)

I'll be closing this issue for the purposes of git hygiene since this is pretty much done, but if there is anything outstanding then we should probably discuss that

PR #411, opened 2025-04-22
Style/trace gemini embedding

atellou:
Changed a print on gemini request to Info tracing message


0xMochan:
Great catch! not sure how this was missed!
Note, we want to greatly improve observability and tracing throughout rig in #381 , which is something we'll be tackling soon!

PR #414, opened 2025-04-23
fix: deepseek function call conversion typo

joshua-mo-143:
Following on from #370, we appear to have missed a very minor detail that would otherwise cause the deepseek function calling stuff to not work.

Thanks @byeblack for the catch!

(edit: Review is *probably* not required here, however seeing as it looks like we're planning to release relatively soon if not next week I've opted to give a heads up just in case)


jizizr:
In fact, apart from OpenAI and Anthropic, other providers have the same issue.

0xMochan (approved):

PR #415, opened 2025-04-23
fix: function call conversion typo

jizizr:
Following on from #414 ,In fact, not only Deepseek, but other providers besides OpenAI and Anthropic also have the same issue.


joshua-mo-143:
Thanks for opening the PR @jizizr ! Currently awaiting some internal discussion (as we haven't actually discussed internally what the path forward is yet due to timezones) but should be moving on this PR pretty quickly soon :slightly_smiling_face:

Issue #420, opened 2025-04-24
bug: Dynamic tools lose context in multi-turn conversations


byeblack:
- [x] I have looked for existing issues (including closed) about this

## Bug Report

Dynamic Tools lose their context during multi-turn conversations when using Agents, causing failure in complex calculations that require maintaining tool information across multiple turns.

## Reproduction

Steps to reproduce:
1. Create multiple dynamic tools (e.g., add, subtract, multiply, divide)
2. Create an Agent using a vendor client model with these dynamic tools
3. Use multi-turn functionality with the prompt: "Calculate ((15 + 25) * (100 - 50)) / (200 / (10 + 10))"
4. Observe the Agent fails to complete the calculation due to lost tool context

Detailed demonstration available at:  
https://github.com/byeblack/rig-multi-turn-demo

## Expected behavior

The Agent should:
1. Maintain awareness of all dynamic tools throughout the multi-turn conversation
2. Successfully break down and compute complex nested expressions
3. Return the correct final result (200 in this case)

## Additional context
- Simple single operations work correctly
- The demo shows both the failure case and what successful behavior should look like

byeblack:
We can solve this problem by caching the prompt in PromptRequest.send and letting Agent.completion recalculate the dynamic context and dynamic tools every time, but this is not the best way.

PR #421, opened 2025-04-24
fix: multi-turn losing dynamic context and tools

byeblack:
Fixes https://github.com/0xPlaygrounds/rig/issues/420

### DIFF
--- a/rig-core/src/agent/prompt_request.rs
+++ b/rig-core/src/agent/prompt_request.rs
@@ -82,6 +82,8 @@ impl<M: CompletionModel> PromptRequest<'_, M> {
         };
 
         let mut current_max_depth = 0;
+        let mut cache_documents = Vec::new();
+        let mut cache_tools = Vec::new();
         // We need to do atleast 2 loops for 1 roundtrip (user expects normal message)
         while current_max_depth <= self.max_depth + 1 {
             current_max_depth += 1;
@@ -94,11 +96,23 @@ impl<M: CompletionModel> PromptRequest<'_, M> {
                 );
             }
 
-            let resp = agent
-                .completion(prompt.clone(), chat_history.to_vec())
-                .await?
-                .send()
-                .await?;
+            let req = if current_max_depth == 1 {
+                let req = agent
+                    .completion(prompt.clone(), chat_history.to_vec())
+                    .await?;
+                cache_documents.extend_from_slice(req.get_documents());
+                cache_tools.extend_from_slice(req.get_tools());
+
+                req
+            } else {
+                agent
+                    .completion(prompt.clone(), chat_history.to_vec())
+                    .await?
+                    .documents(cache_documents.clone())
+                    .tools(cache_tools.clone())
+            };
+
+            let resp = req.send().await?;
 
             chat_history.push(prompt);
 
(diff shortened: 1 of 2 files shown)


0xMochan:
Hey, I appreciate the detailed issue and pull request. I was talking with some coworkers today and I think there was 2 paths forward here:
1. Perform a set of tool caching to ensure tools carry over
2. Gather raggable texts throughout the entire message history instead of just the latest prompt
  a) Alternatively, grab the latest user message from the message history.
  
I do think at some point, we'll add more features to do general text capture from `Message` structs since that struct is quite unwieldy. The tool caching is def. more involved so I do want to think about it in this context, as it might need to be generalized for other situations (like someone ragging on the outside or via the middleware stuff in #346 ).

byeblack:
Yes, this was just my temporary solution and admittedly not well-considered. Fetching dynamic context and tools from message history is indeed a much lighter approach. For example:

```rust  
...      
        // Agent.completion
        // let rag_text = msg.rag_text();  

        let rag_text = chat_history  
            .iter()  
            .chain(std::iter::once(&prompt))  
            .filter_map(|msg| msg.rag_text())  
            .collect::<Vec<_>>()  
            .join("\n")  
            .into();  
...
```  
However, I’m unsure if retrieval will still work reliably when the content becomes too long.

byeblack:
Yes, this was just my temporary solution and admittedly not well-considered. Fetching dynamic context and tools from message history is indeed a much lighter approach. For example:

```rust  
...      
        // Agent.completion
        // let rag_text = msg.rag_text();  

        let rag_text = chat_history  
            .iter()  
            .chain(std::iter::once(&prompt))  
            .filter_map(|msg| msg.rag_text())  
            .collect::<Vec<_>>()  
            .join("\n")  
            .into();  
...
```  
However, I’m unsure if retrieval will still work reliably when the content becomes too long.

joshua-mo-143:
> Yes, this was just my temporary solution and admittedly not well-considered. Fetching dynamic context and tools from message history is indeed a much lighter approach. For example:
> 
> ```rust
> ...      
>         // Agent.completion
>         // let rag_text = msg.rag_text();  
> 
>         let rag_text = chat_history  
>             .iter()  
>             .chain(std::iter::once(&prompt))  
>             .filter_map(|msg| msg.rag_text())  
>             .collect::<Vec<_>>()  
>             .join("\n")  
>             .into();  
> ...
> ```
> 
> However, I’m unsure if retrieval will still work reliably when the content becomes too long.

What about if we try getting messages for simply the last `x` or so messages? That way it might still be able to keep the content relevant while still allowing new rag tools to be brought in (by virtue of bringing in fresh texts from messages).

0xMochan:
> What about if we ragging text from the last `x` or so messages? That way it might still be able to keep the content relevant while still allowing new rag tools to be brought in (by virtue of bringing in fresh texts from messages).

The most common case of this problem is with `.multi_turn` with a variable `max_depth` set. This means `x` would end up needing to be atleast `max_depth` to capture the earliest user message.

---
There's actually a deeper problem with dynamic tool calling. In a longer chain of interactions, you might invoke a dynamically called tool earlier in the message history, but with a re-prompt, you may not pull up the called tool. This means there will be tool calls in message history that don't have tools from the provider's POV. This can cause some weird behavior that I think gets solved by just prompting the provider in the preamble that there might be tool calls that are from previous invocations (I've seen this in Goose). I suppose this is up to the user to realize as us editting preambles is quite invasive.
---

I think the most suitable solution, is traversing up chat_history and ragging based off of the latest user message that has raggable text.

0xMochan:
Superceded by #425

PR #423, opened 2025-04-25
fix: OpenAI provider streaming tool call response for local LLM

kosz78:
This PR fix problem when we work with local LM using LM Studio over OpenAI provider and use streaming tool call response.
LM Studio respond function name `Some("")` for arguments chunks


0xMochan:
This might be handled in the upcoming #388 streaming refactor

PR #424, opened 2025-04-25
feat: think tool, vector store tool, better agent tool

### LINKED ISSUE #408: feat: agent as tool

## Feature Request
Be able to attach an agent as a tool. This would allow agent-directed flow since the agent itself can choose what agents to refer to.

### Motivation
- Studying the [OpenAI Agent SDK](https://github.com/openai/openai-agents-python/blob/main/examples/agent_patterns/agents_as_tools.py) led to this idea.

### Proposal
- Implement `Tool` for `Agent` (and maybe `VectorStore`)

### Alternatives
- Wrapper `Tool`

0xMochan:
This PR adds 2 major tools alongside some intricate examples:

## [Think Tool](anthropic.com/engineering/claude-think-tool)
This was introduced in an anthropic blog post and primarily gives the agent a bit of space to "think" before producing a final result. This primarily requires some sort of agent looping which is handled by the `.prompt(...).multi_turn(...)` aspect from the multi-turn pr (#370).

## Vector Store as a tool
Similar to agents, this presents an alternative to ragging via tool calling rather than workflow based ragging (explicitly done during `agent.completion`). This form of searching works super well with multi-turn agents esp with the think tool.

---

This PR _also_ fixes an issue with the current agent tool implementation (from #696) which leads to agent tools not using the name of the agent (which means agents that leverage multiple agent tools would not function).

## Examples
- `anthropic_think_tool` | this showcases the original think tool on it's own
- `anthropic_think_tool_with_other_tools` | this showcases how the think tool can _improve_ other tool usage
- `complex_agentic_loop_claude` | this showcases a complex loop using think tool, agents as tools, and vector store as tools.


joshua-mo-143:
I ended up just removing the Sync requiring for `call()` and it seems to compile just fine after moving everything to standard async.

If we still want to progress on this, I'd be happy to take over this ticket

joshua-mo-143 (changes requested):
Looks mostly OK I think.

With regards to `think_tool`: should we make a new folder for tools then add it to a `think.rs` file? Seems more appropriate than having `think_tool.rs` in the top level (and we can still export it as `use tool::think as think_tool`)

rig-core/src/vector_store/mod.rs
```diff
@@ -118,3 +121,59 @@ fn prune_document(document: serde_json::Value) -> Option<serde_json::Value> {
...
 }
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct VectorStoreOutput {
+    pub score: f64,
+    pub id: String,
+    pub document: Value,
+}
+
+impl<T: VectorStoreIndex> Tool for T {
```
joshua-mo-143: please use `where` clause instead of syntax sugar trait bounds (keeps docs.rs nice and tidy as it doesn't seem to format the syntax sugar trait bounds very well...)

joshua-mo-143 (changes requested):
Just one minor change then I think we can push this through

0xMochan:
> Looks mostly OK I think.
> 
> With regards to `think_tool`: should we make a new folder for tools then add it to a `think.rs` file? Seems more appropriate than having `think_tool.rs` in the top level (and we can still export it as `use tool::think as think_tool`)

I agree! this is sorta a relic of when i first made this PR months ago.

rig-core/src/vector_store/mod.rs
0xMochan: it's a shame that they added this syntax sugar but didn't auto-generate the proper docs for it. feels like an oversight on the gen tool.

joshua-mo-143:
@0xMochan looks like just one thing to fix then this is good to merge

PR #425, opened 2025-04-25
fix: dynamically pull rag text from chat history

### LINKED ISSUE #420: bug: Dynamic tools lose context in multi-turn conversations

0xMochan:
Fixes #420 and supersedes #421 .

The main fix is via not gathering the raggable text from only the current prompt but also going up from latest to oldest methods.

Basically, if the prompt has no raggable text, traverse backwards up chat history until you find some.

### DIFF
--- a/rig-core/examples/rag_dynamic_tools_multi_turn.rs
+++ b/rig-core/examples/rag_dynamic_tools_multi_turn.rs
@@ -1,0 +1,186 @@
+use anyhow::Result;
+use rig::{
+    completion::{Prompt, ToolDefinition},
+    embeddings::EmbeddingsBuilder,
+    providers::openai::{Client, TEXT_EMBEDDING_ADA_002},
+    tool::{Tool, ToolEmbedding, ToolSet},
+    vector_store::in_memory_store::InMemoryVectorStore,
+};
+use serde::{Deserialize, Serialize};
+use serde_json::json;
+use std::env;
+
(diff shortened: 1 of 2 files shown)


0xMochan:
CC @byeblack

byeblack:
> CC @byeblack

Thank you for your proactive efforts on this issue. After some testing, it has proven to effectively resolve my current problem while maintaining clean implementation.

joshua-mo-143:
lgtm ser

PR #428, opened 2025-04-27
Donot use async closure + Bump mcp-core

tarunKoyalwar:
## Proposed Changes

- Remove use of async closure 
> ~Async closures are unstable in Rust 2021 edition, but are stabilized in Rust 2024 edition~, needs rust 1.85
- Bump mcp-core , mcp-core-macros and address compile errors 
> 1. fix missing match case
> 2. add annotations in tool macro use

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6519,9 +6519,9 @@ dependencies = [
 
 [[package]]
 name = "mcp-core"
-version = "0.1.42"
+version = "0.1.46"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "18a8e4375fe44579ffd9c7a8d682059ee33f1bf2bf36cad94cbe1ebff2e5b111"
+checksum = "2766f3db0a24119127ba5e380539f2a6351210456a6911de28dbaa8ce2761080"
 dependencies = [
  "actix-web",
  "anyhow",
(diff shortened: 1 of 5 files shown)


tarunKoyalwar:
cc: @0xMochan

0xMochan:
Hey, I appreciate the PR. Async closures are available in Rust Edition 2021, they bump our min rust version from 1.81 to 1.85. The cargo.toml is set for rust edition 2021.

Is there a reason why you need the older rust version?

tarunKoyalwar:
Hey @0xMochan, thanks for the context!
Locally, I’m still using Rust 1.84, and it works fine with the latest release. I wanted to try out the `multi_turn`method/feature from the main branch but ran into a compiler error. The error originated from a specific statement, and since it didn’t seem like a common pattern, I thought it would be helpful to raise a PR to address it, along with the related changes for mcp-core.

tarunKoyalwar:
By the way, just curious — do you have any MSRV policy, or is it still undefined at this point?

0xMochan:
It's def. a good time for us to solidify a good policy here, we've talked internally about it. I'll flag it tomorrow!

I do agree, i was a bit hasty using it here and likely an `async move` block was sufficient (i presume this block was much more complex during dev hence why i grabbed async closure off the rip). I've been pretty excited to see this feature stabilized.

I think, for a library that moves as fast as we do and considering rust's tenacity for backwards compatibility, I'd look towards a MSRV policy that keeps us modern.

0xMochan (approved):

PR #429, opened 2025-04-28
fix: update broken link 

PixelPil0t1:
I found that the previous link pointed to a non-existing repository.
I updated it to the current working repository RIG_AI/RIG_AI, as there are no other available alternatives at the moment.
If you have a better or more specific repository in mind, please suggest it, and I will update the link accordingly.


0xMochan:
Hi, appreciate the PR here. I think the fixed url is wrong. The correct one is https://github.com/0xPlaygrounds/rig

Issue #419, opened 2025-04-24
bug: Impossible to build and agent without a system prompt using certain models


aheart:
- [x] I have looked for existing issues (including closed) about this

## Bug Report

Reproducible on rig-core 0.11.1
Some OpenAI models don't support messages with role System (e.g. o1-mini).
When creating an agent using such models you get an error:
```
ProviderError: {
  "error": {
    "message": "Unsupported value: 'messages[0].role' does not support 'system' with this model.",
    "type": "invalid_request_error",
    "param": "messages[0].role",
    "code": "unsupported_value"
  }
}
```
I expected that removing the preamble would fix this, but I kept getting the same error.
Looking at the code I noticed that the preamble is not an Option<String> here: https://github.com/0xPlaygrounds/rig/blob/870f29a771d6d0c35faf211157c3cdc99124ff2c/rig-core/src/agent/completion.rs#L44
And during conversion it's populated with an empty string here:
https://github.com/0xPlaygrounds/rig/blob/f4214540a5e90e166d17235593cd9863899c5aba/rig-core/src/agent.rs#L478

That's why eventually an empty system message will be added here:
https://github.com/0xPlaygrounds/rig/blob/f4214540a5e90e166d17235593cd9863899c5aba/rig-core/src/providers/openai/completion.rs#L610

## Reproduction

Create agent using OpenAI's o1-mini model. Send a completion request.

## Expected behavior

No system message is added to the history if no preamble is set.

joshua-mo-143:
This appears to be a design flaw in the way that agent/completion requests are built. 

We could probably add a field or method to represent whether or not a given feature should be utilised or not (and then go from there). I would have preferred to add it as a const value within a trait, but given our current abstraction that's not really feasible.

0xMochan:
It absurd how inconsistent these models are. the o-series of models from OpenAI _do_ take a preamble, but instead of defining it as a `System` prompt they take it as a `Developer` prompt.

They used to auto-coerce this for us, but I guess not.

0xMochan:
After further investigation, this **only** applies to `o1-mini`

aheart:
It's very confusing. `o1-mini` doesn't appear to support the developer role either. The only viable option might be to include the system prompt with the user role. It might even be better to just give up on using `o1-mini` for agents.
https://community.openai.com/t/developer-role-not-accepted-for-o1-o1-mini-o3-mini/1110750/7

joshua-mo-143:
[...]

PR #430, opened 2025-04-28
fix: system and developer messages for openai

0xMochan:
Using `developer` only for `o1-mini` instead of all models.

Reasoning:
- Some people use the openai client for other providers
- Checking for models that start with `o` is fallible for alternative providers
- `system` messages work for the newer reasoning models.

### DIFF
--- a/rig-core/src/providers/openai/completion.rs
+++ b/rig-core/src/providers/openai/completion.rs
@@ -177,6 +177,7 @@ pub struct Choice {
 #[derive(Debug, Serialize, Deserialize, PartialEq, Clone)]
 #[serde(tag = "role", rename_all = "lowercase")]
 pub enum Message {
+    #[serde(alias = "developer")]
     System {
         #[serde(deserialize_with = "string_or_one_or_many")]
         content: OneOrMany<SystemContent>,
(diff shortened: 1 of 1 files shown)


rig-core/src/providers/openai/completion.rs
```diff
@@ -647,21 +648,27 @@ impl CompletionModel {
...
-                let mut val = serde_json::to_value(&msg)?;
-                if let Message::System { .. } = msg {
-                    // TODO: better detection of `o` models
-                    if self.model.starts_with('o') {
+        // o1-mini crashes if not using `developer` role for system messages. We can't globally
+        //  change openai due to other providers that can be used with openai.
+        //
+        // I'm hesitant to only chk models that start with `o` as that can interfere with other
+        //  providers.
+        let full_history = if self.model.starts_with("o1-mini") {
```
joshua-mo-143, line 656: This doesn't actually work. ref #419 

See the link: https://community.openai.com/t/developer-role-not-accepted-for-o1-o1-mini-o3-mini/1110750/7

(o1-mini and o1-preview don't support system **or** developer messages, lmao)

joshua-mo-143:
We ended up reverting the preliminary changes and merging this due to the aforementioned link of o1 not supporting system **or** developer messages.

If you're reading this from the future: don't use `o1` models with preamble, it'll just cause OpenAI to tell you that system (or developer, respectively) messages are not supported.

Issue #431, opened 2025-04-29
bug: latest changes merged into main breaks example test in AWS Bedrock


mkranjac:
- [x] I have looked for existing issues (including closed) about this

## Bug Report
Latest changes on `completion_requests` discovered bug with base64 encoding of context documents.

## Reproduction
Running `rig-bedrock/examples/agent_with_bedrock.rs` returns an error on test with context

## Expected behavior
existing examples shouldn't break

mkranjac:
Fixed with #432

Issue #407, opened 2025-04-21
feat: Support Cloning VectorStoreIndex and Idempotent Qdrant Inserts


mztlive:
I have looked for existing issues (including closed) about this

Feature Request

Improve the usability of VectorStoreIndex and the document insertion behavior in the Qdrant VectorStore implementation.

Motivation

In most real-world applications, VectorStoreIndex is initialized once and reused throughout the system. However, since the type does not implement Clone, it becomes inconvenient to pass it across components or threads. The lack of clonability limits flexibility, especially in multi-threaded or modular applications.

Additionally, in the Qdrant VectorStore implementation, every inserted point is assigned a UUID by default (PointStruct.id = Uuid::new_v4().to_string()). This leads to duplicate entries even when inserting the same document multiple times, which breaks idempotency and unnecessarily bloats the database.

Proposal
	1.	Make VectorStoreIndex clonable or easily shareable
	•	If Clone cannot be implemented due to internal non-cloneable state (e.g. database clients, async runtimes), an Arc<VectorStoreIndex>-based pattern can be officially supported or recommended.
	•	Optionally, expose a .shareable() method or a helper wrapper that encapsulates sharing logic.
	2.	Allow user-defined IDs when inserting points into Qdrant
	•	Provide an option to pass a custom document ID (e.g. a hash of the document content) when constructing PointStruct, enabling idempotent inserts.
	•	Introduce a configuration flag to toggle between Uuid and content-derived IDs.

Alternatives
	•	Manually wrap VectorStoreIndex in an Arc outside the library, though this puts the burden on the user and leads to inconsistent usage patterns.
	•	Override the UUID generation behavior in Qdrant inserts manually, but this requires understanding of internal implementation details and reduces out-of-the-box usability.

joshua-mo-143:
What about having an ID generation strategy for inserts similar to the `NonceSequence` strategy from `ring`? We could then probably extrapolate some defaults out of that which would make it much easier to generate IDs.

mztlive:
> What about having an ID generation strategy for inserts similar to the [`NonceSequence` trait from `ring`](https://docs.rs/ring/latest/ring/aead/trait.NonceSequence.html)? We could then probably extrapolate some defaults out of that which would make it much easier to generate IDs.
> 
[...]

PR #432, opened 2025-04-29
fix: fixed bug with base64 encoding on AWS Bedrock

mkranjac:


mkranjac:
relates to [issue](https://github.com/0xPlaygrounds/rig/issues/431)

rig-bedrock/src/types/document.rs
```diff
@@ -1,41 +1,60 @@
...
 pub(crate) use crate::types::media_types::RigDocumentMediaType;
 use base64::{prelude::BASE64_STANDARD, Engine};
+use ring::digest::{Context, Digest, SHA256};
 
 #[derive(Clone)]
 pub struct RigDocument(pub Document);
 
+impl RigDocument {
+    pub fn fingerprint(&self) -> String {
+        let mut context = Context::new(&SHA256);
```
mkranjac, line 16: This is unfortunate constraint of AWS Bedrock. 
All documents needs to have a unique name, otherwise it returns an error, so I added lib `ring` to create name from document bytes.

 If you have some better idea I am open to suggestion

0xMochan: hmm, other ideas would be a global counter that can be used or a random UUID?

joshua-mo-143: I've commented on this in #407 but would be a good idea to either have a way for users to write their own ID generation strategy or just use a preset one

mkranjac: I think @0xMochan is correct with random UUID, because if you have long history you might upload same document multiple times and it will result with an error with fingerprint approach... 
If you are all OK with this, I will change PR to use random UUID for documents names.

joshua-mo-143: Yeah I think generating random uuids is fine for now. We can retroactively fix this later on if need be.

0xMochan:
Is this ready for review?

rig-bedrock/src/types/document.rs
```diff
@@ -15,27 +15,32 @@ impl TryFrom<RigDocument> for aws_bedrock::DocumentBlock {
...
+            Some(ContentFormat::Base64) => BASE64_STANDARD
+                .decode(value.0.data)
+                .map_err(|e| CompletionError::ProviderError(e.to_string()))?,
+            _ => value.0.data.as_bytes().to_vec(),
+        };
+
         let data = aws_smithy_types::Blob::new(document_data);
         let document_source = aws_bedrock::DocumentSource::Bytes(data);
 
+        let document_name = Uuid::new_v4().simple().to_string();
```
0xMochan, line 39: maybe format this as like "document-{UUID}". This is a name that a LLM will see, so i think the name should reflect that a bit. (feel free to improve that).

0xMochan (changes requested):
Small notes but otherwise should be good

mkranjac:
@0xMochan updated PR with requested changes

0xMochan (approved):
lgtm! good fixes

Issue #434, opened 2025-04-30
feat: Tool examples refactor


mkranjac:
- [x] I have looked for existing issues (including closed) about this

## Feature Request
Tool examples refactor

### Motivation
We have at the moment >> 30 copy-paste tool examples, mostly `Add`, `Adder` and `Subtract`.
It would remove a lot of duplicate code if we just reuse implementation from one place

### Proposal
Refactor all common example tool(s) into a separate crate

0xMochan:
Yea, this would be a short-term win with our examples. tbh, the current set of examples is serving as a stopgap for our lack of comprehensive integration testing. Our short-medium term plan is to provide some GA that runs only on the genereated release PR so that we can comprehensively test for regressions thru the codebase.

This should let us clean up a ton of the built-in examples (maybe even we can add folders and such) and make it more consistent. This issur would be a great short-term win for rig tho to make it much easier to parse thru the examples!

PR #436, opened 2025-05-02
Feat/rig shared

mkranjac:
Relates to [issue #434](https://github.com/0xPlaygrounds/rig/issues/434)

Couple of notes:
- Most of the crates have similar dependencies but with small variations on minor versions, using Workspace dependencies aims to simplify updates and removes redundant compilation. If you have any drawbacks with this approach I will revert it back to original.
- Added fixtures for most commonly used mocks ( mostly for embeddings )
- I named this crate `rig-shared`, at the moment it’s not ideal because it’s mostly for fixtures and common tools but if you have better alternative I will rename it
- currently only AWS Bedrock crate is using this shared crate just for POC, if everything else is good I can refactor other crates.


mkranjac:
@0xMochan @joshua-mo-143 please let me know if this is correct approach

0xMochan:
TBH, i'm not too fond of a crate (even if internal) that is only used to reduce duplication among examples. the examples' duplications come from the fact that we lack a proper integration testing suite. internally, we've discussed introducing and rolling some examples as apart of said suite on running against commits against our release branches. since these only get bumped on the PRs that get merged (+ we can always cancel if needed), it'll ensure that we don't burn too many credits in testing.

Then, we can skim through and try to isolate examples that actually highlight a specific use-case (rather than just duplicating every example for every provider). There's also a case of some examples that highlight openai-compatible services that could be collapsed into a single example, though I'm not entirely sure what that would look like.

joshua-mo-143:
I agree with @0xMochan here.

On the other hand, we could really do with slimming down the examples list as there are now 80(!!!!) items in the examples folder. It's likely we will need to do some preliminary work first to figure out what examples we actually need to keep for integration testing and which ones we should either keep in the `rig-core` examples or turn into a cookbook of some kind.

Issue #378, opened 2025-04-03
feat: Add support for Mistral


joshua-mo-143:
## Model Provider Integration Request

It's Mistral.

### Resources

https://docs.mistral.ai/api/

joshua-mo-143:
> Any help needed? I would like to help.

Feel free!

PR #437, opened 2025-05-03
Add mistral provider

adrianncovaci:
## Resources
https://docs.mistral.ai/api/

See #378


joshua-mo-143:
Hi, thanks for creating this PR!

Can you also please add an example for chat completions?

adrianncovaci:
> Can you also please add an example for chat completions?

Do you mean an example which does the actual request?

Other providers test the chat completion response by deserializing the json type from their docs to the actual rust struct defined by the provider. I do the same via `test_response_deserialization`, perhaps I should rename this test for clarity?

joshua-mo-143:
You need to add the following macro below for the `embed_texts` function:

```rust
#[cfg_attr(feature = "worker", worker::send)]
```

It will make the function WASM compliant. Should be good to go I think other than that!

joshua-mo-143:
Lgtm - tested

PR #440, opened 2025-05-06
refactor: Introduce Client Traits and Testing

### LINKED ISSUE #149: feat: client versioning

yavens:


joshua-mo-143:
~~@yavens It looks like the formatting has gone really strange and now there's an extra linebreak between every single LoC, is this intentional?~~

Nevermind, I'll just update it myself :sweat_smile:

joshua-mo-143 (approved):
lgtm

PR #442, opened 2025-05-07
fix: OpenAI provider streaming tool call response for local LLM

kosz78:
This PR fix problem when we work with local LM using LM Studio over OpenAI provider and use streaming tool call response.
LM Studio respond function name Some("") for arguments chunks


0xMochan:
Calling on @yavens since they primary adjusted this code during the #388 PR.

yavens:
Gotta love OpenAI "compatibility"

Change looks fine to me, will need to run the streaming examples for the other OAI providers to make sure they didn't break.

joshua-mo-143:
tested, LGTM

joshua-mo-143 (approved):

Issue #452, opened 2025-05-17
bug: Wrong dim value for gemini EMBEDDING_004


adrianncovaci:
According to https://ai.google.dev/gemini-api/docs/models , the output dimension size for `text-embedding-004` is 768.

https://github.com/0xPlaygrounds/rig/blob/main/rig-core/src/providers/gemini/embedding.rs#L39

this is set to `1024`

PR #453, opened 2025-05-17
fix: Fix `dims` value for gemini's `EMBEDDING_004` (#452)

adrianncovaci:


joshua-mo-143:
LGTM, thanks for sending in the fix!

(note: ref https://ai.google.dev/gemini-api/docs/models#text-embedding)

PR #460, opened 2025-05-24
fix(extractor): correct typo in extractor prompt

QwF8PatkuU:
Fix a typo by changing "event with default values" to "even with default values" in the extractor prompt.


joshua-mo-143 (approved):
lgtm - thanks for the PR

PR #461, opened 2025-05-24
fix(message): correct ToolCall to Message conversion

### LINKED ISSUE #459: Message From ToolCall Type Is Correct?
I'm not very clear on why ToolCall is converted to ToolResult when it becomes a Message?

It seems to be assistant content and tool call?

```rust
impl From<ToolCall> for Message {
    fn from(tool_call: ToolCall) -> Self {
        Message::User {
            content: OneOrMany::one(UserContent::ToolResult(ToolResult {
                id: tool_call.id,
                content: OneOrMany::one(ToolResultContent::text(tool_call.function.name)),
            })),
        }
    }
}
```

QwF8PatkuU:
Fixes #459

Correct ToolCall to Assistant message conversion

### DIFF
--- a/rig-core/src/completion/message.rs
+++ b/rig-core/src/completion/message.rs
@@ -601,11 +601,8 @@ impl From<OneOrMany<UserContent>> for Message {
 
 impl From<ToolCall> for Message {
     fn from(tool_call: ToolCall) -> Self {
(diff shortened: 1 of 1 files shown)


joshua-mo-143 (approved):
lgtm

PR #463, opened 2025-05-25
feat(rig-744): Add support for Milvus vector store

### LINKED ISSUE #451: feat: Add support for milvus vector store

Milvus is an open-source vector database that suits AI applications of every size from running a demo chatbot in Jupyter notebook to building web-scale search that serves billions of users. In this guide, we will walk you through how to set up Milvus locally within minutes and use the Python client library to generate, store and search vectors.

### Resources

Source Code: [https://github.com/milvus-io/milvus](https://github.com/milvus-io/milvus)
Rust SDK: [https://github.com/minio/minio-rs](https://github.com/minio/minio-rs)
Official Documents: https://milvus.io/docs/quickstart.md

joshua-mo-143:
Fixes #451 
Fixes RIG-744

### DIFF
--- a/rig-milvus/src/lib.rs
+++ b/rig-milvus/src/lib.rs
@@ -1,0 +1,288 @@
+use reqwest::StatusCode;
+use rig::{
+    Embed, OneOrMany,
+    embeddings::{Embedding, EmbeddingModel},
+    vector_store::{VectorStoreError, VectorStoreIndex},
+};
+use serde::{Deserialize, Serialize};
+
+/// Represents a vector store implementation using Milvus - <https://milvus.io/> as the backend.
+pub struct MilvusVectorStore<M> {
+    /// Model used to generate embeddings for the vector store
+    model: M,
+    base_url: String,
+    client: reqwest::Client,
+    database_name: String,
+    collection_name: String,
+    token: Option<String>,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct CreateRecord {
+    document: String,
+    embedded_text: String,
+    embedding: Vec<f64>,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct InsertRequest<'a> {
+    data: Vec<CreateRecord>,
+    collection_name: &'a str,
+    db_name: &'a str,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct SearchRequest<'a> {
+    collection_name: &'a str,
+    db_name: &'a str,
+    data: Vec<f64>,
+    anns_field: &'a str,
+    limit: usize,
+    output_fields: Vec<&'a str>,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct SearchResult<T> {
+    code: i64,
+    data: Vec<SearchResultData<T>>,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct SearchResultData<T> {
+    id: i64,
+    distance: f64,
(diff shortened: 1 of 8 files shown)


joshua-mo-143:
Workflow cancelled due to unnecessary jobs being run (I basically only updated the README and the previous `clippy` commit made the CI pass)

rig-milvus/src/lib.rs
```diff
@@ -0,0 +1,288 @@
...
+            let res = res.text().await.unwrap();
+
+            return Err(VectorStoreError::DatastoreError(res.to_string().into()));
+        }
+
+        Ok(())
+    }
+}
+
+impl<M: EmbeddingModel + std::marker::Sync + Send> VectorStoreIndex for MilvusVectorStore<M> {
```
0xMochan, line 192:
```suggestion
impl<M: EmbeddingModel + Sync + Send> VectorStoreIndex for MilvusVectorStore<M> {
```

i think we generally don't use full formed paths for `Send+Sync` anywhere else (unless this was copied from another implementation)

rig-milvus/src/lib.rs
```diff
@@ -0,0 +1,288 @@
...
+        let body = serde_json::to_string(&body).unwrap();
+
+        let res = client
+            .body(body)
+            .send()
+            .await
+            .map_err(|x| VectorStoreError::DatastoreError(x.to_string().into()))?;
+
+        if res.status() != StatusCode::OK {
+            let res = res.text().await.unwrap();
```
0xMochan, line 270: Is this `unwrap` safe? I think either use a `.expect` OR probably bubble the error up. I guess it's only when the response is errored but i suppose a default error message could be "failed getting error message".

*tbh, we could think about enabling a `.unwrap` lint in clippy.*

rig-milvus/src/lib.rs
```diff
@@ -0,0 +1,288 @@
...
+    Embed, OneOrMany,
+    embeddings::{Embedding, EmbeddingModel},
+    vector_store::{VectorStoreError, VectorStoreIndex},
+};
+use serde::{Deserialize, Serialize};
+
+/// Represents a vector store implementation using Milvus - <https://milvus.io/> as the backend.
+pub struct MilvusVectorStore<M> {
+    /// Model used to generate embeddings for the vector store
+    model: M,
```
0xMochan, line 12: This might be opinionated but, is it better to keep the bounds of `M` on the struct itself instead of the method. like, theoretically, you could create a vector store given a model `M` that isn't bound my `EmbeddingModel` but you wouldn't be able to do anything with it.

I think i've seen both ways in the wild (rig-surrealdb seems to restrict the struct via generic bounds).

rig-milvus/src/lib.rs
```diff
@@ -0,0 +1,288 @@
...
+        let body = serde_json::to_string(&body).unwrap();
+
+        let res = client
+            .body(body)
+            .send()
+            .await
+            .map_err(|x| VectorStoreError::DatastoreError(x.to_string().into()))?;
+
+        if res.status() != StatusCode::OK {
+            let res = res.text().await.unwrap();
```
0xMochan, line 222: ditto on this `unwrap` (i saw the other `unwrap` first)

0xMochan (changes requested):
A couple of nitpicks/changes, also testing this locally still

rig-milvus/src/lib.rs
joshua-mo-143: Copied from the Qdrant impl - we might need to investigate and check whether or not the bounds are actually required in the Qdrant case in a later PR.

joshua-mo-143:
Pushed a new commit to clean up some of the incidental slop that I pushed yesterday. I've additionally added a new error type to `VectorStoreError` to account for when we have to do manual HTTP requests (when the third party service doesn't have a Rust SDK)

0xMochan (approved):
looks good!

Issue #446, opened 2025-05-11
bug: Document Content Dropped by Some Providers When Using UserContent::Document


hollygrimm:
- [X] I have looked for existing issues (including closed) about this

## Bug Report
There's an issue with how document content is handled across different LLM providers. Currently, when using `UserContent::Document` for [document content](https://github.com/0xPlaygrounds/rig/blob/main/rig-core/src/completion/request.rs#L278), the behavior varies inconsistently by provider:

- **Ollama**: Completely drops document content from messages sent to the model
- **Anthropic**: Processes the Document as a PDF and errors out
- **OpenAI**: Correctly converts it to text content and works as expected

This bug was introduced in commit [2d45ad52f61dc8e21ab1e6fc08fe0096f3167ebf](https://github.com/0xPlaygrounds/rig/commit/2d45ad52f61dc8e21ab1e6fc08fe0096f3167ebf#diff-6598f1c9b2863bc96f16fa07d3f5eaf739f0b4b19333ed220f1a80be179548e6R572)

## Reproduction
1. Run the [rag_ollama.rs](https://github.com/0xPlaygrounds/rig/blob/main/rig-core/examples/rag_ollama.rs ) example
2. Observe that document content is missing from the chat payload sent to the model

## Expected behavior
The LLM should receive the document content as context so it can properly answer questions based on that context. Currently, with Ollama, this doesn't happen - the documents are processed but never included in the payload.

## Proposed fix
Ensure all providers properly handle `UserContent::Document` by updating their `TryFrom` implementations to check for document content and convert it appropriately based on the `DocumentMediaType`.

For example, in Ollama's implementation, document content is currently dropped here:
```rust
match uc {
    crate::message::UserContent::Text(t) => texts.push(t.text),
    crate::message::UserContent::Image(img) => images.push(img.data),
    _ => {} // Document content is dropped here
}
```
And Anthropic always processes the Document as a PDF regardless of the actual content type:
```rust
message::UserContent::Document(message::Document { data, format, .. }) => {
    let source = DocumentSource {
        data,
        media_type: DocumentFormat::PDF,
        r#type: match format {
            Some(format) => format.try_into()?,
            None => SourceType::BASE64,
        },
    };
    Ok(Content::Document { source })
}
```

Temporary workaround: Our current fix is to use `UserContent::Text` instead of `UserContent::Document` in the `normalized_documents()` method, but the right approach would be for all providers to handle document content correctly.

[...]

PR #465, opened 2025-05-26
fix(anthropic): Only PDF docs are supported by their API

joshua-mo-143:
In reference to #446, it was noticed that the business logic for Anthropic's documents actually only support PDF. We should be enforcing this in our own API to ensure that users see the expected behaviour rather than automatically setting it to PDF.


0xMochan (approved):
Looks good 👍🏽

PR #492, opened 2025-06-08
fix(rig-773): xAI embeddings endpoint is wrong

### LINKED ISSUE #490: xAI embeddings endpoint is wrong
The xAI embedding endpoint is wrong.<br><br>In their API reference, it is actually referred to as "tokenize-text".<br><br>Ref: [https://docs.x.ai/docs/api-reference#tokenize-text](https://docs.x.ai/docs/api-reference#tokenize-text)

joshua-mo-143:
Fixes #490 
Fixes RIG-773

See this API reference: https://docs.x.ai/docs/api-reference#tokenize-text


cvauclair (approved):

cvauclair (changes requested):
Holdup, I don't think this is actually an embedding endpoint :thinking: 

According to their [swaggerUI docs](https://api.x.ai/docs/#/), they actually have an undocumented embedding endpoint.

Tokenization is a very different operation than embedding.

joshua-mo-143:
> Holdup, I don't think this is actually an embedding endpoint 🤔
> 
> According to their [swaggerUI docs](https://api.x.ai/docs/#/), they actually have an undocumented embedding endpoint.
> 
> Tokenization is a very different operation than embedding.

That's a bit weird then all things considered. @ProHaller brought up on Discord that the `/v1/embeddings` endpoint doesn't actually exist... so that might be a bit of an issue.

I think if this blocks a release, we should just wait until the next release. This does not seem to be an immediate enough issue that we have to urgently solve this beforehand.

cvauclair:
If the `/v1/embeddings` endpoint doesnt actually exist, then that means xAI just doesn't support embeddings and we can just remove the models/integrations for it

cvauclair (approved):

PR #495, opened 2025-06-08
feat(rig-760): enable reasoning for openrouter

### LINKED ISSUE #476: feat: support reasoning for OpenRouter

## Feature Request

OpenRouter supports reasoning by adding a `reasoning` field onto the assistant message. We should support this.

### Motivation

Higher provider compatibility => more Rig adoption

joshua-mo-143:
Fixes #476 
Fixes RIG-760

Marked as draft until an agreement is reached that this is the final implementation.

The reason why this is a draft is because it's currently unclear whether we either want to do this, or to have an `additional_params` variable (of `serde_json::Map<String, serde_json::Value>` type) to soak up all extra parameters in the OpenAI message type (using `serde::flatten`).

### DIFF
--- a/rig-core/examples/openrouter_reasoning.rs
+++ b/rig-core/examples/openrouter_reasoning.rs
@@ -1,0 +1,56 @@
+use anyhow::{anyhow, Result};
+use rig::completion::Completion;
+use rig::prelude::*;
+use rig::providers::openrouter::Message;
+
+use rig::providers;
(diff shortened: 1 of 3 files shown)


joshua-mo-143:
Closing because we're officially adding reasoning/thinking soon and therefore will use a new branch to work off of rather than just trying to blindly integrate it per provider.

PR #508, opened 2025-06-11
feat(rig-780): integrate openAI responses API

### LINKED ISSUE #227: feat: `strict` parameter in ToolDefinition

### LINKED ISSUE #400: feat: openai responses api

## Feature Request
- Integrate the responses API to be used alongside raw completions.
- Create new traits for the new paradigms of server-side chat history, etc.
- Further the integration by enabling usage via Agents (potentially via post-#346)

### Motivation
OpenAI seemingly has cracked a new API paradigm (rip standardization) and is now moving towards the Responses API as the main way to interact with their models.

[Image](https://platform.openai.com/docs/guides/tools-computer-use#overview)
> _Computer use seemingly is locked behind the responses api_

The responses api also has features to store the message history via chat id tracking server-side. This isn't necessary for every use-case but is a use-case we might want to think about supporting.

### Proposal
- [ ] Create a `responses` function endpoint on the OpenAI Client for raw usage with the endpoint.
- [ ] Create a set of traits to generalize said endpoints (perhaps waiting to see how industry reacts to the new endpoints).

### Alternatives
- Avoid doing traits and integration with agents until there's more data. people can make server-side traits.

### LINKED ISSUE #503: Integrate OpenAI Responses API
### Description

* OpenAI has recently released their "Responses API" which aims to be a simpler API for agentic workflows involving LLMs
* Although the completion API is mostly at parity with this new API and will be maintained indefinitely, it is missing one crucial feature: it does not expose the reasoning performed by CoT models!

### Implementation

* Migrate our OpenAI integration to the Responses API
  * Note: We should keep the completion types since some of these are being used by other provider integrations
* Here is a rough mapping of how we can map types/fields of our common models to the Responses API request/response models
  * Request
    * System prompt: Either use the `instructions` field or provide a "developer" or "system" message in the `input` field
    * Chat history -> [`input`](https://platform.openai.com/docs/api-reference/responses/create#responses-create-input) field accepts an array of developer/user/assistant messages
    * Tools -> [`tools`](https://platform.openai.com/docs/api-reference/responses/create#responses-create-tools) field
  * Response
    * Response message, reasoning summary and/or tool call -> [`output`](https://platform.openai.com/docs/api-reference/responses/object#responses/object-output) field 
    * Usage metrics -> [`usage`](https://platform.openai.com/docs/api-reference/responses/object#responses/object-usage) field

joshua-mo-143:
Fixes #503 
Fixes #400 
Fixes #227 
Fixes RIG-780

A list of breaking changes:
- Added `call_id` as an Option<String> in all core tool result/call type variants to enable usage with OpenAI. If the provider doesn't specifically need it, it can be set as None.

To-do:
- ~~Streaming~~
- ~~Streaming with tools~~


rig-core/src/providers/openai/responses_api/mod.rs
```diff
@@ -0,0 +1,1087 @@
...
+    /// The maximum number of output tokens.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub max_output_tokens: Option<u64>,
+    /// Toggle to true for streaming responses.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub stream: Option<bool>,
+    /// The temperature. Set higher (up to a max of 1.0) for more creative responses.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub temperature: Option<f64>,
+    // TODO: Fix this before opening a PR!
```
0xMochan, line 47: Seems like a lingering TODO. we don't really leverage `tool_choice` in any of our abstractions but it came up in my thoughts when considering some of the problems I ran into #424.

Probably should either defer to a future PR or something but wanted to highlight it.

rig-core/src/providers/openai/responses_api/mod.rs
```diff
@@ -0,0 +1,1087 @@
...
+    // tool_choice: Option<T>,
+    /// The tools you want to use. Currently this is limited to functions, but will be expanded on in future.
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub tools: Vec<ResponsesToolDefinition>,
+    /// Additional parameters
+    #[serde(flatten)]
+    pub addtl_params: AddtlParams,
```
0xMochan: Is there a reason why this is shortened as `addtl` and not `additional_parameters` like we use elsewhere?

rig-core/src/providers/openai/responses_api/mod.rs
```diff
@@ -0,0 +1,1087 @@
...
+    pub model: String,
+    /// Token usage
+    pub usage: Option<ResponsesUsage>,
+    /// The model output (messages, etc will go here)
+    pub output: Vec<Output>,
+    /// Tools
+    pub tools: Vec<ResponsesToolDefinition>,
+    /// Additional parameters
+    #[serde(flatten)]
+    pub addtl_params: AddtlParams,
```
0xMochan: ditto from earlier

rig-core/src/providers/openai/responses_api/mod.rs
```diff
@@ -0,0 +1,1087 @@
...
+                                ..
+                            }) => Ok::<_, message::MessageError>(Message::ToolResult {
+                                tool_call_id: call_id.expect("The tool call ID should exist"),
+                                output: {
+                                    let res = content.first();
+                                    match res {
+                                        completion::message::ToolResultContent::Text(Text {
+                                            text,
+                                        }) => text,
+                                        _ => return  Err(MessageError::ConversionError("This API only currently supports text tool results".into()))
```
0xMochan, line 1004: The responses api also supports image output right? we don't really have our typing setup to make that easy to handle but I'm curious if you put any thought into that.

rig-core/examples/agent_with_tools.rs
```diff
@@ -26,6 +26,8 @@ impl Tool for Adder {
     type Args = OperationArgs;
     type Output = i32;
 
+    // NOTE: Use of `additionalProperties` is required if you're using OpenAI's strict mode
+    // which is on by default. Otherwise, you don't need to include it.
```
0xMochan: Since this is added here, did you test the `#[tool]` macro with it adding these openai specific things for tool definitions.

I tested the `rig-core-derive` example `async_tool` and got the following error:
```rs
thread 'main' panicked at rig-core/rig-core-derive/examples/rig_tool/async_tool.rs:51:64:
called `Result::unwrap()` on an `Err` value: CompletionError(ProviderError("{\n  \"error\": {\n    \"message\": \"Invalid schema for function 'async_operation': In context=(), 'additionalProperties' is required to be supplied and to be false.\",\n    \"type\": \"invalid_request_error\",\n    \"param\": \"tools[0].parameters\",\n    \"code\": \"invalid_function_parameters\"\n  }\n}"))
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

0xMochan (changes requested):
Couple of remarks but this looks fantastic! the responses api looked daunting but I am curious how we can leverage more of it's feature-set beyond the completion compatibility layer. the api allows you to have stateful messages (server-side message history) alongside some other things which would be neat to have a follow up PR for more examples.

- The `#[rig_tool]` macro is borked due to the `additionalParameters` being required (ofc OpenAI still makes odd API design choises).
- Since this completely changes the `OpenAI` client, there _will_ be people's code breaking if they were using the older completion endpoint. I see that is still preserved but an example that showcases how to use the legacy client with other providers would be helpful
  - Ofc the built-in providers have "no-changes" but i'm sure someone will be using legacy clients.
- Couple of other remarks on todos.

rig-core/src/providers/openai/responses_api/mod.rs
joshua-mo-143: It does, yeah. However the original goal was to just initially get parity with Completions as I don't really want to replace multiple APIs at once unless we have to

rig-core/examples/agent_with_tools.rs
joshua-mo-143: I have a feeling we might need to do some janky `json_utils::merge` type stuff with this before sending the request to OpenAI. Does not look like the macro is 100% compatible.

joshua-mo-143: magically fixed :sunglasses:

rig-core/src/providers/openai/responses_api/mod.rs
joshua-mo-143: We can do this in a future PR I think. Should be quite easy

0xMochan (changes requested):
Still getting a
```rs
thread 'main' panicked at rig-core/rig-core-derive/examples/rig_tool/async_tool.rs:51:64:
called `Result::unwrap()` on an `Err` value: CompletionError(ProviderError("{\n  \"error\": {\n    \"message\": \"Invalid schema for function 'async_operation': In context=(), 'required' is required to be supplied and to be an array including every key in properties. Missing 'delay_ms'.\",\n    \"type\": \"invalid_request_error\",\n    \"param\": \"tools[0].parameters\",\n    \"code\": \"invalid_function_parameters\"\n  }\n}"))
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
on `cargo run --example async_tool`

joshua-mo-143:
> Still getting a
> 
> ```rust
> thread 'main' panicked at rig-core/rig-core-derive/examples/rig_tool/async_tool.rs:51:64:
> called `Result::unwrap()` on an `Err` value: CompletionError(ProviderError("{\n  \"error\": {\n    \"message\": \"Invalid schema for function 'async_operation': In context=(), 'required' is required to be supplied and to be an array including every key in properties. Missing 'delay_ms'.\",\n    \"type\": \"invalid_request_error\",\n    \"param\": \"tools[0].parameters\",\n    \"code\": \"invalid_function_parameters\"\n  }\n}"))
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```
> 
> on `cargo run --example async_tool`

Ended up just adding a `required` field to the tool macro. I think this will probably be the best way forward as otherwise there may be a significant amount of code duplication.

0xMochan:
> Since this completely changes the OpenAI client, there will be people's code breaking if they were using the older completion endpoint. I see that is still preserved but an example that showcases how to use the legacy client with other providers would be helpful
>   - Ofc the built-in providers have "no-changes" but i'm sure someone will be using legacy clients.

was this addressed? it's fairly common to use the `OpenAI` client with custom providers. an example that showcases the legacy client would be suitable.

Otherwise, i think i can approve.

joshua-mo-143:
> > Since this completely changes the OpenAI client, there will be people's code breaking if they were using the older completion endpoint. I see that is still preserved but an example that showcases how to use the legacy client with other providers would be helpful
> > 
> > * Ofc the built-in providers have "no-changes" but i'm sure someone will be using legacy clients.
> 
> was this addressed? it's fairly common to use the `OpenAI` client with custom providers. an example that showcases the legacy client would be suitable.
> 
> Otherwise, i think i can approve.

I'll add a relevant example and make sure it compiles and passes CI then we'll get this merged in.

PR #511, opened 2025-06-17
feat(rig-754): support custom client configurations

### LINKED ISSUE #467: feat: support custom HTTP client configuration for proxy and advanced network settings

### Motivation

The current OpenAI client implementation completely encapsulates HTTP client creation internally, preventing users from customizing HTTP client configurations, particularly proxy settings. In certain use cases, users need to:

1. **Corporate Network Environments**: In corporate networks, external API access typically requires routing through proxy servers
2. **Geographic Restrictions**: Some regions may require proxy usage to access OpenAI API
3. **Custom Network Configurations**: Users may need to customize timeout settings, connection pool sizes, TLS configurations, etc.
4. **Network Monitoring and Debugging**: During development, users may need to route traffic through proxy tools (like Burp Suite) for monitoring and debugging HTTP requests
5. **Load Balancing**: In high-concurrency scenarios, users may need to configure multiple proxies for load distribution

The current implementation limits user flexibility in these scenarios.

### Proposal

I suggest implementing one of the following approaches to support custom HTTP client configuration:

#### Option 1: Expose HTTP Client Field

```rust
impl Client {
    /// Get the underlying HTTP client for custom configuration
    pub fn http_client(&self) -> &reqwest::Client {
        &self.http_client
    }
    
    /// Set a custom HTTP client
    pub fn with_http_client(mut self, client: reqwest::Client) -> Self {
        self.http_client = client;
        self
    }
}
```

#### Option 2: Support Custom HTTP Client Injection

```rust
impl Client {
    /// Create an OpenAI client with a custom HTTP client
    pub fn with_custom_client(http_client: reqwest::Client, base_url: &str) -> Self {
        Self {
            base_url: base_url.to_string(),
            http_client,
        }
    }
    
    /// Create a client from custom HTTP client and API key
    pub fn from_custom_client(api_key: &str, http_client: reqwest::Client) -> Self {
        Self::with_custom_client(http_client, OPENAI_API_BASE_URL)
    }
}
```

Usage example:
```rust
use reqwest::{Client as ReqwestClient, Proxy};

// Create HTTP client with proxy
let proxy = Proxy::http("http://proxy.example.com:8080")?
    .basic_auth("username", "password");

let http_client = ReqwestClient::builder()
    .proxy(proxy)
    .timeout(Duration::from_secs(30))
    .build()?;

// Create OpenAI client with custom HTTP client
let openai_client = Client::from_custom_client("your-api-key", http_client);
```

**I recommend Option 2** because it:
- Maintains backward compatibility of the API
- Provides users with complete control
- Simple implementation that doesn't break existing encapsulation

#### Alternative 1: Configuration Struct Approach
```rust
#[derive(Default)]
pub struct ClientConfig {
    pub proxy: Option<Proxy>,
    pub timeout: Option<Duration>,
    pub user_agent: Option<String>,
}

impl Client {
    pub fn with_config(api_key: &str, config: ClientConfig) -> Result<Self, Error> {
        // Create HTTP client based on configuration
    }
}
```

**Drawbacks**: Requires adding fields for every possible configuration option, lacks flexibility, and cannot support all reqwest configuration options.

#### Alternative 2: Builder Pattern
```rust
pub struct ClientBuilder {
    api_key: String,
    base_url: String,
    proxy: Option<Proxy>,
    timeout: Option<Duration>,
}

impl ClientBuilder {
    pub fn proxy(mut self, proxy: Proxy) -> Self { /* ... */ }
    pub fn timeout(mut self, timeout: Duration) -> Self { /* ... */ }
    pub fn build(self) -> Result<Client, Error> { /* ... */ }
}
```

**Drawbacks**: Requires duplicating reqwest's existing Builder functionality, high maintenance cost, and still cannot cover all configuration options.

#### Alternative 3: Environment Variable Support Only
Relying solely on reqwest's environment variable support (`http_proxy`, `https_proxy`, etc.).

**Drawbacks**:
- Not flexible enough, cannot use different proxy configurations within the same program
- Does not support programmatic proxy rotation
- Does not support proxy scenarios requiring authentication
- Cannot customize other HTTP client configurations

**Why the proposed solution was chosen**:
The proposed Option 2 is optimal because it fully leverages reqwest's existing powerful configuration capabilities, provides users with maximum flexibility, while keeping the API clean and backward compatible. Users can configure any reqwest-supported functionality according to their needs, including proxies, timeouts, TLS, connection pools, etc., without requiring library maintainers to provide separate APIs for each configuration option.

joshua-mo-143:
Fixes #467 
Fixes RIG-754

Additionally adds an upgrade to the Debug implementation for every client such that the API key will never be exposed when using Debug. This is a highly important implementation detail as the user should not need to work around this if/when implementing observability systems that additionally cover Rig.

### DIFF
--- a/rig-core/src/providers/anthropic/client.rs
+++ b/rig-core/src/providers/anthropic/client.rs
@@ -1,5 +1,4 @@
 //! Anthropic client api implementation
-
 use super::completion::{CompletionModel, ANTHROPIC_VERSION_LATEST};
 use crate::client::{impl_conversion_traits, CompletionClient, ProviderClient};
 
@@ -68,10 +67,27 @@ impl<'a> ClientBuilder<'a> {
     }
 }
 
-#[derive(Clone, Debug)]
+#[derive(Clone)]
 pub struct Client {
+    /// The base URL
     base_url: String,
+    /// The API key
+    api_key: String,
+    /// The underlying HTTP client
     http_client: reqwest::Client,
+    /// Default headers that will be automatically added to any given request with this client (API key, Anthropic Version and any betas that have been added)
+    default_headers: reqwest::header::HeaderMap,
+}
+
+impl std::fmt::Debug for Client {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("Client")
+            .field("base_url", &self.base_url)
+            .field("http_client", &self.http_client)
+            .field("api_key", &"<REDACTED>")
+            .field("default_headers", &self.default_headers)
+            .finish()
+    }
 }
 
 impl Client {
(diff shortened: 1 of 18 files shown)


joshua-mo-143:
@schwifty-brows @mztlive PR opened - please review at your discretion 🙂 

I also noticed that specific implementation details of certain model providers are inconsistent, but that's a problem for another issue at the moment and can be solved later.

schwifty-brows:
hey @joshua-mo-143 checking out now

schwifty-brows (approved):
lgtm 

this will allow us to inject a custom http client with setting specific to our relative environments

schwifty-brows:
when can we get a new version?

joshua-mo-143:
> when can we get a new version?

Next week according to the release cycle, assuming no blockers (there normally aren't any, but just in case there are).

PR #512, opened 2025-06-17
fix(rig-782): tool calls in OpenAI stream delta need to be option

### LINKED ISSUE #509: bug: tool_calls in DeltaResponse struct need to be Option?
When debugging with OpenAI's LLM, I found that sometimes their SSE response looks like this:

Image

The tool_call field is filled with null.

and then I asked OpenAI's chat result as follows:

Image

joshua-mo-143:
Fixes #509 
Fixes RIG-782

The `send_compatible_streaming_request` function could really use some breaking down for visibility, but we can do that in another PR.


joshua-mo-143:
~~Blocked until #508 is merged~~ Unblocked

gold-silver-copper:
Closing this as obsolete.

The linked issue #509 is already closed as completed, and current `main` handles the exact `tool_calls: null` stream delta shape via `json_utils::null_or_vec` on `StreamingDelta.tool_calls`. It also has a focused regression test, `test_streaming_delta_with_null_tool_calls`.

This landed through #1786 / commit `425c96ef`. I verified the current behavior locally with `cargo test -p rig-core test_streaming_delta_with_null_tool_calls --lib`, so this older `Option<Vec<...>>` patch is no longer needed.

PR #514, opened 2025-06-17
feat(rig-784): support thinking/reasoning

### LINKED ISSUE #513: feat: support thinking/reasoning
## Motivation

The motivation for this is pretty obvious, but in case it wasn't: many models are increasingly starting to support reasoning. We should also therefore support this to ensure we stay SOTA.

## Scope

Add it as an `AssistantContent` enum.

joshua-mo-143:
Fixes #513 
Fixes RIG-784


joshua-mo-143:
Closed for now - will merge to alternative branch

PR #515, opened 2025-06-17
fix(rig-783): tool call example doesn't work with Gemini and OpenRouter

### LINKED ISSUE #507: bug: Only the first MCP tool is recognized by Gemini agent when adding multiple tools
Hi, I'm encountering an issue when trying to add multiple MCP tools to a Gemini agent using the `rig` library.

I’m using the following code to list all available tools from a `@modelcontextprotocol/server-filesystem` (MCP) server and attach them to a Gemini agent:

```rust
#[tokio::main]
async fn main() -> Result<(), ()>
{
    let api_key = std::env::var( "GOOGLE_API_KEY" ).map_err( | _ | () )?;

    let client = rig::providers::gemini::Client::new( &api_key );

    let mut agent_builder = client.agent( "gemini-2.5-pro-preview-05-06" );

    let mcp_client = ClientBuilder::new( 
        ClientStdioTransport::new( 
            "npx", 
            &[ "-y", "@modelcontextprotocol/server-filesystem", "/tmp" ]
        ).map_err( | _ | () )? 
    ).build();

    mcp_client.open().await.map_err( | _ | () )?;

    let tools = mcp_client.list_tools( None, None ).await.map_err( | _ | () )?.tools;

    agent_builder = tools
    .into_iter()
    .fold(agent_builder, |builder, tool| {
        builder.mcp_tool(tool, mcp_client.clone().into())
    });

    let agent = agent_builder.build();

    let mut stream = agent.stream_chat( "List /tmp directory", vec![] ).await.unwrap();

    while let Some(chunk) = stream.next().await 
    {
        match chunk {
            Ok(AssistantContent::Text(text)) => {
                println!("{}", text.text);
            }
            Ok(AssistantContent::ToolCall(tool_call)) => {
                let res = agent
                    .tools
                    .call(
                        &tool_call.function.name,
                        tool_call.function.arguments.to_string(),
                    )
                    .await
                    .map_err(|e| std::io::Error::other(e.to_string())).unwrap();

                println!("\nTool result:\n{res}");
            }
            Err(e) => {
                eprintln!("Error: {e}");
                break;
            }
        }
    }
    
    Ok( () )
}
```

However, when running the program, **only the first tool seems to be recognized and callable** by the Gemini model. For example, even when the `list_directory` tool is available and correctly configured, the model still replies:

> I am sorry, I cannot list directory contents. I can only read files.

To investigate further, I modified the code to skip the first five tools:

```rust
agent_builder = tools
    .into_iter()
    .skip(5) // skip the first five tools
    .fold(agent_builder, |builder, tool| {
        builder.mcp_tool(tool, mcp_client.clone().into())
    });
```

Now, the model **correctly uses `list_directory`** and returns the expected result:

```
Tool result:  
[DIR] .ICE-unix  
[FILE] .X0-lock  
[DIR] .X11-unix  
...
```

The MCP server returns the following tools (names only for brevity):

1. read_file  
2. read_multiple_files  
3. write_file  
4. edit_file  
5. create_directory  
6. list_directory  
7. directory_tree  
8. move_file  
9. search_files  
10. get_file_info  
11. list_allowed_directories  

It seems that the Gemini agent is **only registering the first tool added**, or only considering the first one internally. Skipping the first N tools makes the N+1-th tool work, suggesting that the loop is not accumulating them properly — even though `.fold()` is used.

I’m using:

```
[dependencies]
tokio = { version = "1.45.1", features = ["full"] }
tokio-stream = "0.1.17"
rig-core = { version = "0.13.0", features = [ "mcp" ] }
mcp-core = "0.1.50"
serde = { version = "1.0", features = ["derive"] }
serde_json = { version = "1.0.140" }
```

Let me know if I can provide more info or help debug this further.

Thanks!

### LINKED ISSUE #510: bug: Tool call example doesn't work with Gemini and OpenRouter Providers

## Bug Report
The tool call example (and static tool calls in general) does not work with Gemini and OpenRouter providers. 

## Reproduction
(See [example](https://github.com/0xPlaygrounds/rig/blob/main/rig-core/examples/agent_with_tools.rs) )
Gemini:
```rust
// same dependency and decl as the example 
#[tokio::main]
    tracing_subscriber::fmt()
        .with_max_level(tracing::Level::DEBUG)
        .with_target(false)
        .init();

    // Create Gemini/OpenrRouter client
    let client= providers::gemini::Client::from_env();

    // Create agent with a single context prompt and two tools
    let calculator_agent = client
        .agent(providers::gemini::completion::GEMINI_2_5_FLASH_PREVIEW_05_20)
        .preamble("You are a calculator here to help the user perform arithmetic operations. Use the tools provided to answer the user's question.")
        .max_tokens(1024)
        .tool(Adder)
        .tool(Subtract)
        .build();

    println!("Calculate 2 - 5");

    println!(
        "OpenAI Calculator Agent: {}",
        calculator_agent.prompt("Calculate 2 - 5").await?
    );

    Ok(())
}
```

OpenRouter: 
```rust
#[tokio::main]
    tracing_subscriber::fmt()
        .with_max_level(tracing::Level::DEBUG)
        .with_target(false)
        .init();

    // Create Gemini/OpenrRouter client
    let client= providers::openrouter::Client::from_env();

    // Create agent with a single context prompt and two tools
    let calculator_agent = client
        .agent(providers::openrouter::completion::GEMINI_FLASH_2_0)
        .preamble("You are a calculator here to help the user perform arithmetic operations. Use the tools provided to answer the user's question.")
        .max_tokens(1024)
        .tool(Adder)
        .tool(Subtract)
        .build();

    println!("Calculate 2 - 5");

    println!(
        "OpenAI Calculator Agent: {}",
        calculator_agent.prompt("Calculate 2 - 5").await?
    );

    Ok(())
}
```

## Expected behavior
Tool calls, and response of 3. 

## Actual Output
Gemini:
```bash
2025-06-14T19:54:40.157646Z DEBUG Received response
OpenAI Calculator Agent: I'm sorry, I cannot fulfill this request. I can only add numbers.
```

OpenRouter (`openrouter::completion::GEMINI_FLASH_2_0` used in this case) 
```bash
❯ cargo run
   Compiling recreation v0.1.0 (/home/****/*****/recreation)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.85s
     Running `target/debug/recreation`
Calculate 2 - 5
2025-06-14T20:00:28.442110Z DEBUG starting new connection: https://openrouter.ai/
2025-06-14T20:00:28.443734Z DEBUG connecting to 104.22.48.189:443
2025-06-14T20:00:28.474124Z DEBUG connected to 104.22.48.189:443
2025-06-14T20:00:29.180775Z DEBUG pooling idle connection for ("https", openrouter.ai)
2025-06-14T20:00:29.180935Z  INFO OpenRouter completion token usage: "Prompt tokens: 31 Total tokens: 45"
OpenAI Calculator Agent: ```tool_code
print(2 - 5)
```

## Suspicions 
The failure mode between gemini provider and openrouter provider is different. OpenRouter, when prompted and asked what tools are presented to it, will comment that it cannot see any tools. In both cases, the tool call's print out `[tool-call]` is not seen anywhere. 

Gemini's request seems well formatted:
```json
}
"contents": [
    {
      "parts": [
        {
          "text": "Calculate 2 - 5"
        }
      ],
      "role": "user"
    }
  ],
  "tools": [
    {
      "functionDeclarations": {
        "name": "add",
        "description": "Add x and y together",
        "parameters": {
          "type": "object",
          "properties": {
            "y": {
              "type": "number",
              "description": "The second number to add"
            },
            "x": {
              "type": "number",
              "description": "The first number to add"
            }
          }
        }
      },
      "codeExecution": null
    },
    {
      "functionDeclarations": {
        "name": "subtract",
        "description": "Subtract y from x (i.e.: x - y)",
        "parameters": {
          "type": "object",
          "properties": {
            "x": {
              "type": "number",
              "description": "The number to subtract from"
            },
            "y": {
              "type": "number",
              "description": "The number to subtract"
            }
          }
        }
      },
      "codeExecution": null
    }
  ],
  "toolConfig": null,
  "generationConfig": {
    "maxOutputTokens": 1024
  },
  "safetySettings": null,
  "systemInstruction": {
    "parts": [
      {
        "text": "You are a calculator here to help the user perform arithmetic operations. Use the tools provided to answer the user's question."
      }
    ],
    "role": "model"
  }
}
```

## Screenshots
Image

## Additional context
[Malachid on tool call errors using ollama](https://www.malachid.com/blog/2025-04-06-rig-ollama/).

joshua-mo-143:
Fixes #510 
Fixes RIG-783

### DIFF
--- a/rig-core/src/providers/gemini/completion.rs
+++ b/rig-core/src/providers/gemini/completion.rs
@@ -151,13 +151,7 @@ pub(crate) fn create_request_body(
             .collect::<Result<Vec<_>, _>>()?,
         generation_config: Some(generation_config),
         safety_settings: None,
-        tools: Some(
-            completion_request
-                .tools
-                .into_iter()
-                .map(Tool::try_from)
-                .collect::<Result<Vec<_>, _>>()?,
-        ),
+        tools: Some(Tool::try_from(completion_request.tools)?),
         tool_config: None,
         system_instruction,
     };
@@ -176,11 +170,51 @@ impl TryFrom<completion::ToolDefinition> for Tool {
                 Some(tool.parameters.try_into()?)
             };
         Ok(Self {
-            function_declarations: FunctionDeclaration {
+            function_declarations: OneOrMany::one(FunctionDeclaration {
                 name: tool.name,
                 description: tool.description,
                 parameters,
-            },
+            }),
+            code_execution: None,
+        })
+    }
+}
+
+impl TryFrom<Vec<completion::ToolDefinition>> for Tool {
+    type Error = CompletionError;
+
+    fn try_from(tools: Vec<completion::ToolDefinition>) -> Result<Self, Self::Error> {
+        let mut functions = Vec::new();
+
+        for tool in tools {
+            let parameters =
+                if tool.parameters == serde_json::json!({"type": "object", "properties": {}}) {
+                    None
+                } else {
+                    match tool.parameters.try_into() {
+                        Ok(schema) => Some(schema),
+                        Err(e) => {
+                            let emsg = format!(
+                                "Tool '{}' could not be converted to a schema: {:?}",
+                                tool.name, e,
+                            );
+                            return Err(CompletionError::ProviderError(emsg));
+                        }
+                    }
+                };
+
+            functions.push(FunctionDeclaration {
+                name: tool.name,
+                description: tool.description,
+                parameters,
+            });
+        }
+
+        let function_declarations: OneOrMany<FunctionDeclaration> = OneOrMany::many(functions)
+            .map_err(|x| CompletionError::ProviderError(x.to_string()))?;
+
+        Ok(Self {
+            function_declarations,
             code_execution: None,
         })
     }
@@ -235,7 +269,7 @@ pub mod gemini_api_types {
     // Gemini API Types
     // =================================================================
     use serde::{Deserialize, Serialize};
-    use serde_json::Value;
+    use serde_json::{json, Value};
 
     use crate::{
         completion::CompletionError,
@@ -445,11 +479,7 @@ pub mod gemini_api_types {
                     };
                     Ok(Part::FunctionResponse(FunctionResponse {
                         name: id,
-                        response: Some(serde_json::from_str(&content).map_err(|e| {
-                            message::MessageError::ConversionError(format!(
-                                "Failed to parse tool response: {e}"
-                            ))
-                        })?),
+                        response: Some(json!({ "result": content })),
                     }))
                 }
                 message::UserContent::Image(message::Image {
@@ -581,7 +611,7 @@ pub mod gemini_api_types {
         /// with a maximum length of 63.
         pub name: String,
         /// The function response in JSON object format.
-        pub response: Option<HashMap<String, serde_json::Value>>,
+        pub response: Option<serde_json::Value>,
     }
 
     /// URI based data.
@@ -841,7 +871,7 @@ pub mod gemini_api_types {
     /// The Schema object allows the definition of input and output data types. These types can be objects, but also
     /// primitives and arrays. Represents a select subset of an OpenAPI 3.0 schema object.
     /// From [Gemini API Reference](https://ai.google.dev/api/caching#Schema)
-    #[derive(Debug, Deserialize, Serialize)]
+    #[derive(Debug, Deserialize, Serialize, Clone)]
     pub struct Schema {
         pub r#type: String,
         #[serde(skip_serializing_if = "Option::is_none")]
@@ -934,7 +964,7 @@ pub mod gemini_api_types {
     #[serde(rename_all = "camelCase")]
     pub struct GenerateContentRequest {
         pub contents: Vec<Content>,
-        pub tools: Option<Vec<Tool>>,
+        pub tools: Option<Tool>,
         pub tool_config: Option<ToolConfig>,
         /// Optional. Configuration options for model generation and outputs.
         pub generation_config: Option<GenerationConfig>,
@@ -961,11 +991,11 @@ pub mod gemini_api_types {
     #[derive(Debug, Serialize)]
     #[serde(rename_all = "camelCase")]
     pub struct Tool {
-        pub function_declarations: FunctionDeclaration,
+        pub function_declarations: OneOrMany<FunctionDeclaration>,
         pub code_execution: Option<CodeExecution>,
     }
 
-    #[derive(Debug, Serialize)]
+    #[derive(Debug, Serialize, Clone)]
     #[serde(rename_all = "camelCase")]
     pub struct FunctionDeclaration {
         pub name: String,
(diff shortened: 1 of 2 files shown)


joshua-mo-143:
cc @winston-bosan PR opened, feel free to review at your discretion :slightly_smiling_face:

winston-bosan:
Can confirm, example tool use and custom tools for gemini provider is working!
`rig-core = { git = "https://github.com/0xPlaygrounds/rig.git", branch = "fix-rig-783-tool-call-example-doesnt-work-gemini-and-openrouter" , features = ["all", "mcp"] }`

rig-core/src/providers/gemini/completion.rs
```diff
@@ -445,11 +479,7 @@ pub mod gemini_api_types {
                     };
                     Ok(Part::FunctionResponse(FunctionResponse {
                         name: id,
-                        response: Some(serde_json::from_str(&content).map_err(|e| {
-                            message::MessageError::ConversionError(format!(
-                                "Failed to parse tool response: {e}"
-                            ))
-                        })?),
+                        response: Some(json!({ "result": content })),
```
winston-bosan: *NVM it is just vertex-ai v. gemini-api thing. They have slightly different FunctionResponsePart nesting scheme and I was expecting them to be the same.  
~~This barely qualifies as a comment and I don't usually complain when things work. But this is subtlely different from google's function-calling impl and example.~~
~~see [google on function calling](https://ai.google.dev/gemini-api/docs/function-calling?example=meeting#javascript_5)~~
~~The `{ result }` wrap is slightly suspicious. This would mean that we have a response object wrapping a result like this:~~

```js
// Create a function response part
const function_response_part = {
  name: tool_call.name,
  response: { result }
}
```

Looking like:
```js
 {
  name: some_name,
  response: { 
    result: {
        brightness: 25, 
        ...
      }
   }
}
```

This is at odds with the [google cloud ai doc](https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/function-calling). Which doesn't have the additional layer of nesting: 
```json
{
  "role": "user",
  "parts": [
    {
      "functionResponse": {
        "name": "get_current_weather",
        "response": {
          "temperature": 20,
          "unit": "C"
        }
      }
    }
  ]
}
```

winston-bosan: Any reason why we don't have 
`Some(json!({ "result": serde_json::from_str(&content)? }))`
instead of
`Some(json!({ "result": content }))`

joshua-mo-143: Ah right yes - sorry, should've caught this one! Adding it now :slightly_smiling_face:

bajie-git:
This modification seems to introduce new issues,
According to the Gemini documentation, the type of "tools" is array. This modification will cause the type of "tools" to become object
[gemini doc](https://ai.google.dev/gemini-api/docs/function-calling)

This can lead to type errors on some intermediary APIs

Perhaps it would be more reasonable to modify it to ```tools: Some(vec![Tool::try_from(completion_request.tools)?])```

PR #526, opened 2025-06-22
fix: Retain multi-turn tool call results in case of response error

wonrax:
In a multi-turn loop, the LLM might send an empty message with no tool call to indicate it wants to stop. This causes the `Completion::completion` method to return an error, exiting the loop early without pushing the previous tool results into the chat history. If this incomplete chat history is used subsequently, OpenAI returns this error:

"An assistant message with 'tool_calls' must be followed by tool messages responding to each 'tool_call_id'. The following tool_call_ids did not have response messages: call_0EkD..."

This commit fixes the issue by updating the chat history with the previous tool results before performing the completion request.

Without modifying the library, an alternative fix would be changing the prompt to require the LLM to always respond with a valid string.


joshua-mo-143:
Hi, do you know if this is reproducible? Seems a bit odd that they'd just send an empty message

wonrax:
Yeah, I actually encountered this problem myself. That's why I created this PR.

In our use case, the agent is instructed to communicate with users through other means such as sending Discord messages using tools. That's very likely why the LLM knows not to output anything to stop the agent loop.

image

wonrax:
As mentioned in the commit message, the fix for this specific completion error is instructing the LLM to always output something other than an empty message.

This PR fixes the inconsistency for any error thrown while we're requesting chat completion in case the user wants to continue using the chat history.

joshua-mo-143:
If you pull from main the clippy lint should fix itself

joshua-mo-143 (approved):
lgtm

PR #539, opened 2025-06-26
Migrate all crates to Rust 2024

quangIO:


quangIO:
Upgrades everything to Edition 2024 (why not right?)

joshua-mo-143:
Thanks for your hard work @quangIO - will merge now :smile:

PR #541, opened 2025-06-27
feat(ollama): add support for OLLAMA_API_BASE_URL environment var

rmsc:
This allows configuring the ollama url via the OLLAMA_API_BASE_URL environment variable.

### DIFF
--- a/rig-core/src/providers/ollama.rs
+++ b/rig-core/src/providers/ollama.rs
@@ -104,7 +104,8 @@ impl ProviderClient for Client {
     where
         Self: Sized,
     {
-        Client::default()
+        let api_base = std::env::var("OLLAMA_API_BASE_URL").expect("OLLAMA_API_BASE_URL not set");
+        Self::from_url(&api_base)
     }
 }
 


joshua-mo-143:
Hi, it looks like there were a number of clippy errors that have been fixed from a recent merge - you just need to git pull the changes into your branch and I believe that should make this pass CI!

rig-core/src/providers/ollama.rs
```diff
@@ -104,7 +104,8 @@ impl ProviderClient for Client {
     where
         Self: Sized,
     {
-        Client::default()
+        let api_base = std::env::var("OLLAMA_API_BASE").expect("OLLAMA_API_BASE not set");
```
joshua-mo-143: Should be `OLLAMA_API_BASE_URL` for clarity.

joshua-mo-143 (changes requested):
Just one change required then lgtm :slightly_smiling_face:

rig-core/src/providers/ollama.rs
rmsc: You're absolutely right, somehow missed the `const`

PR #545, opened 2025-06-30
fix(rig-794): parse openAI SSE response error

### LINKED ISSUE #534: bug: parse openai SSE response error

## Bug Report

Som openai-compatible API returns response in this format, `data:[\n]{}` not `data:[space]{}`

```
data:
 {"id":"chatcmpl-BlugotBWaIJzVEj6kpJwXonBuZ5Yq","object":"chat.completion.chunk","created":1750759030,"model":"gpt-4.1-2025-04-14","choices":[{"delta":{"content":"可以"},"index":0,"finish_reason":null,"logprobs":null}],"content":"你好！有什么我可以","lastOne":false}
```

It doesn't work with this format

https://github.com/0xPlaygrounds/rig/blob/main/rig-core/src/providers/openai/streaming.rs#L128

joshua-mo-143:
Fixes #534 
Fixes RIG-794

### DIFF
--- a/rig-core/src/providers/openai/streaming.rs
+++ b/rig-core/src/providers/openai/streaming.rs
@@ -125,9 +125,11 @@ pub async fn send_compatible_streaming_request(
                 }
                 // Otherwise full data line
                 else {
-                    let Some(data) = line.strip_prefix("data: ") else {
+                    let Some(data) = line.strip_prefix("data:") else {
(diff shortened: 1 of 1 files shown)


joshua-mo-143:
@EddieIvan01 This should fix it I believe - please review at your discretion

PR #555, opened 2025-07-03
feat: add `.extended_details` to `PromptRequest`

### LINKED ISSUE #433: feat: extended details from prompting

## Feature Request
Add the ability to gather used tokens and other per-prompt details from `.prompt` via the `PromptBuilder`.

### Motivation
While `.completion` provides the full completion request, it's still fairly useful to do some accounting with the higher level API, especially considering the amount of potential tokens being used via multi-turn. This would allow looping to accumulate tokens usage and present them seamlessly to the user w/o much api hassle.

### Proposal
```rs
agent
  .prompt(...)
  .extended_details()  // or some better name
```

Add a boolean (or type-state) to `PromptBuilder`. Come up with an `PromptResponse` that accumulates tokens used and other useful details about the raw responses made.

### Alternatives
Separate method alongside `.completion`.

0xMochan:
Fixes #433

Leverages typestate back on `PromptRequest` since the `.send` needs to return a struct instead of a string. otherwise, this allows you to easily encapsulate aggregated token usage with multi-turn usage.

## Testing
`cargo run --example multi_turn_agent_extended`

### DIFF
--- a/rig-core/examples/multi_turn_agent_extended.rs
+++ b/rig-core/examples/multi_turn_agent_extended.rs
@@ -41,7 +41,7 @@ async fn main() -> anyhow::Result<()> {
         .extended_details()
         .await?;
 
-    println!("\n\nOpenAI Calculator Agent: {:?}", result);
+    println!("\n\nOpenAI Calculator Agent: {result:?}");
 
     // Prompt the agent again and print the response
     let result = agent
@@ -50,7 +50,7 @@ async fn main() -> anyhow::Result<()> {
         .extended_details()
         .await?;
 
-    println!("\n\nOpenAI Calculator Agent: {:?}", result);
+    println!("\n\nOpenAI Calculator Agent: {result:?}");
 
     Ok(())
 }
--- a/rig-core/examples/multi_turn_agent_extended.rs
+++ b/rig-core/examples/multi_turn_agent_extended.rs
@@ -137,6 +137,7 @@ impl Tool for Subtract {
     }
 }
 
+#[derive(Deserialize, Serialize)]
 struct Multiply;
 
 impl Tool for Multiply {
@@ -172,6 +173,7 @@ impl Tool for Multiply {
     }
 }
 
+#[derive(Deserialize, Serialize)]
 struct Divide;
 
 impl Tool for Divide {
--- a/rig-core/examples/multi_turn_agent_extended.rs
+++ b/rig-core/examples/multi_turn_agent_extended.rs
@@ -1,0 +1,208 @@
+use rig::prelude::*;
+use rig::{
+    completion::{Prompt, ToolDefinition},
+    providers::anthropic,
+    tool::Tool,
+};
+use serde::{Deserialize, Serialize};
+use serde_json::json;
+
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    tracing_subscriber::fmt()
+        .with_max_level(tracing::Level::DEBUG)
+        .with_target(false)
+        .init();
+
+    // Create OpenAI client
+    let openai_client = anthropic::Client::from_env();
+
+    // Create RAG agent with a single context prompt and a dynamic tool source
+    let agent = openai_client
+        .agent(anthropic::CLAUDE_3_5_SONNET)
+        .preamble(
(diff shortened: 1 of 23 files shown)


rig-core/examples/multi_turn_agent_extended.rs
```diff
@@ -0,0 +1,208 @@
+use rig::prelude::*;
```
joshua-mo-143, line 1: We definitely need to stop adding so many of these examples :joy: one day we'll consolidate them all

rig-core/src/agent/prompt_request.rs
```diff
@@ -1,19 +1,26 @@
...
+pub struct Standard;
+pub struct Extended;
+
+impl PromptType for Standard {}
+impl PromptType for Extended {}
+
 /// A builder for creating prompt requests with customizable options.
 /// Uses generics to track which options have been set during the build process.
-pub struct PromptRequest<'a, M: CompletionModel> {
+pub struct PromptRequest<'a, S: PromptType, M: CompletionModel> {
```
joshua-mo-143, line 29: adding trait bounds to a struct unnecessarily is an anti pattern (and if the field isn't pub, it's impossible to instantiate it from the user side anyway)

rig-core/examples/multi_turn_agent_extended.rs
```diff
@@ -0,0 +1,208 @@
...
+        }))
+        .expect("Tool Definition")
+    }
+
+    async fn call(&self, args: Self::Args) -> Result<Self::Output, Self::Error> {
+        let result = args.x - args.y;
+        Ok(result)
+    }
+}
+
```
joshua-mo-143, line 139: Missing `#[derive(Deserialize, Serialize)]` for consistency across tools

rig-core/examples/multi_turn_agent_extended.rs
```diff
@@ -0,0 +1,208 @@
...
+        .expect("Tool Definition")
+    }
+
+    async fn call(&self, args: Self::Args) -> Result<Self::Output, Self::Error> {
+        let result = args.x * args.y;
+        Ok(result)
+    }
+}
+
+struct Divide;
```
joshua-mo-143, line 177: Missing `#[derive(Deserialize, Serialize)]` for consistency across tools

joshua-mo-143 (changes requested):
Some minor changes to be made but overall lgtm

rig-core/examples/multi_turn_agent_extended.rs
joshua-mo-143: sorry just realised this is pointing entirely at the wrong line - should be L140 where `struct Multiply;` is

0xMochan: honestly, i copied most of this from `multi_turn_agent`

rig-core/src/agent/prompt_request.rs
0xMochan: This is the typestate pattern and is needed to be able to return 2 different types (on the `.send` method) based on whether `.extended_details` is used in the builder. The alternative is setting a boolean in the struct and using if statements (like we do for the other builder options) but this isn't possible bc the type of the return needs to be different if the option exists (returning structured data versus just a String).

0xMochan:
btw, another thought i had. some providers call it `input_tokens` and `output_tokens` while others do `prompt_tokens` and `completion_tokens`. which is the better naming.

i'm wondering if we should use `input`/`output` since it's more applicable.

rig-core/examples/multi_turn_agent_extended.rs
joshua-mo-143: Ok no worries. At some point we will probably need to do a code audit for consistency anyway

joshua-mo-143:
> btw, another thought i had. some providers call it `input_tokens` and `output_tokens` while others do `prompt_tokens` and `completion_tokens`. which is the better naming.
> 
> i'm wondering if we should use `input`/`output` since it's more applicable.

I think input/output probably makes more sense from a semantic point of view in my opinion as it's absolutely unambiguous which each one is while prompt/completion does leave a little bit of room. I think. Although at this point I think most people who are using Rig will find the terms to be interchangeable

PR #560, opened 2025-07-05
feat: Add `max_tokens` method to `ExtractorBuilder`

snowmead:
Adds `max_tokens` method to `ExtractorBuilder` for Anthropic models that require this parameter.


joshua-mo-143:
lgtm - thank you for contributing!

PR #562, opened 2025-07-05
docs(rig-803): improve documentation for multi-turn

### LINKED ISSUE #561: docs: improve documentation for multi-turn
There is some missing documentation on multi-turn which makes it difficult to understand/use as many users are running into a DX papercut with `MaxDepthError` (and not understanding what the error actually means).

## Solution
Write the relevant documentation.

joshua-mo-143:
Fixes #561 
Fixes RIG-803


ProHaller (approved):
That clarifies the relationship between MaxDepthError and .multi_turn() clearly! Nice.

PR #563, opened 2025-07-05
feat: support inserting documents as a trait

### LINKED ISSUE #524: feat: support inserting documents as a trait

## Feature Request

Currently, every vector store provider integration that we have has an `insert_documents` method. It would be a great idea to formalise this as a trait. By doing so:
- We can reference this method without the compiler needing to know what type it is
- We can ensure that this behaviour is guaranteed across vector stores.

This would also be a good chance to implement a new, different type of `insert_docs` method that also takes a user-defined ID generation strategy (although perhaps out of initial scope for the first PR).

### Motivation

Better UX, better type reusability.

### Proposal

Move all `insert_document` methods under a new `InsertDocuments` trait.

rajiknows:
fixes #524 

1. added a trait `InsertDocument` in rig-core/vector-store.
2. updated all the `insert_document` methods from db to use the new trait


rig-core/src/vector_store/mod.rs
```diff
@@ -29,6 +32,17 @@ pub enum VectorStoreError {
     ExternalAPIError(StatusCode, String),
 }
 
+/// Trait for inserting documents into a vector store.
+#[async_trait]
```
joshua-mo-143: I don't believe the use of `#[async_trait]` here to be necessary. There's definitely a way to do this without needing to add a new crate (hint: look at the `VectorStoreIndex` return type!)

rig-milvus/src/lib.rs
```diff
@@ -132,9 +133,11 @@ impl<M: EmbeddingModel> MilvusVectorStore<M> {
             output_fields: vec!["id", "distance"],
         }
     }
+}
 
-    /// Insert vectors (with metadata) into your Milvus instance.
-    pub async fn insert_documents<Doc: Serialize + Embed + Send>(
+#[async_trait]
```
joshua-mo-143: Unnecessary once the `async_trait` bound has been removed from the core trait impl

rig-postgres/src/lib.rs
```diff
@@ -109,8 +110,11 @@ impl<Model: EmbeddingModel> PostgresVectorStore<Model> {
             document, document, self.distance_function, self.documents_table
         )
     }
+}
 
-    pub async fn insert_documents<Doc: Serialize + Embed + Send>(
+#[async_trait]
```
joshua-mo-143: Unnecessary once the async_trait bound has been removed from the core trait impl

rig-qdrant/src/lib.rs
```diff
@@ -55,8 +56,11 @@ impl<M: EmbeddingModel> QdrantVectorStore<M> {
         params.limit = Some(limit as u64);
         params
     }
+}
 
-    pub async fn insert_documents<Doc: Serialize + Embed + Send>(
+#[async_trait]
```
joshua-mo-143: Unnecessary once the async_trait bound has been removed from the core trait impl

rig-scylladb/src/lib.rs
```diff
@@ -167,8 +168,29 @@ impl<M: EmbeddingModel> ScyllaDbVectorStore<M> {
...
+    }
+
+    /// Generate query vector from text
+    async fn generate_query_vector(&self, query: &str) -> Result<Vec<f32>, VectorStoreError> {
+        let embedding = self.model.embed_text(query).await?;
+        Ok(embedding.vec.iter().map(|&x| x as f32).collect())
+    }
+}
+
+#[async_trait]
```
joshua-mo-143: Unnecessary once the async_trait bound has been removed from the core trait impl

rig-surrealdb/src/lib.rs
```diff
@@ -68,6 +69,40 @@ impl SearchResult {
     }
 }
 
+#[async_trait]
```
joshua-mo-143: Unnecessary once the async_trait bound has been removed from the core trait impl

rig-surrealdb/src/lib.rs
```diff
@@ -68,6 +69,40 @@ impl SearchResult {
...
+impl<Model: EmbeddingModel + Send + Sync, C: Connection + Send + Sync> InsertDocuments
```
joshua-mo-143: Please use a `where` clause for multiple generics. It makes it much easier from a readability standpoint to check generics trait bounds at a glance.

joshua-mo-143 (changes requested):
Few changes need to be made, but otherwise this looks pretty solid!

re: async_trait:
The reason why I have asked for it to not be added in is that now that we're in 1.75 Rust and beyond, the crate is primarily for dealing with `dyn`  async traits. I believe at this point that we don't necessarily need `async_trait` yet as we're still in the first implementation - if we want to add it in, it should be something that we do in a future PR as it means that we can decouple this feature from needing in a new crate.

rig-core/src/vector_store/mod.rs
rajiknows: sorry for not looking that out , should've checked VectorStoreIndex , jumped straight to solution :(

joshua-mo-143: No worries. The whole point of a review is to help you get better as an engineer, so please don't take this as a personal attack :slightly_smiling_face:

joshua-mo-143 (approved):
lgtm

PR #567, opened 2025-07-08
feat(rig-798): make a `rig-wasm` crate

### LINKED ISSUE #550: feat: make a rig-wasm crate

## Feature Request

We have floated around the idea of compiling Rig to WASM internally for a while now, and it's pretty obvious given that we have added a test specifically for this.

Now is a good time to formalise this as there is some scope to add a Rig crate that is compatible with wasm-bindgen and generates JS code that can be used as if it was a JS library (albeit the user will need to load the library since it's WASM).

This ticket is likely to be broken up into several other tickets as this is an extremely large body of work.

### Motivation

User adoption

### Proposal

Add a new crate with all original functionality from the core rig crate, but expose it using wasm_bindgen.

### Alternatives

Not doing it

joshua-mo-143:
Fixes #550 
Fixes RIG-798

Do not merge until all items on the implementation checklist have been checked off!


joshua-mo-143:
Closing PR for now to save CI minutes.

PR #575, opened 2025-07-09
fix: issue with gemini when there are not tools

morenol:
Fixes:

```
CompletionError: ProviderError: Cannot create OneOrMany with an empty vector.
```

for a agent created with gemini and no tools:

```rust
       // client is a gemini completion client
        let agent = client
            .agent(provider.model())
            .preamble(&metadata.preamble)
            .build();
```


morenol:
maybe duplicated of #576 (?)

PR #576, opened 2025-07-09
fix(rig-809): gemini function declarations should not be OneOrMany

### LINKED ISSUE #574: bug: gemini function declaration empty list error is ambigious

## Bug Report
When using Gemini with tools, if functions are non-existent, you can receive a very ambiguous error.

```
    1: ProviderError: Cannot create OneOrMany with an empty vector.
```

## Implementation
```
let function_declarations: OneOrMany<FunctionDeclaration> = OneOrMany::many(functions)
    .map_err(|x| CompletionError::ProviderError(x.to_string()))?;
```
This piece of code should add extra context that `function_declarations` expected more than one item. Atm, it just wraps the `OneOrMany` error which makes it very elusive to find the source of the problem.

## Expected behavior
A more descriptive error (with context).

## Additional context
I'm sure there's other places that this happens but ideally every `OneOrMany` error has some context attached.

joshua-mo-143:
Fixes #574 
Fixes RIG-809

### DIFF
--- a/rig-core/src/providers/gemini/completion.rs
+++ b/rig-core/src/providers/gemini/completion.rs
@@ -169,12 +169,13 @@ impl TryFrom<completion::ToolDefinition> for Tool {
             } else {
                 Some(tool.parameters.try_into()?)
             };
+
         Ok(Self {
-            function_declarations: OneOrMany::one(FunctionDeclaration {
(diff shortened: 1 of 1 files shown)


0xMochan (approved):
Even better than an error message!

PR #578, opened 2025-07-12
fix(rig-810): prompting OpenAI reponses with message history fails

### LINKED ISSUE #577: bug: Prompting OpenAI Responses with message history fails due to wrong enum variant conversion

## Bug Report

Prompting OpenAI's Responses API fails when used with message history due to it being converted to the wrong enum variant (`OutputMessage` vs. `Message`).

A PR will be opened shortly!

edit: OK this seemed a bit more tricky than before. The ID field as an input is technically not required, but if you put one in it needs to be in the OpenAI format.

joshua-mo-143:
Fixes #577 
Fixes RIG-810

Tested with `debate.rs` to ensure it actually does work.

### DIFF
--- a/rig-core/src/providers/openai/responses_api/mod.rs
+++ b/rig-core/src/providers/openai/responses_api/mod.rs
@@ -903,6 +903,7 @@ pub enum Message {
     },
     Assistant {
         content: OneOrMany<AssistantContentType>,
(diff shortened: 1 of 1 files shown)


abrarmohi (approved):
Looks good to me. Tested it with some message history and it seems to be working with these changes.

PR #601, opened 2025-07-23
fix(rig-820): ensure call ID is properly propagated

### LINKED ISSUE #600: bug: OpenAI Multiturn prompt streaming - Call ID not being set

## Bug Report
I'm working on getting multi-turns working with a streaming interface. I followed the example, it works with Anthropic but OpenAi will panic when it tries to feed the tool call results to openai.

## Reproduction
Take a look at my fork. I had to add the required field to all the tool definitions also.
https://github.com/ChristianPavilonis/rig/blob/multi-turn-example/rig-core/examples/multi_turn_streaming.rs

## Expected behavior
I expect it to fill the call_id in the struct and not panic.

## Additional context
Output when running `cargo run --example multi_turn_streaming`

```
     Running `./rig/target/debug/examples/multi_turn_streaming`
Response: To calculate the expression \( 2 \times (3 + 5) / 9 \), you can break it down into parts:

1. First, calculate \( (3 + 5) \) which equals \( 8 \).
2. Then, multiply that result by \( 2 \): \( 2 \times 8 = 16 \).
3. Finally, divide \( 16 \) by \( 9 \).

Now, I will use the appropriate tools to carry out these operations.To calculate the expression \( 2 \times (3 + 5) / 9 \), you can break it down into parts:

1. First, calculate \( (3 + 5) \) which equals \( 8 \).
2. Then, multiply that result by \( 2 \): \( 2 \times 8 = 16 \).
3. Finally, divide \( 16 \) by \( 9 \).

Now, I will use the appropriate tools to carry out these operations.tool_call: ToolCall { id: "fc_688038fda9c4819a8714b97be537e3c90deec1b9f23c4c10", call_id: None, function: ToolFunction { name: "add", arguments: Object {"x": Number(3), "y": Number(5)} } }
tool_result: "8"
tool_call: ToolCall { id: "fc_688038fdef10819ab2df831deea0b6f70deec1b9f23c4c10", call_id: None, function: ToolFunction { name: "multiply", arguments: Object {"x": Number(2), "y": Number(8)} } }
tool_result: "16"
tool_call: ToolCall { id: "fc_688038fe24a4819ab3a4c6a3060bc65e0deec1b9f23c4c10", call_id: None, function: ToolFunction { name: "divide", arguments: Object {"x": Number(16), "y": Number(9)} } }
tool_result: "1"

thread 'main' panicked at rig-core/src/providers/openai/responses_api/mod.rs:233:54:
The tool call ID should exist!
```

joshua-mo-143:
Fixes #600 
Fixes RIG-820

We weren't setting the call ID in the streaming function, so inevitably nothing was passed in which caused OpenAI to freak out.

edit: There were some additional changes that were required to be made, so the PR title will be amended to reflect this.

### DIFF
--- a/rig-core/Cargo.toml
+++ b/rig-core/Cargo.toml
@@ -42,7 +42,7 @@ as-any = { workspace = true }
 anyhow = { workspace = true }
 assert_fs = { workspace = true }
 tokio = { workspace = true, features = ["full"] }
-tracing-subscriber = { workspace = true }
+tracing-subscriber = { workspace = true, features = ["env-filter"] }
 tokio-test = { workspace = true }
 serde_path_to_error = { workspace = true }
 base64 = { workspace = true }
--- a/rig-core/examples/multi_turn_streaming.rs
+++ b/rig-core/examples/multi_turn_streaming.rs
@@ -109,7 +109,7 @@ where
                         let tool_call_msg = AssistantContent::ToolCall(tool_call.clone());
 
                         tool_calls.push(tool_call_msg);
-                        tool_results.push((tool_call.id, tool_result));
+                        tool_results.push((tool_call.id, tool_call.call_id, tool_result));
 
                         did_call_tool = true;
                         // break;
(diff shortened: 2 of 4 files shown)


joshua-mo-143:
@ChristianPavilonis Please feel free to review at your discretion. 

~~You'll need to make every field `required` in the tool definition parameters for OpenAI in the example, but I think in the interest of not having billions of examples we can just add this as a footnote to the Rig docsite.~~ Amended with the newest commit. I ended up just deriving `schemars::JsonSchema` because it makes more sense and prevents the JSON schema from being so fragile.

ChristianPavilonis (approved):
Tested it. It works great.

PR #606, opened 2025-07-23
feat(rig-823): impl size hint for OneOrMany types

### LINKED ISSUE #605: feat: implement size_hint on Iter, IterMut and IntoIter

## Feature Request
Implementing `size_hint` is possible and would allow the `.collect` calls to optimize the capacity of the receiving container.

### Motivation
Free performance improvement

### Proposal
The `Iterator` trait has a method `size_hint` that library authors should implement whenever possible so that the `.collect` can pre-allocate the memory upfront and avoid multiple allocations.

```rust
impl Iterator for IntoIter {
    // Pseudo-code but should look like
    fn size_hint(&self) -> (usize, Option<usize>) {
        let first = if self.first.is_some() { 1 } else { 0 };
        let max = self.rest.size_hint().1.unwrap_or(0) + first;
        if max > 0 {
          (1, Some(max))
        } else {
          (0, Some(0))
        }
    }
}
```

joshua-mo-143:
Fixes #605 
Fixes RIG-823

cc @Sytten - I believe this seems to work? I pretty much just copy and pasted your implementation in and it seems to be fine (and have added tests to prove it). Theoretically speaking there should always be at least one item, but seeing as I would much rather not voluntarily panic unless there is a need to do so, I've just returned the `(0, Some(0))` instead.

### DIFF
--- a/rig-core/src/one_or_many.rs
+++ b/rig-core/src/one_or_many.rs
@@ -462,6 +462,29 @@ mod test {
     }
 
     #[test]
+    fn test_size_hint() {
+        let foo = "bar".to_string();
+        let one_or_many = OneOrMany::one(foo);
+        let size_hint = one_or_many.iter().size_hint();
+        assert_eq!(size_hint.0, 1);
+        assert_eq!(size_hint.1, Some(1));
+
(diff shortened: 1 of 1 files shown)


Sytten:
Looks good to me! So quick :)

PR #615, opened 2025-07-26
Refactor clients with builder pattern

### LINKED ISSUE #610: feat: allow creating provider client with reqwest client directly

## Feature Request

Right now the way to create a client is as follow:
```rust
rig::providers::openai::Client::new(
            &env::var("OPENAI_API_KEY").expect("OPENAI_API_KEY must be set"),
        )
        .with_custom_client(base_client);
```
But internally the new will create a reqwest client that is then dropped. It would be better to be able to create a client with the reqwest client directly.

### Motivation
Performance improvement for people sharing a client. I am using multiple providers so I am sharing one reqwest client between them since the client will handle the pooling of connections.

### Proposal
The easiest way would probably to add a Builder pattern to the Client creation so we could do:
```rust
rig::providers::openai::Client::builder()
   .with_key("sdadsa")
   .with_http_client(base_client)
   .build()
```

Sytten:
Closes #610

### DIFF
--- a/rig-core/examples/agent_with_galadriel.rs
+++ b/rig-core/examples/agent_with_galadriel.rs
@@ -6,10 +6,13 @@ use std::env;
 #[tokio::main]
 async fn main() -> Result<(), anyhow::Error> {
     // Create Galadriel client
-    let client = providers::galadriel::Client::new(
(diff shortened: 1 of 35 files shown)


joshua-mo-143 (approved):
lgtm

PR #618, opened 2025-07-27
chore(rig-829): update schemars

### LINKED ISSUE #616: chore: update schemars
Okay so this keeps happening to me and I need to write it down before I forget again.

I've been using rig for different projects, and there's this issue with the extractor that gets me every single time. I set up a new project, add rig, write my code —everything compiles fine... then when I run it, it just breaks with these weird errors.

And here's the thing - I waste ages trying to figure out what's wrong. Like genuinely 30-45 minutes each time, checking my code, googling the errors, trying different things. Then suddenly I remember, it's that schemars version thing again!

Turns out rig needs EXACTLY this version:
```toml
schemars = { version = "0.8", features = ["derive"] }
```
But like... this isn't written anywhere obvious? So when cargo suggests the latest version, I add it without thinking. And then everything breaks at runtime.

# Why I'm finally writing this
This has happened to me three or four times now. And I forgot the root cause every time.

# Can we fix this somehow?

## A few ways we could make this better:
**easiest:** Just add a better error message? Like when it fails, check the schemars version and say something like:
Hey, looks like you're using latest version of schemars  but rig needs 0.8
Add this to your Cargo.toml:
```toml
schemars = { version = "0.8", features = ["derive"] }
```
**better:** Maybe rig could include the right version? 

or just: Put a big warning in the docs somewhere. Like right at the top of the extractor examples - "IMPORTANT: You need schemars 0.8, not the latest version"
Last thoughts

I'm happy to help fix this - I can write documentation or whatever is needed. I just thought I should finally report it properly, because if it's happened to me this many times, it's likely happening to others too.
Cheers for all the work on rig btw. Apart from this small thing, it's been great for building agents!

P.S. I'm going to put a sticky note on my monitor that says "SCHEMARS 0.8" so I don't forget again 😅

joshua-mo-143:
Fixes #616 
Fixes RIG-829


joshua-mo-143:
It looks like this is blocked because of `mcp-core` basically breaking on trying to use `schemars@^0.9`.

Sorry @hghalebi, this might have to wait a bit at least until we've had `rmcp` support in a major release so we don't immediately break functionality.

hghalebi:
So just we may add a hint in the documentation?

joshua-mo-143:
ref https://github.com/0xPlaygrounds/rig-docs/pull/25

joshua-mo-143:
closing as it'll be faster to just open a new PR without the conflicts

PR #656, opened 2025-08-03
fix: Ollama provider handling of canonical URLs

### LINKED ISSUE #655: bug: Ollama provider is unfriendly to "/"-terminated URLs

## Bug Report

Both "http://localhost:11434" and "http://localhost:11434/" URLs are technically the same URL. Moreover, the latter is formed according to the specs. But when Ollama provider client is built with it any chat completion request fails with "405 method not allowed".

After some investigation I tracked the problem down to this line: https://github.com/0xPlaygrounds/rig/blob/f55eeb121d56e1dd19a76680475253a9df5a09f3/rig-core/src/providers/ollama.rs#L137

## Expected behavior

Successful execution of the request by Ollama.

## Additional context

The problem becomes the most annoying when the base URL is initially represented as a `url::Url` instance because it canonicalizes all URLs.

vrurg:
Make the provider not only play well with canonical URLs but also implement the ground for safer URL management in the future.

Fixes #655.


vrurg:
I have accidentally spotted a fix for the same problem in the `openai` provider. But I believe the approach proposed here is cleaner because it would intercept badly formed URLs while the agent is being built, not when an HTTP request gets submitted. This would significantly reduce debugging times in particular for both RIG users and developers because accidentally malformed `post` paths would also be nailed down early and clearly.

vrurg:
Unfortunately, I cannot extend this PR to other providers due to lack of time. It looks like it can be copy/pasted into the `openai` code though.

joshua-mo-143:
What version of Rust are you using? You might need to update to get the `fmt` test to pass

vrurg:
It would be great to have a .rustfmt.toml in the root of the project. I accidentally left my own fmt config in ~/src/rust where I keep all Rust-related projects and this is what caused problems with formatting. Having local project config would prevent such accidents in the future.

Anyway, must be OK now.

Issue #658, opened 2025-08-04
bug: OpenAI completions is implemented with responses from v0.14 onward


sredman:
- [x] I have looked for existing issues (including closed) about this

## Bug Report

Upon upgrading from rig v0.13 to v0.14+, requests which were previously being sent as chat completions are now being sent as responses.

## Reproduction

Run a light http server listening on localhost:8888 and see what URL is hit, or just read the error messages when reqwest fails to send and observe the target URL.

v0.13 and v0.14 usage:
```
use rig::{self, completion::Completion};

#[tokio::main]
async fn main() -> Result<(), Box<dyn std::error::Error>> {
    println!("Hello, world!");

    let client = rig::providers::openai::Client::from_url("api_key", "http://localhost:8888");

    let agent = client.agent("model").build();

    // Send a completion request with the agent
    let completion_request = agent.completion("Prompt", Default::default()).await?;

    let _completion_response = completion_request.send().await?;

    Ok(())
}
```

## Expected behavior
Using the `agent.completion` should always send a /chat/completion request

## Screenshots
Not helpful

## Additional context
While this is probably equivalent behavior for anyone using the real OpenAI API, 3rd party APIs (such as LocalAI) which do not support `/responses` are broken.

This seems to be due to the providers::openai::client::CompletionClient returning a [ResponsesCompletionModel](https://github.com/0xPlaygrounds/rig/blob/b5c975ff0753c35d71b68655639fdaf6b18b6749/rig-core/src/providers/openai/client.rs#L132C1-L132C75) introduced in #508 (https://github.com/0xPlaygrounds/rig/pull/508/files#diff-df7412cc113c9245df30d8f1b16320265feb4f16fce84c0be1f834ad4e1d8fe5R83)

Changing that to `completion::CompletionModel` (and resolving the resulting type errors) seems to be a too-good-to-be-true fix, so I mostly intend with this issue to understand whether this is intentional behavior by rig.

sredman:
Just saw the note in https://github.com/0xPlaygrounds/rig/blob/f55eeb121d56e1dd19a76680475253a9df5a09f3/rig-core/examples/openai_agent_completions_api.rs#L2 -- This is apparently intended behavior.

PR #659, opened 2025-08-04
feat: Read the OPENAI_BASE_URL env variable when constructing an OpenAI client from_env

sredman:
Allow constructing an OpenAI client with a custom base url from the environment variable OPENAI_BASE_URL, similar to the Python client.

This PR also migrates all the examples which were previously directly reading the environment to instead use the `openai::client::from_env` constructor.

Tested by building. I'm not able to run the examples E2E due to #658 , but the behavior appears to be correct for some that I spot-checked.

### DIFF
--- a/rig-core/examples/agent.rs
+++ b/rig-core/examples/agent.rs
@@ -1,14 +1,11 @@
 use rig::prelude::*;
-use std::env;
 
 use rig::{completion::Prompt, providers};
 
 #[tokio::main]
 async fn main() -> Result<(), anyhow::Error> {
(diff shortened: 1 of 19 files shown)


sredman:
Tested by locally manipulating some examples to use the /chat/completions API; Those that I tested seem to be doing what they indicate they should be doing.

joshua-mo-143:
Looks like you just need to run `cargo fmt` to format the required files then this should be good to go!

joshua-mo-143 (approved):
lgtm, thanks for the PR!

PR #666, opened 2025-08-04
feat: added get_tool_definitions

### LINKED ISSUE #662: feat: allow getting `Vec<ToolDefinition>` from `ToolSet`

## Feature Request

Currently we're at a position where the user has to separately add all the tools, then create their own `Vec<ToolDefinition>`.

This is not ideal and reduces DX in cases where the user might need to drop down an abstraction level from agents to using completion models directly.

### Motivation

Increase DX

### Proposal

Make a new `get_tool_definitions` function on `ToolSet` that can do the proposed feature.

rajiknows:
fixes #662


joshua-mo-143 (changes requested):
Please add a test for regression monitoring

Issue #648, opened 2025-07-31
bug: Gemini provider fails with `.prompt(...).multi_turn(...)` when using an MCP


afuentesan:
- [x] I have looked for existing issues (including closed) about this

## Bug Report
I'm testing the **Gemini** provider together with the MCP **@modelcontextprotocol/server-filesystem**:
[modelcontextprotocol/servers/filesystem](https://github.com/modelcontextprotocol/servers/tree/main/src/filesystem)

When using Gemini (`gemini-2.5-flash`), this produces:

```
Err: CompletionError: RequestError: Message conversion error: expected value at line 1 column 2
```

When switching to OpenAI (`gpt4.1`), it works fine.
So this seems to be specific to the Gemini provider when working with MCP.

## Reproduction
```rust
use rig::completion::Prompt;
use rmcp::{transport::{ConfigureCommandExt, TokioChildProcess}, ServiceExt};
use tokio::process::Command;


async fn gemini_agent_provider() -> Result<(rig::agent::Agent<rig::providers::gemini::completion::CompletionModel>, rmcp::service::RunningService<rmcp::RoleClient, ()>), String>
{
    let api_key = std::env::var( "GEMINI_FLASH_API_KEY" ).map_err( | e | e.to_string() )?;

    let client = rig::providers::gemini::Client::new( &api_key );

    let agent = client.agent( "gemini-2.5-flash" );

    let mcp_client: rmcp::service::RunningService<rmcp::RoleClient, ()> = mcp_client().await?;

    let tools = mcp_client.list_all_tools().await.map_err( | e | e.to_string() )?;
    
    let agent = tools
        .into_iter()
        .fold(agent, |agent, tool| agent.rmcp_tool(tool, mcp_client.clone()))
        .build();

    Ok( ( agent, mcp_client ) )
}

async fn mcp_client() -> Result<rmcp::service::RunningService<rmcp::RoleClient, ()>, String>
{    
    let client = ()
    .serve(TokioChildProcess::new(Command::new("npx").configure(
        |cmd| {

            cmd.arg( "-y" ).arg( "@modelcontextprotocol/server-filesystem" ).arg( "/tmp" );
        },
    )).map_err( | e | e.to_string() )?)
    .await.map_err( | e | e.to_string() )?;

    Ok( client )
}

async fn run_agent( agent : rig::agent::Agent<rig::providers::gemini::completion::CompletionModel> ) -> Result<String, String>
{
    agent
[...]

PR #683, opened 2025-08-10
Fix SSE parsing in Gemini provider

quangIO:
This fixes tool calling issues on some gemini models. The bug here won't be observed if the server only returns small, completed chunks.

We should not process SSE *byte*stream line-by-line. Therefore, we need to have correct buffering. Instead of reinventing the wheels, we introduce new dependencies (reqwest-eventsource) for parsing SSE chunks.

Related to: https://github.com/0xPlaygrounds/rig/issues/648

The example is pretty redundant but it should document `tool_call.function.name` should be used instead of `id` in gemini.


joshua-mo-143:
Nice. ~~I'll review this tomorrow once I'm in office.~~

This might be a good one to add to OpenAI, albeit in a separate ticket.

joshua-mo-143:
Made some amendments so we can get this merged in before (hopefully!) release tomorrow.

joshua-mo-143 (approved):
lgtm

PR #687, opened 2025-08-11
fix: deepseek streaming endpoint

appflowy:
Hello, I should have opened an issue first about the DeepSeek streaming URL. I believe the DeepSeek endpoint should be consistent with the OpenAI endpoint, so this PR removes the `v1` prefix.


joshua-mo-143 (approved):
lgtm

PR #694, opened 2025-08-12
Consistent visibility modifiers in OpenAI

seafraf:
Some OpenAI completion types are inconsistently private, I need them public so that I can convert them into [OpenAPI generated types](https://github.com/Coral-Protocol/coral-rs/blob/261228cff76e5083dbd6f6fe670bbae23c9cf1a0/src/api.rs#L248)

### DIFF
--- a/rig-core/src/providers/openai/completion/mod.rs
+++ b/rig-core/src/providers/openai/completion/mod.rs
@@ -148,14 +148,14 @@ impl Message {
 
 #[derive(Debug, Serialize, Deserialize, PartialEq, Clone)]
 pub struct AudioAssistant {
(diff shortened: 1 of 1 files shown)


joshua-mo-143:
Hi, thanks for the PR! Just reviewed. ~~Amendments should not take too long to implement, once they pass CI I'll merge as it's a relatively small change.~~ (edit: see below amended review)

joshua-mo-143 (approved):
actually, LGTM - if the breaking changes for the integration are really that much of an issue that we need accessor patterns we'll just do it in a separate ticket as it will need to probably affect the entire module/integration

Issue #696, opened 2025-08-12
feat: impl Tool for Agent<M>


joshua-mo-143:
- [x] I have looked for existing issues (including closed) about this

## Feature Request

Self explanatory, but split #424 into separate tickets.

This issue primarily concerns implementing `Tool` for `Agent<M>` so that it can be easily used as part of a multi-agent system. The "how" of the implementation is pretty self explanatory (just write the impl block) so that section has been omitted from this issue.

### Motivation

Multi agent systems.

dmackdev:
Hi, I'd be interested in picking this up. Could you assign it to me please?

dmackdev:
I've just had a look through the related PR https://github.com/0xPlaygrounds/rig/pull/424 from this issue's description - are the requirements for this issue already satisfied by this [implementation](https://github.com/0xPlaygrounds/rig/pull/424/commits/a2309f66e24b6a5292d703d64f8d4cec998848b0#diff-915a679a34878d42aa08708892a8627e4e5c7b1e82b47393f10dcc8e837a59d3R16)?

dmackdev:
One minor change to that is that I think the `Tool::call` implementation could just call `self.prompt(args.prompt).await`.

joshua-mo-143:
> One minor change to that is that I think the `Tool::call` implementation could just call `self.prompt(args.prompt).await`.

The PR is quite old and assumes that `Tool::call` still requires Sync, which was why it ended up requiring the block_on. I would prefer that we add one tool at a time instead of all 3 so that it's much easier to revert if there are any issues.

PR #704, opened 2025-08-15
feat: implement Tool for Agent

dmackdev:
Closes https://github.com/0xPlaygrounds/rig/issues/696.

- Implements `Tool` for `Agent`, so that an `Agent` can be added as a Tool to be used by another Agent.
- Adds new example `agent_with_agent_tool` to demonstrate an Agent using another Agent as a Tool.

### DIFF
--- a/rig-core/src/agent/tool.rs
+++ b/rig-core/src/agent/tool.rs
@@ -3,10 +3,12 @@ use crate::{
     completion::{CompletionModel, Prompt, PromptError, ToolDefinition},
     tool::Tool,
 };
+use schemars::{JsonSchema, schema_for};
 use serde::{Deserialize, Serialize};
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
 pub struct AgentToolArgs {
+    /// The prompt for the agent to call.
     prompt: String,
 }
 
@@ -27,16 +29,8 @@ impl<M: CompletionModel> Tool for Agent<M> {
                 {}",
                 self.preamble.clone()
             ),
-            parameters: serde_json::json!({
-                "type": "object",
-                "properties": {
-                    "prompt": {
-                        "type": "string",
-                        "description": "The prompt for the agent to call."
-                    }
-                },
-                "required": ["prompt"]
-            }),
+            parameters: serde_json::to_value(schema_for!(AgentToolArgs))
+                .expect("converting JSON schema to JSON value should never fail"),
         }
     }
 
--- a/rig-core/src/agent/tool.rs
+++ b/rig-core/src/agent/tool.rs
@@ -1,0 +1,50 @@
+use crate::{
+    agent::Agent,
+    completion::{CompletionModel, Prompt, PromptError, ToolDefinition},
+    tool::Tool,
+};
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct AgentToolArgs {
+    prompt: String,
+}
+
+impl<M: CompletionModel> Tool for Agent<M> {
+    const NAME: &'static str = "agent_tool";
+
+    type Error = PromptError;
+    type Args = AgentToolArgs;
+    type Output = String;
+
+    async fn definition(&self, _prompt: String) -> ToolDefinition {
+        ToolDefinition {
+            name: <Self as Tool>::name(self),
+            description: format!(
+                "A tool that allows the agent to call another agent by prompting it. The preamble
+                of that agent follows:
+                --- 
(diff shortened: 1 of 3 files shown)


rig-core/src/agent/tool.rs
```diff
@@ -0,0 +1,50 @@
...
+impl<M: CompletionModel> Tool for Agent<M> {
+    const NAME: &'static str = "agent_tool";
+
+    type Error = PromptError;
+    type Args = AgentToolArgs;
+    type Output = String;
+
+    async fn definition(&self, _prompt: String) -> ToolDefinition {
+        ToolDefinition {
+            name: <Self as Tool>::name(self),
```
dmackdev, line 24: This is distinguish between the `Agent::name` method and the `Tool::name` trait method. The `Tool::name` trait method implementation here returns a valid Tool name. 
This is important because the `Agent::name` method could return an invalid Tool name - either an invalid one set by the user, or, if not set, the default name "Unnamed Agent" (which is invalid due to the whitespace).

dmackdev:
@joshua-mo-143 I have tested this locally by modifying the [`agent_with_tools`](https://github.com/0xPlaygrounds/rig/blob/main/rig-core/examples/agent_with_tools.rs) example to introduce an "Agent-Tool using Agent" that uses the existing "Calculator Agent" as a Tool. I can see from the logs that the "Agent-Tool using Agent" calls the "Calculator Agent", which in turn calls the Subtract tool.

Would you like me to add/amend an example to demonstrate this?

joshua-mo-143:
> @joshua-mo-143 I have tested this locally by modifying the [`agent_with_tools`](https://github.com/0xPlaygrounds/rig/blob/main/rig-core/examples/agent_with_tools.rs) example to introduce an "Agent-Tool using Agent" that uses the existing "Calculator Agent" as a Tool. I can see from the logs that the "Agent-Tool using Agent" calls the "Calculator Agent", which in turn calls the Subtract tool.
> 
> Would you like me to add/amend an example to demonstrate this?

Yeah that would be great actually

rig-core/src/agent/tool.rs
```diff
@@ -0,0 +1,50 @@
+use crate::{
+    agent::Agent,
+    completion::{CompletionModel, Prompt, PromptError, ToolDefinition},
+    tool::Tool,
+};
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
joshua-mo-143: Use `schemars::JsonSchema` (the derive macro) here - it's much easier to convert into a JSON schema rather than using fragile JSON

joshua-mo-143 (changes requested):
One change required but that's basically it

rig-core/src/agent/tool.rs
dmackdev: Makes sense. Done.

rig-core/src/agent/tool.rs
```diff
@@ -0,0 +1,44 @@
...
+    completion::{CompletionModel, Prompt, PromptError, ToolDefinition},
+    tool::Tool,
+};
+use schemars::{JsonSchema, schema_for};
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
+pub struct AgentToolArgs {
+    /// The prompt for the agent to call.
+    prompt: String,
```
joshua-mo-143, line 12: Please add the `#[schemars(required)]` macro above this, as it's required for certain providers that the prompt is required for it to pass the argument into the tool.

joshua-mo-143 (changes requested):
Sorry missed something last review - final change and then we'll merge!

rig-core/src/agent/tool.rs
dmackdev: I am not very familiar with the Schemars crate, but wouldn't this be redundant?

From the attribute [docs](https://graham.cool/schemars/deriving/attributes/#required), it would appear that it only affects fields of type `Option`.

Also from the logs when running my new example, I can see it already marks the `prompt` field as required without this attribute:
```
{
  ...
  "parameters": {
    "title": "AgentToolArgs",
    "type": "object",
    "required": [
      "prompt"
    ],
    "properties": {
      "prompt": {
        "description": "The prompt for the agent to call.",
        "type": "string"
      }
    },
    "additionalProperties": false
  },
  ...
}
```

dmackdev: Let me know what you think.

joshua-mo-143: Ah right, yeah - just had a closer look, lgtm!

0xMochan:
Nice, i wasn't able to finish my branch for this. glad it got in there!

PR #713, opened 2025-08-19
fix: Allow CompletionModel Chat to call functions

sredman:
When running the calculator_chatbot example, I observed it would persistently fail with MaxDepthError, like

```
> What is the quotient of 7 and 3?
Error: MaxDepthError: (reached limit: 0)
```

Indicating it was given a max depth of 0, and thus could never work.

This patch gives a multi_turn of 1 which allows basic usage. I'm open for discussion of whether this should be propagated as a decision to the Agent, but I don't see a clean way to do that, and I gather this is primarily example code anyway.

### DIFF
--- a/rig-core/src/agent/completion.rs
+++ b/rig-core/src/agent/completion.rs
@@ -226,6 +226,7 @@ impl<M: CompletionModel> Chat for Agent<M> {
     ) -> Result<String, PromptError> {
         let mut cloned_history = chat_history.clone();
         PromptRequest::new(self, prompt)
+            .multi_turn(1)
             .with_history(&mut cloned_history)
             .await
     }


joshua-mo-143:
Hi, I've actually been thinking about rewriting the `cli_chatbot` function to allow for more flexibility. At the moment while it does work, it is a bit inflexible (and could probably do with adding some extra options), which means that the `Chat` trait itself could probably be more flexible. Let me think about this and get back to you!

0xMochan:
When we added the `PromptRequest` pattern with the `multi_turn` and `with_history` options, it was also discussed whether the `Chat` trait should still exist. Unfortunately, that conversation was forgotten but it might be relevant when considering if `Chat` trait should revamp or deprecate.

joshua-mo-143:
relevant: https://github.com/0xPlaygrounds/rig/issues/742

(added because I think this deserves a bit more attention)

joshua-mo-143:
Superceded by #756 for now while we figure out what the correct way to architect this is.

We will *probably* end up rewriting this entire integration from the ground up. I have a feeling the best way to do this would be to have some kind of `StreamingChat` trait that additionally allows for streamed responses from the CLI chatbot.

However, I think this might warrant some small internal discussion just to get the right of all the details before we proceed further on this.

PR #724, opened 2025-08-20
Add verify client

Sytten:
Allow user to verify if a client is valid without consuming tokens.

The way I built this:
1. Try to use a key verification api if available
2. Fallback to models api
3. Otherwise always return Ok

### DIFF
--- a/rig-bedrock/src/client.rs
+++ b/rig-bedrock/src/client.rs
@@ -103,6 +103,14 @@ impl ImageGenerationClient for Client {
         ImageGenerationModel::new(self.clone(), model)
     }
 }
+
+impl VerifyClient for Client {
+    async fn verify(&self) -> Result<(), VerifyError> {
+        // No API endpoint to verify the API key
+        Ok(())
+    }
+}
+
 impl_conversion_traits!(
     AsTranscription,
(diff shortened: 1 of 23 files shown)


0xMochan:
I've quickly scrolled through this, good idea here! Another endpoint I think would help is token counting as most providers have routes to quickly tokenize a prompt for accounting (ofc, this would be more involved than just verifying so different PR).

Sytten:
Agreed @0xMochan, unsure if we want to generalize the VerifyClient to include that?

One thing that would pair well too is a "list models" function.

0xMochan:
> Agreed @0xMochan, unsure if we want to generalize the VerifyClient to include that?
> 
> One thing that would pair well too is a "list models" function.

perhaps it might make sense as a `MetaClient` set of traits and methods. includes listing models, verifying, and also tokenization.

this shouldn't all be done in one PR, but the seeds of this idea could be thought through.

joshua-mo-143:
apparently there's conflicts here? I don't see any though

lgtm though, and yes we should definitely try to look into trying to expand this out!

Sytten:
@joshua-mo-143 fixed it

PR #733, opened 2025-08-25
feat: Pauseable streams

### LINKED ISSUE #566: feat: pauseable streams

## Feature Request

It should be theoretically possible to pause a stream. The end goal of this would be to save compute by allowing users to pause and resume streams at will.

By using a `tokio::sync::watch::channel` with a bool type, we can control the flow of a stream. See below for a (ChatGPT-generated) example of this in action:

```rust
struct PauseableStream {
    paused_rx: watch::Receiver<bool>,
    counter: usize,
}

impl Stream for PauseableStream {
    type Item = usize;

    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
        if *self.paused_rx.borrow() {
            // Register the task to be notified when the paused value changes.
            let _ = self.paused_rx.changed();
            cx.waker().wake_by_ref(); // try polling again later
            Poll::Pending
        } else {
            let val = self.counter;
            self.counter += 1;
            Poll::Ready(Some(val))
        }
    }
}
```

### Motivation

Better overall control over streamed responses

Dev79844:
- This PR introduces pauseable streams. 

Fixes #566

### DIFF
--- a/rig-core/src/streaming.rs
+++ b/rig-core/src/streaming.rs
@@ -19,11 +19,43 @@ use crate::message::{AssistantContent, Reasoning, Text, ToolCall, ToolFunction};
 use futures::stream::{AbortHandle, Abortable};
 use futures::{Stream, StreamExt};
 use serde::{Deserialize, Serialize};
+use tokio::sync::watch;
 use std::boxed::Box;
 use std::future::Future;
 use std::pin::Pin;
 use std::sync::atomic::AtomicBool;
 use std::task::{Context, Poll};
+
+/// Control for pausing and resuming a streaming response
+pub struct PauseControl {
+    pub(crate) paused_tx: watch::Sender<bool>,
+    pub(crate) paused_rx: watch::Receiver<bool>,
+}
+
+impl PauseControl {
+    pub fn new() -> Self {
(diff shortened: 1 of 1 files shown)


joshua-mo-143:
Hi @Dev79844, I saw your comment about asking if we should impl this for the agent streaming stuff but you appear to have deleted it.

I think at the moment we may want to proceed on this first, then once the agent multi turn streaming API stabilises (see #739 ) we can definitely get the go-ahead on that, as well as figuring out how to add functionality to stop that stream as I know it will be required for things like guardrails.

Dev79844:
Hey @joshua-mo-143 
Got it! I will remove this PR from draft state and when the multi turn streaming API stabilises, will work on adding the streams at agent level.

joshua-mo-143:
Looks like CI has broken. Will wait for CI to pass before a review can happen.

Dev79844:
Hey @joshua-mo-143 
I have fixed the issues.

joshua-mo-143:
Great work @Dev79844 - thank you for this PR! LGTM

Issue #721, opened 2025-08-20
bug(gemini): Empty type when we pass a vector to extractor


YouKnow-sys:
- [X] I have looked for existing issues (including closed) about this

## Bug Report
when we for example create a extractor that extract a `Vec<T>` somehow when running it the log shows that the generated schema have a empty type for items inside the vector.
seems like in some step the internal type is lost
here is what the log show me
```json
"functionDeclarations": [
  {
    "name": "submit",
    "description": "Submit the structured data you extracted from the provided text.",
    "parameters": {
      "type": "array",
      "items": {
        "type": ""
      }
    }
  }
]
```

## Reproduction
I modified [this example](https://github.com/0xPlaygrounds/rig/blob/main/rig-core/examples/gemini_extractor.rs) to extract a list of Person (not correct but should work)
```rust
use std::marker::PhantomData;

use rig::client::{CompletionClient, ProviderClient};
use rig::providers::gemini;
use schemars::JsonSchema;
use serde::{Deserialize, Serialize};

#[derive(Debug, Deserialize, JsonSchema, Serialize)]
struct Person {
    /// The person's first name, if provided (null otherwise)
    pub first_name: Option<String>,
    /// The person's last name, if provided (null otherwise)
    pub last_name: Option<String>,
    /// The person's job, if provided (null otherwise)
    pub job: Option<String>,
}

#[tokio::main]
async fn main() -> Result<(), anyhow::Error> {
    tracing_subscriber::fmt()
        .with_max_level(tracing::Level::DEBUG)
        .with_target(false)
        .init();

    let params = gemini::completion::gemini_api_types::AdditionalParameters::default();

    // Create Gemini client
    let client = gemini::Client::from_env();

    // Create extractor
    let data_extractor = client
        .extractor::<Vec<Person>>(gemini::completion::GEMINI_2_0_FLASH)
        .additional_params(serde_json::to_value(params)?)
        .build();

    let person = data_extractor
        .extract("[\"Hello my name is John Doe! I am a software engineer.\" , \"Hello my name is John Doe! I am a software engineer.\"]")
        .await.unwrap();

    println!("GEMINI: {}", serde_json::to_string_pretty(&person).unwrap());

    Ok(())
}
```

## Expected behavior
here is dumped object schema
```rust
Object {
    "$defs": Object {
        "Person": Object {
            "properties": Object {
                "first_name": Object {
                    "description": String("The person's first name, if provided (null otherwise)"),
                    "type": Array [
[...]

PR #748, opened 2025-08-27
fix: empty type in Vec<T> schema conversion for Gemini API (#721)

Soroushsrd:
### Problem
When creating extractors for Vec<T> types (e.g., Vec<Person>), the generated Gemini API schema contains an empty type field for array items, causing invalid function declarations
### Root Cause: 
The schemars crate generates JSON schemas with $ref references to avoid duplicating type definitions:
```
{
  "type": "array",
  "items": {
    "$ref": "#/$defs/Person"  ---> reference to Person definition
  },
  "$defs": {
    "Person": {
      "type": "object",
      "properties": { ... }
    }
  }
}
```
but the TryFrom<Value> for Schema implementation doesn't handle $ref references, so when it encounters an items object with only $ref (no type field), it defaults to an empty string through unwrap_or_default().

### Solution
Added schema flattening that resolves all $ref references before converting to Gemini's Schema format
Updated TryFrom<Value> for Schema to flatten schemas before conversion
Tested new methods/functionalities to make sure everything works


Soroushsrd:
fixes #721

rig-core/src/providers/gemini/completion.rs
```diff
@@ -1091,11 +1091,103 @@ pub mod gemini_api_types {
         pub items: Option<Box<Schema>>,
     }
 
+    pub fn flatten_schema(mut schema: Value) -> Result<Value, CompletionError> {
```
joshua-mo-143, line 1099: docstrings

rig-core/src/providers/gemini/completion.rs
```diff
@@ -1091,11 +1091,103 @@ pub mod gemini_api_types {
...
+                    resolve_refs(item, defs)?;
+                }
+            }
+            _ => {}
+        }
+
+        Ok(())
+    }
+
+    fn parse_ref_path(ref_str: &str) -> Result<String, CompletionError> {
```
joshua-mo-143, line 1172: docstrings

joshua-mo-143 (changes requested):
LGTM, just needs some docstrings.

You may need to update your Rust version to see new lints.

rig-core/src/providers/gemini/completion.rs
Soroushsrd: Thanks. i think these changes should do it

PR #756, opened 2025-08-30
feat(rig-904): Rework CLI chatbot integration

### LINKED ISSUE #742: feat: rework CLI chatbot integration

## Feature Request

The `cli_chatbot` function is currently a bit limited. We should really rework it:
- Instead of taking `impl Chat`, it should take `Agent<M>`
- It should also take a max depth for multi-turn
- Perhaps some other functionality too? Can't really think of anything at the moment.

### Motivation

Significantly improved DX

FayCarsons:
Fixes #742 

## Changes 
- Replaced the `cli_chatbot` function with `Chatbot` and `ChatbotBuilder`
- Type-safe builder using the "type-state" pattern, where the `build` method is not available until the agent has been set
- Added configuration options for multi-turn depth and displaying token usage
- Added streaming requests, with response chunks printed as they are received
- Refactored examples using `cli_chatbot`

## API
```rust
let chatbot = ChatbotBuilder::new()
  .agent(my_agent)
  .multi_turn_depth(1)
  .show_usage()
  .build();

chatbot.run().await?;
```

## Notes
- I wasn't able to match the API @joshua-mo-143 suggested exactly (`Chatbot::builder`) due to inference issues with the generic parameter in `Agent<M>`. 
- The multi-agent example had to be refactored significantly to match the new API. It now composes agents by providing one as a tool to the other, as opposed to the previous approach where both are placed in a struct implementing `Chat`. I'm not 100% that this still aligns with the original intention.

### DIFF
--- a/rig-core/src/cli_chatbot.rs
+++ b/rig-core/src/cli_chatbot.rs
@@ -1,44 +1,193 @@
 use std::io::{self, Write};
 
-use crate::completion::{Chat, Message, PromptError};
+use futures::StreamExt;
 
-/// Utility function to create a simple REPL CLI chatbot from a type that implements the
-/// `Chat` trait.
-pub async fn cli_chatbot(chatbot: impl Chat) -> Result<(), PromptError> {
-    let stdin = io::stdin();
-    let mut stdout = io::stdout();
-    let mut chat_log = vec![];
+use crate::{
+    agent::{Agent, prompt_request::streaming::MultiTurnStreamItem},
+    completion::{CompletionError, CompletionModel, Message, PromptError},
+    streaming::StreamingPrompt,
+};
 
-    println!("Welcome to the chatbot! Type 'exit' to quit.");
-    loop {
-        print!("> ");
-        // Flush stdout to ensure the prompt appears before input
-        stdout.flush().unwrap();
+/// Type-state representing an empty `agent` field in `ChatbotBuilder`
+pub struct AgentNotSet;
 
-        let mut input = String::new();
-        match stdin.read_line(&mut input) {
-            Ok(_) => {
-                // Remove the newline character from the input
-                let input = input.trim();
-                // Check for a command to exit
-                if input == "exit" {
-                    break;
-                }
-                tracing::info!("Prompt:\n{}\n", input);
+/// Builder pattern for CLI chatbots
+///
+/// # Example
+/// ```rust
+/// let chatbot = ChatbotBuilder::new().agent(my_agent).show_usage().build();
+///
+/// chatbot.run().await?;
+pub struct ChatbotBuilder<A> {
+    agent: A,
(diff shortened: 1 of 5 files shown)


rig-core/src/cli_chatbot.rs
```diff
@@ -1,44 +1,193 @@
...
+        ChatbotBuilder {
+            agent: AgentNotSet,
+            multi_turn_depth: 0,
+            show_usage: false,
+        }
+    }
+}
+
+impl ChatbotBuilder<AgentNotSet> {
+    pub fn new() -> Self {
```
FayCarsons, line 38: Ideally, `Chatbot` would have a static `builder` function to create builders, but because the generic parameter couldn't be inferred users would have to specify the type i.e. `Chatbot::<ResponsesCompletionModel>::builder`.

We could also consolidate the types into one but then I'm not sure what the `build` method would be achieving.

joshua-mo-143:
Awesome work @FayCarsons !

Not going to lie, I am wondering whether it would be better to have `A: impl Chat` in this. The problem is that the `Chat` trait itself actually needs refactoring to accommodate for multi-turn, but because there has been zero work scoped out for that yet... well, we can't really do anything about it yet. I think this will work for now however (and is reasonably Good Enough that we can merge it).

rig-core/src/cli_chatbot.rs
joshua-mo-143: ~~Sorry just remembered to respond to this - we could probably do `fn builder() -> ChatbotBuilder<AgentNotSet>`. Using `.agent()` basically sets the internal type again anyway, so it wouldn't really matter.~~

~~What do you think?~~

Actually nevermind, I see your point - just tried it locally and because `Chatbot<M>` requires the trait bound at the struct level, we basically can't infer this part.

FayCarsons:
Thank you! Honestly I had the exact same thought when trying to work out the multi-agent example, it would definitely make it easier to implement similar patterns.

joshua-mo-143 (commented):
lgtm - just wanted to reply to your comment very quickly in case you wanted to make amendments for it

rig-core/src/cli_chatbot.rs
FayCarsons: Hmm yeah it's kind of frustrating because the type shouldn't matter at all, but adding that as a method to `Chatbot` gives us a type error at the call site unless we specify the generic param.

Or did you mean as a method of `ChatbotBuilder`?

joshua-mo-143: Nah I meant as a method for `Chatbot`.

I think it should be OK for now. Not much we can really do we want to show that the struct requires `M: CompletionModel`. Once we solve the `Chat` trait issue, I think it'll be evident on how we can improve it.

PR #757, opened 2025-08-30
Fix tool definitions in calculator example

FayCarsons:
The calculator example tool definitions are missing the `required` field which causes them to fail with GPT-4, this adds those fields


joshua-mo-143 (approved):
lgtm

Issue #496, opened 2025-06-08
feat: request modelling for every single provider


joshua-mo-143:
- [X] I have looked for existing issues (including closed) about this

## Feature Request

Sub-ticket of #381.

There's a lot more work ahead of this ticket, but we have to start somewhere. As completion models are the most used model type in this library, this makes the most sense.

### Motivation

Advancing #381

### Proposal

In accordance with the overarching epic, we should probably model requests for each provider so that we can add IDs to each request (whether it's an agent ID or a conversation ID).

This work primarily involves adding a breaking change to the `CompletionModel` trait so that we have an associated `Request` type - then all we need to do is impl `From<CompletionRequest>` for each request made.

### Alternatives 

None

Sarah-2003:
Hi can you please assign me this

Sarah-2003:
#693 Can u please assign me this

joshua-mo-143:
Assigning to myself as this is directly blocking some other work I need to implement.

PR #780, opened 2025-09-05
Update lib.rs: make json_utils public

xfanplus:
json_utils is very useful to implement your own model provider integration by defining types that implement the CompletionModel and EmbeddingModel traits.

### DIFF
--- a/rig-core/src/lib.rs
+++ b/rig-core/src/lib.rs
@@ -118,7 +118,7 @@ pub mod extractor;
 #[cfg(feature = "image")]
 #[cfg_attr(docsrs, doc(cfg(feature = "image")))]
 pub mod image_generation;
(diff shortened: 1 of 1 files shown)


joshua-mo-143:
Hi, the `json_utils` module is not intended to be a part of the public API so making it `pub` is unfeasible. It is suggested to instead make your own request type and use `TryFrom<(String, CompletionRequest)>` (where String is a model name), as JSON values generally can be quite brittle and awkward to work with in Rust (as noted in #496).

However, the entirety of `rig-core` (and currently the rest of the repo) is MIT, we plan on keeping it MIT-licensed and we haven't had to add anything to the `json_utils` module for a very, very long time - so you may copy as you wish :slightly_smiling_face:

PR #781, opened 2025-09-05
feat: Add extractor hooks and validation system

### LINKED ISSUE #770: feat: extractor hooks

## Feature Request

We should have a way to validate that a struct validates all arbitrary requirements without leaving the extractor retry loop.

We should also have a way to add introspection to the extractor (like prompt hooks).

### Motivation

Increase DX, observability... etc

### Proposal
Something like the below:
```rust
trait ExtractorHook: Clone + Send + Sync {
     // .. some methods
}

trait ExtractorValidatorHook: Clone + Send + Sync {
    fn validate() -> impl Future<Output = Result<String, Error>>;
}
```

I wonder if it would be a good idea for us to attach the error as a tool result back to the conversation if we are unable to meet the requirements.

Soroushsrd:
- Add ExtractorHook trait for lifecycle observability
- Add ExtractorValidatorHook trait for custom validation
- Add ExtractorWithHooks extension trait
- Add comprehensive documentation and examples
- Add integration tests for hook functionality"

closes #770

### DIFF
--- a/rig-core/src/extractor.rs
+++ b/rig-core/src/extractor.rs
@@ -27,12 +27,28 @@
 //!     .await
 //!     .expect("Failed to extract data from text");
 //! ```
-
-use std::marker::PhantomData;
+//! # Hooks and Validation
+//!
+//! For advanced observability and validation, use the hook system:
+//!
+//! ```rust
+//! use rig::extractor::{ExtractorWithHooks, ExtractorHook, ExtractorValidatorHook};
+//!
+//! // Create hooks and validators
+//! let hooks: Vec<Box<dyn ExtractorHook>> = vec![Box::new(LoggingHook)];
+//! let validators: Vec<Box<dyn ExtractorValidatorHook<Person>>> = vec![Box::new(AgeValidator)];
+//!
+//! // Use with any extractor - must import ExtractorWithHooks trait
+//! let result = extractor.extract_with_hooks(text, hooks, validators).await?;
+//!
 
 use schemars::{JsonSchema, schema_for};
 use serde::{Deserialize, Serialize};
-use serde_json::json;
+use serde_json::{Value, json};
+use std::boxed::Box;
+use std::future::Future;
+use std::marker::PhantomData;
+use std::pin::Pin;
 
 use crate::{
     agent::{Agent, AgentBuilder},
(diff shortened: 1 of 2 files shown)


Soroushsrd:
Hey Josh. I would love to hear your take on how I implemented this feature. If it can be done better, please let me know

rig-core/src/extractor.rs
```diff
@@ -53,6 +69,271 @@ pub enum ExtractionError {
...
+///
+/// let validators: Vec<Box<dyn ExtractorValidatorHook<Person>>> = vec![
+///     Box::new(AgeValidator { min_age: 18, max_age: 120 }),
+/// ];
+///
+/// let result = extractor.extract_with_hooks(text, hooks, validators).await?;
+/// ```
+///
+/// [`extract_with_hooks`]: ExtractorWithHooks::extract_with_hooks
+pub trait ExtractorWithHooks<T>
```
joshua-mo-143, line 321: I think this is generalising perhaps a bit too far for the scope. We should ideally have an `ExtractorWithHooks` struct I think for now. If it's evident that users actually want this trait because they need to build their own extractors (and the extractor hooks into other things), then I think we should consider it then.

rig-core/src/extractor.rs
```diff
@@ -53,6 +69,271 @@ pub enum ExtractionError {
...
+///             Ok(())
+///         })
+///     }
+/// }
+/// ```
+/// You can chain multiple validators together. They are executed in order, and the first
+/// validation failure will cause the extraction attempt to fail:
+///
+/// ```rust
+/// let validators: Vec<Box<dyn ExtractorValidatorHook<Person>>> = vec![
```
joshua-mo-143, line 276: We should try to abstract away as much as possible the need to use `Vec<Box<T>>` from the user side. It is a bit clunky in terms of DX.

rig-core/src/extractor.rs
```diff
@@ -66,6 +347,63 @@ where
...
+        validators: Vec<Box<dyn ExtractorValidatorHook<T>>>,
+    ) -> Pin<Box<dyn Future<Output = Result<T, ExtractionError>> + Send + '_>> {
+        let text_msg = text.into();
+
+        Box::pin(async move {
+            let mut last_error = None;
+
+            for i in 0..=self.retries {
+                tracing::debug!(
+                    "Attempting to extract Json. Retries left:{retries}",
```
joshua-mo-143, line 371: should be `JSON` or `json`, not `Json`

rig-core/src/extractor.rs
```diff
@@ -66,6 +347,63 @@ where
...
+                        let data_value =
+                            serde_json::to_value(&data).unwrap_or(serde_json::Value::Null);
+                        for hook in &hooks {
+                            hook.on_success(i, &data_value).await;
+                        }
+                        return Ok(data);
+                    }
+                    Err(e) => {
+                        tracing::warn!(
+                            "Attempt number {i} to extract Json failed: {e:?}. Retrying..."
```
joshua-mo-143, line 393: same here

rig-core/src/extractor.rs
```diff
@@ -66,6 +347,63 @@ where
...
 
+impl<M, T> ExtractorWithHooks<T> for Extractor<M, T>
+where
+    M: CompletionModel,
+    T: JsonSchema + for<'a> Deserialize<'a> + Serialize + Send + Sync,
+{
+    /// This implementation creates the complete extraction lifecycle with
+    /// observability and validation. It manages the retry loop, coordinates
+    /// hook calls, and ensures validators are executed in the correct sequence.
+    fn extract_with_hooks(
```
joshua-mo-143, line 358: fyi I believe when you are implementing a trait for a type (that has `fn() -> <some_future>`, you can use `async fn` rather than just rawdogging the future

rig-core/src/extractor.rs
```diff
@@ -155,6 +493,49 @@ where
         Ok(serde_json::from_value(raw_data)?)
     }
 
+    async fn extract_validated_json(
+        &self,
+        text: impl Into<Message> + Send,
+        attempt: u64,
+        hooks: &[Box<dyn ExtractorHook>],
```
joshua-mo-143, line 500: as mentioned before, we should try to avoid using arrays of `Box<T>` where possible

rig-core/tests/extractor_hooks.rs
```diff
@@ -0,0 +1,457 @@
...
+        Err(e) => {
+            println!("EXPECTED FAILURE: Validation failed after retries: {}", e);
+            assert!(events.iter().any(|e| e.starts_with("On error")));
+        }
+    }
+}
+
+#[tokio::test]
+#[ignore]
+async fn test_multiple_validators() {
```
joshua-mo-143, line 353: general comment: An example to hold one of these tests in would be great. I know there's literally a gazillion examples already (and we will be cleaning it up!), but for now an example would be great.

joshua-mo-143 (changes requested):
Great work! Some changes needed but I think otherwise mostly OK. Good job on ensuring docstrings, too.

Generally speaking:
- Avoid `Vec<Box<T>>` (and `&[Box<T>]`) where possible as an input argument - we should be handling this internally (see `ToolSet` struct if stuck). For example:

```struct
struct Hooks(Vec<Box<dyn ExtractorHook>>);

impl Hooks {
    fn add_hook<T>(mut self, hook: T) -> Self 
        where 
        T: ExtractorHook + 'static {
        self.0.push(Box::new(hook));
        self
    }
}
```
- Having `ExtractorWithHooks<T>` might be a bit too far of a leap in terms of traits at the moment. Ideally we should have a way to simply transition from one struct, `Extractor<M, T>` to an `ExtractorWithHooks<M, T>`.
- Please add an example to the `rig-core` examples list.

Soroushsrd:
Got it. thank you for the comments. will sort them out.

PR #793, opened 2025-09-08
chore: optional candidates token count

### LINKED ISSUE #788: bug: issues with gemma-3-27b-it with gemini

## Bug Report
Error: CompletionError(ProviderError("{\n  \"error\": {\n    \"code\": 400,\n    \"message\": \"Developer instruction is not enabled for models/gemma-3-27b-it\",\n    \"status\": \"INVALID_ARGUMENT\"\n  }\n}\n"))
## Reproduction
Use gemini_agent.rs but with model `gemma-3-27b-it`.

This happens even when preamble is `None` due to this: https://github.com/0xPlaygrounds/rig/blob/2013a6c27a2da0ae425670de2f65c1b296d012d1/rig-core/src/agent/builder.rs#L182

## Expected behavior
Expected the agent to be working with that model

## Additional context

I think that is due to this system instruction sent
```json
"systemInstruction": {
    "parts": [
      {
        "thought": false,
        "text": ""
      }
    ],
    "role": "model"
  }
```

morenol:
Should fix #788 

not sure which is the source of truth, but here it is optional:

https://github.com/googleapis/python-genai/blob/842974434824c1d77a14ceec13a9591b5d86db3c/google/genai/types.py#L5339

And I get this error with the current main implementation when using gemma: `gemma-3-27b-it`

```
2025-09-08T20:57:06.984165Z  INFO Response: Err(CompletionError(HttpError(reqwest::Error { kind: Decode, source: Error("missing field `candidatesTokenCount`", line: 25, column: 3) })))
2025-09-08T20:57:06.984217Z ERROR Error: CompletionError(HttpError(reqwest::Error { kind: Decode, source: Error("missing field `candidatesTokenCount`", line: 25, column: 3) }))
2025-09-08T20:57:06.984302Z  INFO return=Err(CompletionError: HttpError: error decoding response body

Caused by:
    0: HttpError: error decoding response body
    1: error decoding response body
    2: missing field `candidatesTokenCount` at line 25 column 3)
Error: CompletionError: HttpError: error decoding response body

```

### DIFF
--- a/rig-core/src/providers/gemini/completion.rs
+++ b/rig-core/src/providers/gemini/completion.rs
@@ -280,7 +280,7 @@ impl TryFrom<GenerateContentResponse> for completion::CompletionResponse<Generat
             .as_ref()
             .map(|usage| completion::Usage {
                 input_tokens: usage.prompt_token_count as u64,
-                output_tokens: usage.candidates_token_count as u64,
+                output_tokens: usage.candidates_token_count.unwrap_or(0) as u64,
                 total_tokens: usage.total_token_count as u64,
             })
             .unwrap_or_default();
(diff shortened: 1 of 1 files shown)


joshua-mo-143:
Ok we should probably reconcile with the Python SDK in that case then as that is actually supported by Google.

PR #823, opened 2025-09-16
feat(rig-795): support file URLs for audio, video, documents

FayCarsons:


rig-core/src/completion/message.rs
```diff
@@ -435,27 +451,44 @@ impl UserContent {
...
         media_type: Option<DocumentMediaType>,
     ) -> Self {
         UserContent::Document(Document {
-            data: data.into(),
+            data: DocumentSourceKind::Base64(data.into()),
             format,
             media_type,
             additional_params: None,
         })
     }
```
FayCarsons: Is `Base64` what we want this variant to be named? It seems like there's a lot of places it effectively used to mean "any raw data", unless I'm misunderstanding something? I wanted to generally take the path of least resistance so I didn't rename it or anything.

rig-core/src/providers/anthropic/completion.rs
```diff
@@ -366,6 +402,7 @@ impl TryFrom<message::Message> for Message {
                                 Ok(ToolResultContent::Text { text })
                             }
                             message::ToolResultContent::Image(image) => {
+                                // NOTE: (@FayCarsons) Do we want to support URLs here?
                                 let DocumentSourceKind::Base64(data) = image.data else {
                                     return Err(MessageError::ConversionError(
                                         "Only base64 strings can be used with the Anthropic API"
```
FayCarsons: Wasn't 100% whether we wanted to count anthropic File API File IDs as URLs, or how interfacing with their File API is (or isn't) being handled

rig-core/src/providers/openai/completion/mod.rs
```diff
@@ -319,6 +319,7 @@ impl TryFrom<message::Message> for Vec<Message> {
                         message::UserContent::Image(message::Image {
                             data, detail, ..
                         }) => {
+                            // NOTE: Does this include data URIs? Or should we be accepting base64 as well  
                             let DocumentSourceKind::Url(url) = data else { return Err(message::MessageError::ConversionError(
                                 "Only image URL user content is accepted with OpenAI Chat Completions API".to_string()
                             ))};
```
FayCarsons: Wasn't sure if URLs include data URIs here or if that would be covered by the base64 variant

rig-core/src/completion/message.rs
joshua-mo-143: Yeah pretty much. The only possible option that can be used as raw data currently (wrt APIs) is base64-encoded strings, which means we have to use Strings.

I guess technically we can use a newtype wrapper to signify that it's *supposed* to be a base64-encoded string and then use a `new()?` pattern, but for large files (ie docs/images) depending on the size that might come at a large overhead. Might need to do some experimentation to see what exactly the overhead is.

rig-core/src/providers/openai/completion/mod.rs
joshua-mo-143: The base64 variant is basically just whatever the base64-encoded string is so you'll probably need to convert it to a data URI yourself. 

(but yes it should be fine - just checked the API and apparently they do support base64 encoded images now)

rig-core/src/providers/anthropic/completion.rs
joshua-mo-143: Huh, their API seems to have actually been updated.

It isn't actually being handled at the moment, but I think we can probably just revisit this in a later PR

rig-core/src/completion/message.rs
FayCarsons: For sure, I think we can probably figure out a way to make it a transparent newtype which gets erased at runtime? But I'll look into that and make an issue if so

rig-core/src/providers/openai/completion/mod.rs
FayCarsons: Awesome, just added data URIs here.

rig-bedrock/src/types/document.rs
```diff
@@ -80,24 +84,36 @@ mod tests {
...
     };
 
     use crate::types::document::RigDocument;
 
     #[test]
     fn test_document_to_aws_document() {
         let rig_document = RigDocument(Document {
-            data: "data".into(),
+            data: DocumentSourceKind::Base64("data".into()),
             format: Some(ContentFormat::String),
```
joshua-mo-143: We can get rid of this `format` field as the primary reason for it existing now no longer exists due to us using `DocumentSourceKind`

rig-core/src/completion/message.rs
```diff
@@ -232,7 +239,7 @@ pub struct Video {
 /// Document content containing document data and metadata about it.
 #[derive(Default, Clone, Debug, Deserialize, Serialize, PartialEq)]
 pub struct Document {
-    pub data: String,
+    pub data: DocumentSourceKind,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub format: Option<ContentFormat>,
```
joshua-mo-143: No longer required due to `DocumentSourceKind` already being able to represent this field

rig-core/src/completion/message.rs
```diff
@@ -220,7 +227,7 @@ pub struct Audio {
 /// Video content containing video data and metadata about it.
 #[derive(Default, Clone, Debug, Deserialize, Serialize, PartialEq)]
 pub struct Video {
-    pub data: String,
+    pub data: DocumentSourceKind,
```
joshua-mo-143, line 228: No longer required due to `DocumentSourceKind` already being able to represent this field

rig-core/src/completion/message.rs
```diff
@@ -208,7 +215,7 @@ impl std::fmt::Display for DocumentSourceKind {
 /// Audio content containing audio data and metadata about it.
 #[derive(Default, Clone, Debug, Deserialize, Serialize, PartialEq)]
 pub struct Audio {
-    pub data: String,
+    pub data: DocumentSourceKind,
```
joshua-mo-143, line 218: No longer required due to `DocumentSourceKind` already being able to represent this field

rig-core/src/providers/anthropic/completion.rs
```diff
@@ -366,6 +402,7 @@ impl TryFrom<message::Message> for Message {
                                 Ok(ToolResultContent::Text { text })
                             }
                             message::ToolResultContent::Image(image) => {
+                                // NOTE: (@FayCarsons) Do we want to support URLs here?
```
joshua-mo-143: you probably need to remove this

joshua-mo-143 (changes requested):
few small changes needed, other than that lgtm - great work fay

FayCarsons:
Having some trouble getting this building + showing me errors and lints locally, will get CI failures fixed

PR #826, opened 2025-09-17
Fix:  parse NDJSON correctly, fixes #825

### LINKED ISSUE #825: bug:  Ollama does not support SSE, only NDJSON 🐈
# Ollama does not support SSE, only NDJSON
## Description:
When using the `stream()` method against Ollama, the client expects SSE (Server-Sent Events) format. 
However, Ollama's API only supports NDJSON streaming.

As a result, the client fails to parse the response and raises an error.

## Uncorrect
https://github.com/0xPlaygrounds/rig/blob/b4a2c32cc89d8c4075ca181cbd2a0c5ee622ae13/rig-core/src/providers/ollama.rs#L559-L636

## Fixed
```rust
    async fn stream(
        &self,
        request: CompletionRequest,
    ) -> Result<streaming::StreamingCompletionResponse<Self::StreamingResponse>, CompletionError>
    {
        let mut request_payload = self.create_completion_request(request)?;
        merge_inplace(&mut request_payload, json!({"stream": true}));

        let response = self
            .client
            .post("api/chat")?
            .json(&request_payload)
            .send()
            .await
            .map_err(|e| CompletionError::ProviderError(e.to_string()))?
            .bytes_stream();

        let stream = Box::pin(stream! {
            for await chunk in response {
                let bytes = match chunk {
                    Ok(b) => b,
                    Err(e) => {
                        yield Err(CompletionError::ResponseError(e.to_string()));
                        continue;
                    }
                };

                match serde_json::from_slice::<CompletionResponse>(&bytes) {
                    Ok(response) => {
                        if response.done {
                            yield Ok(RawStreamingChoice::FinalResponse(
                                StreamingCompletionResponse {
                                    total_duration: response.total_duration,
                                    load_duration: response.load_duration,
                                    prompt_eval_count: response.prompt_eval_count,
                                    prompt_eval_duration: response.prompt_eval_duration,
                                    eval_count: response.eval_count,
                                    eval_duration: response.eval_duration,
                                    done_reason: response.done_reason,
                                }
                            ));
                            break;
                        }

                        if let Message::Assistant { content, tool_calls, .. } = response.message {
                            if !content.is_empty() {
                                yield Ok(RawStreamingChoice::Message(content));
                            }
                            for tool_call in tool_calls {
                                yield Ok(RawStreamingChoice::ToolCall {
                                    id: String::new(),
                                    name: tool_call.function.name,
                                    arguments: tool_call.function.arguments,
                                    call_id: None,
                                });
                            }
                        }
                    }
                    Err(err) => {
                        tracing::error!(?err, "NDJSON error"); 
                        yield Err(CompletionError::JsonError(err));
                    }
                }
            }
        });
```
### 🐈🐾 🐾🐾 🐾 💩

MariaGolubev:
Here are the fixes I propose. I'm not entirely sure everything is correct, so please review the code.
[ #825](https://github.com/0xPlaygrounds/rig/issues/825)

### DIFF
--- a/rig-core/src/providers/ollama.rs
+++ b/rig-core/src/providers/ollama.rs
@@ -53,10 +53,10 @@ use crate::{
     message::{ImageDetail, Text},
     streaming,
 };
-use async_stream::stream;
+use async_stream::try_stream;
 use futures::StreamExt;
 use reqwest;
-use reqwest_eventsource::{Event, RequestBuilderExt};
+// use reqwest_eventsource::{Event, RequestBuilderExt}; // (Not used currently as Ollama does not support SSE)
 use serde::{Deserialize, Serialize};
 use serde_json::{Value, json};
 use std::{convert::TryFrom, str::FromStr};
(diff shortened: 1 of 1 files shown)


joshua-mo-143:
Hi Maria! Thanks for opening the PR.

We will probably need to revert to the previous implementation that was implemented prior to the update. You can find the original implementation here: https://github.com/0xPlaygrounds/rig/blob/9a3a5ef4532c2b75822d8001c1e20f44c40d3aab/rig-core/src/providers/ollama.rs

An additional comment will likely need to be provided to say that ollama doesn't support SSE in `/generate` and `/chat`.

Reference issue https://github.com/ollama/ollama/issues/11586

MariaGolubev:
Overall, Ollama is compatible with OpenAI. It might support SSE through /v1/chat/completions, but I haven’t checked that. However, the standard protocol for Ollama is NDJSON.

joshua-mo-143:
Yeah, but at that point you might as well just use the OpenAI Chat Completions API and use ollama as the base URL... which then makes it impossible to use the reasoning field for ollama.

I think if you switch back to the original linked implementation (which does support NDJSON for the most part as I've used it before and it works pretty much perfectly) then we can get the CI started and merge this in :slightly_smiling_face:

MariaGolubev:
I rewrote the code using the old version as a base. I removed unnecessary allocations, so it should work more efficiently with large message streams. I’m using` try_stream` to make the code more readable, but I’m not sure if I’m handling errors correctly.

joshua-mo-143:
Hey Maria, quick status update: I'll be reviewing this by EOW. It *should* be fine and it doesn't look like there is anything that needs changing, but just need to test it to make sure it works.

MariaGolubev:
Oh, I didn’t read your message. I added the correct error type handling. Before, `EmbeddingError::ProviderError()`  was used everywhere, which is sometimes incorrect. I also added tracing debug information so that errors in services like [Jaegert](https://www.jaegertracing.io/)  can be tracked.

```sh
2025-09-18T16:17:19.468126Z ERROR rig: Failed to send request to Ollama /api/embed: error sending request for url (http://localhost:11434/api/embed)
Error: HttpError: error sending request for url (http://localhost:11434/api/embed)

Caused by:
    0: error sending request for url (http://localhost:11434/api/embed)
    1: client error (Connect)
    2: tcp connect error
    3: Connection refused (os error 111)
```
```sh
2025-09-18T16:15:11.846743Z ERROR rig: Ollama returned error status 404 Not Found with body: {"error":"model \"llama3.2 \" not found, try pulling it first"}
Error: ProviderError: {"error":"model \"llama3.2 \" not found, try pulling it first"}
```

PR #830, opened 2025-09-17
Refactor/non exhaustive  for completion

hghalebi:
# Implement builder pattern for non-exhaustive Reasoning struct

Refactor the Reasoning struct to use the non-exhaustive pattern, preventing
breaking changes when adding new fields in future versions. This change 
introduces ReasoningBuilder as the primary way to construct Reasoning instances.

## Modified components:
- rig-core/src/completion/message.rs: Updated to use ReasoningBuilder

# Motivation:
The #[non_exhaustive] attribute ensures that adding fields to Reasoning
won't break existing code in dependent crates. The builder pattern provides
a clean, chainable API for constructing complex Reasoning instances. **and most importantly, a way to set default values for new configs in the future without breaking downstream old code with previous configs**

# Migration guide:
Before: Reasoning { field1: value1, field2: value2 } or Reasoning::new(value1)
After:  ReasoningBuilder::new().field1(value1).field2(value2).build()

### DIFF
--- a/rig-core/src/completion/message.rs
+++ b/rig-core/src/completion/message.rs
@@ -92,9 +92,11 @@ impl Reasoning {
         self
     }
 
-    pub fn multi(mut self, input: Vec<String>) -> Self {
-        self.reasoning = input;
-        self
+    pub fn multi(input: Vec<String>) -> Self {
+        Self {
+            id: None,
+            reasoning: input,
+        }
     }
 }
 


rig-core/src/completion/message.rs
```diff
@@ -68,36 +68,63 @@ pub enum AssistantContent {
...
 }
 
 impl Reasoning {
     /// Create a new reasoning item from a single item
-    pub fn new(input: &str) -> Self {
-        Self {
-            id: None,
-            reasoning: vec![input.to_string()],
-        }
+    pub fn new(input: &str) -> ReasoningBuilder {
```
joshua-mo-143: This method should still return `Self`.

The normal convention in Rust for creating a builder is using `fn builder()` (and then handling all inputs from within the builder's methods)

rig-core/src/completion/message.rs
```diff
@@ -68,36 +68,63 @@ pub enum AssistantContent {
...
+        
     }
+    // TODO: Deprecate this signature in favor of a parameterless new()
+  
 
-    pub fn multi(input: Vec<String>) -> Self {
+}
+
+#[derive(Clone, Debug, Default)]
+pub struct ReasoningBuilder{
```
joshua-mo-143: This needs to be `#[non_exhaustive]`.

You also need to use `cargo fmt` - it looks like there are some minor formatting issues

rig-core/src/completion/message.rs
```diff
@@ -68,36 +68,63 @@ pub enum AssistantContent {
...
     }
-
+    
+    pub fn multi(mut self,input: Vec<String>) -> Self {
+        
+        self.reasoning = input;
+        self
+       
+    }
+    pub fn add_step(mut self,step: impl Into<String>)->Self{
```
joshua-mo-143: I think we can get rid of this method for now, in regular usage I don't think it is super likely that users need to retain the same reasoning builder and add a step to it.

joshua-mo-143 (changes requested):

rig-core/src/completion/message.rs
hghalebi: If it returns `Self` but not a `builder struct`, what is the point of using the builder pattern? The idea was that if we change the `Reasining` struct with new configs, the builder struct could handle the default config in a way that the old code still works. If we have a new method with parameters, what is the point of using non-executive tag?

hghalebi:
We may need to refactor the OpenAI response api also to adapt to this change:
```
error[E0599]: no function or associated item named `multi` found for struct `completion::message::Reasoning` in the current scope
   --> rig-core/src/providers/openai/responses_api/mod.rs:913:41
    |
913 |                     message::Reasoning::multi(summary).with_id(id),
    |                                         ^^^^^ function or associated item not found in `completion::message::Reasoning`
    |
   ::: rig-core/src/completion/message.rs:72:1
    |
72  | pub struct Reasoning {
    | -------------------- function or associated item `multi` not found for this struct

```

hghalebi:
> We may need to refactor the OpenAI response api also to adapt to this change:
> 
> ```
> error[E0599]: no function or associated item named `multi` found for struct `completion::message::Reasoning` in the current scope
>    --> rig-core/src/providers/openai/responses_api/mod.rs:913:41
>     |
> 913 |                     message::Reasoning::multi(summary).with_id(id),
>     |                                         ^^^^^ function or associated item not found in `completion::message::Reasoning`
>     |
>    ::: rig-core/src/completion/message.rs:72:1
>     |
> 72  | pub struct Reasoning {
>     | -------------------- function or associated item `multi` not found for this struct
> ```

also :

```
error[E0308]: mismatched types
   --> rig-core/src/providers/anthropic/completion.rs:463:17
    |
462 |             } => message::AssistantContent::Reasoning(
    |                  ------------------------------------ arguments to this enum variant are incorrect
463 |                 Reasoning::new(&thinking).optional_id(signature),
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `Reasoning`, found `ReasoningBuilder`
    |
note: tuple variant defined here
   --> rig-core/src/completion/message.rs:67:5
    |
67  |     Reasoning(Reasoning),
    |     ^^^^^^^^^

error[E0308]: mismatched types
   --> rig-core/src/providers/gemini/completion.rs:251:69
    |
251 | ...                   completion::AssistantContent::Reasoning(Reasoning::new(text))
    |                       --------------------------------------- ^^^^^^^^^^^^^^^^^^^^ expected `Reasoning`, found `ReasoningBuilder`
    |                       |
    |                       arguments to this enum variant are incorrect
    |
note: tuple variant defined here
   --> rig-core/src/completion/message.rs:67:5
    |
67  |     Reasoning(Reasoning),
    |     ^^^^^^^^^

error[E0308]: mismatched types
   --> rig-core/src/providers/gemini/completion.rs:481:45
    |
480 | ...                   Some(true) => message::AssistantContent::Reasoning(
    |                                     ------------------------------------ arguments to this enum variant are incorrect
481 | ...                       Reasoning::new(&text),
    |                           ^^^^^^^^^^^^^^^^^^^^^ expected `Reasoning`, found `ReasoningBuilder`
    |
note: tuple variant defined here
   --> rig-core/src/completion/message.rs:67:5
    |
67  |     Reasoning(Reasoning),
    |     ^^^^^^^^^

error[E0599]: no function or associated item named `multi` found for struct `message::Reasoning` in the current scope
   --> rig-core/src/providers/openai/responses_api/mod.rs:913:41
    |
913 |                     message::Reasoning::multi(summary).with_id(id),
    |                                         ^^^^^ function or associated item not found in `message::Reasoning`
    |
   ::: rig-core/src/completion/message.rs:72:1
    |
72  | pub struct Reasoning {
```

PR #834, opened 2025-09-18
chore(rig-949): pin CI deps

FayCarsons:


joshua-mo-143 (approved):
lgtm - just need to set the env in settings and can merge

PR #846, opened 2025-09-22
feat(rig-912): rework `Chat` trait for multi-turn

FayCarsons:


rig-core/src/completion/chat.rs
```diff
@@ -0,0 +1,207 @@
...
+    -> Result<String, PromptError>;
+
+    fn show_usage(&self) -> bool {
+        false
+    }
+
+    fn usage(&self) -> Option<Usage> {
+        None
+    }
+}
```
FayCarsons, line 43: The thought was that the CLI chat loop doesn't need to know how an agent/Chat impl gets responses or prints them, it just needs to do that and then push the final response to history. 

So if we abstract that away we can have a single `run` function which should be easier to maintain.

rig-core/src/completion/chat.rs
```diff
@@ -0,0 +1,207 @@
...
+pub struct ChatImpl<T: Chat>(T);
+
+pub struct AgentImpl<M: CompletionModel + 'static> {
+    agent: Agent<M>,
+    multi_turn_depth: usize,
+    show_usage: bool,
+    // This *could* hold total usage?
+    // Currently it's per-request but maybe that's not what we want
+    usage: Usage,
+}
```
FayCarsons: Realized I don't know if we want running usage or just per-request usage here, went with per-request

FayCarsons:
I made a new file `rig-core/completion/chat.rs` because it felt like this didn't belong in `request.rs` with the `Chat` trait, if anything because that module already has a lot going on. But idk if that's necessary or how you're deciding when to break things out into new files.

0xMochan:
A couple of thoughts, not a formal review (tho I can do that later):

The current `Chat` trait might be worthwhile to rethink from the ground up. Previously, the `Chat` trait was the main way to handle message history but this has been rethought via the `.prompt()` typestate pattern which lets users control message history (via `mut` borrow).

So now it's sorta in a weird state where it likely wants to be designed for chatbots (beyond just CLIs) but doesn't actual do much beyond that. I like how the traits `CliChat` and `ChatImpl` helps abstract out how the agent works which is compatible with the current setup.

Essentially, I'm trying to say if something in the `Chat` trait doesn't make sense, it's def. a good idea to question it and re-think a better design if it's an approach you are interested in exploring.

---

Other than that, should the `CliChat` be apart of the `completions/chat.rs` or should it be in the `cli_chatbot.rs`? Unsure but I assumed that `completions/chat.rs` would be for all chat-like things, open question :).

FayCarsons:
We discussed this a bit in the Rig sync today. It seems like a nice way to compose agents, I know this would have made working on i.e. the multi-agent example a lot easier. Maybe it's worth keeping in `completions/chat` and building out to handle more general chatting which a CLI is an instance of? Though I'm not entirely sure what that would look like, definitely interested though

rig-core/src/completion/chat.rs
```diff
@@ -0,0 +1,203 @@
...
+    streaming::{StreamedAssistantContent, StreamingPrompt},
+};
+use futures::StreamExt;
+use std::io::{self, Write};
+
+pub struct NoImplProvided;
+
+pub struct ChatImpl<T: Chat>(T);
+
+pub struct AgentImpl<M: CompletionModel + 'static> {
```
joshua-mo-143: please use full syntax (eg, `where M: CompletionModel + 'static`)

rig-core/src/cli_chatbot.rs
```diff
@@ -0,0 +1,203 @@
...
+    ) -> ChatBotBuilder<AgentImpl<M>> {
+        ChatBotBuilder(AgentImpl {
+            agent,
+            multi_turn_depth: 1,
+            show_usage: false,
+            usage: Usage::default(),
+        })
+    }
+
+    pub fn chat<T: Chat>(self, chatbot: T) -> ChatBotBuilder<ChatImpl<T>> {
```
joshua-mo-143, line 136: please use full syntax for trait bounds (`where T: Chat`)

rig-core/src/completion/chat.rs
```diff
@@ -0,0 +1,203 @@
...
+            usage: Usage::default(),
+        })
+    }
+
+    pub fn chat<T: Chat>(self, chatbot: T) -> ChatBotBuilder<ChatImpl<T>> {
+        ChatBotBuilder(ChatImpl(chatbot))
+    }
+}
+
+impl<T: Chat> ChatBotBuilder<ChatImpl<T>> {
```
joshua-mo-143: please use full syntax for trait bounds (where T: Chat)

rig-core/src/completion/chat.rs
```diff
@@ -0,0 +1,203 @@
...
+    pub fn build(self) -> ChatBot<ChatImpl<T>> {
+        let ChatBotBuilder(chat_impl) = self;
+        ChatBot(chat_impl)
+    }
+}
+
+impl<M: CompletionModel + 'static> ChatBotBuilder<AgentImpl<M>> {
```
joshua-mo-143: please use full syntax for trait bounds (`where M: CompletionModel + 'static`)

rig-core/src/completion/chat.rs
```diff
@@ -0,0 +1,203 @@
...
+        })
+    }
+
+    pub fn build(self) -> ChatBot<AgentImpl<M>> {
+        ChatBot(self.0)
+    }
+}
+
+#[allow(private_bounds)]
+impl<T: CliChat> ChatBot<T> {
```
joshua-mo-143: please use full syntax for trait bounds (`where T: CliChat`)

rig-core/src/completion/chat.rs
```diff
@@ -0,0 +1,203 @@
...
+        history: Vec<Message>,
+    ) -> Result<String, PromptError> {
+        let res = self.0.chat(prompt, history).await?;
+        println!("{res}");
+
+        Ok(res)
+    }
+}
+
+impl<M: CompletionModel + 'static> CliChat for AgentImpl<M> {
```
joshua-mo-143: please use full syntax for trait bounds

PR #874, opened 2025-09-29
bug(rig-959): Documents in Huggingface are not converted properly

### LINKED ISSUE #852: bug: Documents in Huggingface are not converted properly

## Bug Report

A follow up to #848. We should make sure that documents are converted to prompt texts or are otherwise somehow included in the request.

FayCarsons:
Fixes #852

### DIFF
--- a/rig-core/src/providers/huggingface/completion.rs
+++ b/rig-core/src/providers/huggingface/completion.rs
@@ -322,6 +322,11 @@ impl TryFrom<message::Message> for Vec<Message> {
                                 let text = String::from_utf8_lossy(raw.as_slice()).into();
                                 Ok(UserContent::Text { text })
                             }
+                            message::UserContent::Document(message::Document {
+                                data: message::DocumentSourceKind::Base64(text), ..
+                            }) => {
+                                Ok(UserContent::Text { text })
+                            }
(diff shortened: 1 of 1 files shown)


joshua-mo-143:
It looks like we need to add a `StringLiteral` or similar variant to `DocumentSourceKind`. 

`UserContent::document()` automatically assumes the content is base64-encoded. This is an issue because `CompletionRequest::normalized_documents()` automatically tries to convert every document into a base64 encoded string document, which can confuse the model.

rig-core/src/providers/huggingface/completion.rs
```diff
@@ -305,6 +316,12 @@ impl TryFrom<message::Message> for Vec<Message> {
                                     image_url: ImageUrl { url },
                                 })
                             }
+                            message::UserContent::Document(message::Document {
+                                data: message::DocumentSourceKind::Raw(raw), ..
+                            }) => {
+                                let text = String::from_utf8_lossy(raw.as_slice()).into();
+                                Ok(UserContent::Text { text })
+                            }
```
joshua-mo-143, line 324: document snippet needs adding here too

joshua-mo-143 (changes requested):
1 change required then merge

PR #875, opened 2025-09-29
feat(rig-951): generic HTTP client

FayCarsons:
I think I got this all worked out now that I'm using types from the `http` crate, and using return-position impls in trait methods vs the `async` keyword to enforce `Send` return types, but I figured I'd still put it up here.

Not sure that `Bytes` is the right type to bound request and response bodies by. For instance, `String` does not `impl From<Bytes>` so it's sort of a pain to get the text out of a response. The `Arc` overhead is less than ideal as well. Same with the error handling, we have a `Box<dyn Error>` which I think is generally suboptimal.

But it is, at the very least, typechecking for the anthropic provider

### DIFF
--- a/rig-core/src/completion/chat.rs
+++ b/rig-core/src/completion/chat.rs
@@ -1,0 +1,203 @@
+use crate::{
+    agent::{Agent, MultiTurnStreamItem, Text},
+    completion::{Chat, CompletionError, CompletionModel, PromptError, Usage},
+    message::Message,
+    streaming::{StreamedAssistantContent, StreamingPrompt},
+};
+use futures::StreamExt;
+use std::io::{self, Write};
+
+pub struct NoImplProvided;
+
+pub struct ChatImpl<T: Chat>(T);
+
+pub struct AgentImpl<M: CompletionModel + 'static> {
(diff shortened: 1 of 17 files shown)


rig-core/src/completion/chat.rs
```diff
@@ -0,0 +1,203 @@
+use crate::{
```
joshua-mo-143, line 1: is this file needed? seems to be an old artifact from the chatbot PR

FayCarsons: Didn't catch that, will remove

Sytten:
This broke a key concept that the client should not be created by default if the user is using a custom client, this is wasteful.
There was a reason it was an Option...........
Also creating a client should be a Result if we ever want to add other errors during client building.

joshua-mo-143:
> This broke a key concept that the client should not be created by default if the user is using a custom client, this is wasteful.
> There was a reason it was an Option...........
> Also creating a client should be a Result if we ever want to add other errors during client building.

Hi, this was unfortunately a bit of an oversight that we made while doing the changes as it was a large change. I'll go back over it and make amendments.

PR #880, opened 2025-10-01
fix: Correct data structure for OpenAI responses images and PDFs

### LINKED ISSUE #879: bug: Input image gets serialized incorrectly with OpenAI responses API

## Bug Report

The `InputImage` data shape is wrong for the OpenAI responses API. The OpenAI Responses API expects the `image_url` property to be a string (with all properties flattened on the input content) unlike the completions API.

Correct shape should look like:
```rs
InputImage {
    image_url: String,
    #[serde(default)]
    detail: ImageDetail,
},
```

## Reproduction

```rs
let message = Message::User {
  content: OneOrMany::one(
    UserContent::Image(Image {
        data: DocumentSourceKind::Base64(data),
        media_type: Some(ImageMediaType::PNG),
        detail: Some(ImageDetail::Auto),
        additional_params: None,
    })
  )
};

// Do completion/prompt request with above message
```

You'll get the API error:

```json
{
  "error": {
    "message": "Invalid type for 'input[2].content[0].image_url': expected an image URL, but got an object instead.",
    "type": "invalid_request_error",
    "param": "input[2].content[0].image_url",
    "code": "invalid_type"
  }
}

```

## Expected behavior

The image should be passed through to OpenAI correctly and respond with context from the image.

z0w0:
Updates the way `image_url` is sent to OpenAI for the responses API. The image properties are now flat on the `input_image` in the new API.

Also sends PDFs as `input_file` instead of input text, for better reliability/vision capabilities.

Fixes #879

### DIFF
--- a/rig-core/src/providers/openai/responses_api/mod.rs
+++ b/rig-core/src/providers/openai/responses_api/mod.rs
@@ -8,10 +8,13 @@
 //! let model = openai_client.completion_model("gpt-4o").completions_api();
 //! ```
 use super::{Client, responses_api::streaming::StreamingCompletionResponse};
-use super::{ImageUrl, InputAudio, SystemContent};
+use super::{InputAudio, SystemContent};
 use crate::completion::CompletionError;
 use crate::json_utils;
-use crate::message::{AudioMediaType, Document, DocumentSourceKind, MessageError, MimeType, Text};
+use crate::message::{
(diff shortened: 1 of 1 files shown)


joshua-mo-143:
Thank you for this PR! I am currently travelling but will have a look as soon as I am available

joshua-mo-143 (approved):
lgtm

PR #903, opened 2025-10-07
feat: return usage when streaming completions from a dynamic client

### LINKED ISSUE #902: feat: Return usage metadata when using `CompletionModelDyn::stream`

## Feature Request

Some usage metadata should be returned when using stream completions with a dynamic client.

### Motivation

Currently, `CompletionModelDyn::stream` returns a `StreamingCompletionResponse<()>`. This prohibits the use of the dynamic client API if getting usage information is desired, as none is contained within the `()` given in the final value of the stream.

### Proposal

The trait `CompletionModelDyn` should be altered so that `CompletionModelDyn::stream` now returns something like `StreamingCompletionResponse<FinalCompletionResponse>`, where `FinalCompletionResponse` looks like:

```rust
pub struct FinalCompletionResponse {
    pub usage: Option<Usage>,
}
```

Additionnally, `CompletionModelHandle`'s implementation of `CompletionModel` needs to be altered too, with `type StreamingResponse = FinalCompletionResponse`.

I will link a PR for reference.

### Alternatives

One could just avoid using the dynamic client API and instead maintain an enum of all the clients they care about, but, in my opinion, that would defeat the purpose of such a nice API existing.

lefebvreb:
This is the implementation of my idea for providing the user with usage metadata when streaming completions using the dynamic client API, explained in #902.

Struct names and locations are very much up for debate, as I am not familiar with the project.

Fixes #902.

### DIFF
--- a/rig-core/src/completion/request.rs
+++ b/rig-core/src/completion/request.rs
@@ -364,7 +364,10 @@ pub trait CompletionModelDyn: WasmCompatSend + WasmCompatSync {
     fn stream(
         &self,
         request: CompletionRequest,
-    ) -> WasmBoxedFuture<'_, Result<StreamingCompletionResponse<()>, CompletionError>>;
+    ) -> WasmBoxedFuture<
+        '_,
+        Result<StreamingCompletionResponse<FinalCompletionResponse>, CompletionError>,
+    >;
 
     fn completion_request(
         &self,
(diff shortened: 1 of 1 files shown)


rig-core/src/client/completion.rs
```diff
@@ -59,6 +59,18 @@ pub trait CompletionClient: ProviderClient + Clone {
     }
 }
 
+/// The final streaming response from a dynamic client.
+#[derive(Debug, Deserialize, Clone, Serialize)]
+pub struct FinalCompletionResponse {
```
joshua-mo-143: This is probably better placed in `client/builder.rs` as it's likely only really going to be used for the dynamic client builder

joshua-mo-143 (changes requested):
One change required, other than that this actually mostly looks decent (pending CI pass of course).

rig-core/src/client/completion.rs
lefebvreb: Makes sense! Should be good now.

lefebvreb:
Last merge had minor conflicts with my changes, but it should be resolved now.

Issue #760, opened 2025-08-30
feat: rework in-memory vector store to use Approx. Nearest Neighbour (ANN)


joshua-mo-143:
- [x] I have looked for existing issues (including closed) about this

## Feature Request

Currently, the in-memory vector store is a bit weak. We could really do with beefing it up. At the moment, the search is basically just brute-force cosine similarity searching for each and every single vector... which is really slow. By using methods that can calculate the ANN (Approximate Nearest Neighbour), we can add significant speedups to our implementation.

However to avoid adding in many more dependencies and ensure it can compile to WASM, we will likely need to write the algorithm from scratch. This is a non-trivial endeavor and as such we will need to ensure we have a way of benchmarking the old implementation vs the new implementation.

To start with, we should probably try implementing Locality Sensitive Hashing (LSH) first as it's the most simple to use.

I'll probably end up taking this ticket shortly if nobody else claims it because it's a non-trivial ticket that I would rather see through to completion myself.

### Motivation

Performance speedup/optimisation

### Proposal

Implement LSH and use it in the vector store for cosine similarity search.

Dev79844:
Hey @joshua-mo-143 
I would like to take this up.

joshua-mo-143:
> Hey [@joshua-mo-143](https://github.com/joshua-mo-143) I would like to take this up.

Assigned - as a side note for the impl, you may find using `fastrand` to be better than `rand` for the number generation as it's lower dependency size and can still generate random numbers reliably

Dev79844:
Hey @joshua-mo-143 
I was thinking to write the code in such a way that the algorithm is configurable. there will be search strategies, and user can configure what strategy should be used. this way multiple algorithms can be supported for building the ann index. What do you say?

joshua-mo-143:
> Hey @joshua-mo-143 
> I was thinking to write the code in such a way that the algorithm is configurable. there will be search strategies, and user can configure what strategy should be used. this way multiple algorithms can be supported for building the ann index. What do you say?

Seems OK to me.

However I would be wary of a couple things:
1) Rig still needs to be able to compile to WASM which means adding some dependencies might not work and I would ensure this gets tested before you start writing any code (by adding the dependencies then trying to compile to wasm32 target arch) 
2) Rig itself is intended to be a fairly lean library, so if you end up adding any heavy dependencies you will probably want to feature gate them

joshua-mo-143:
[...]

PR #922, opened 2025-10-13
feat: LSH for in memory vector store

Dev79844:
Fixes: #760

- Added support for LSH using random projections approach
- Modified vector store methods for configuring index strategy

### DIFF
--- a/rig-core/src/vector_store/builder.rs
+++ b/rig-core/src/vector_store/builder.rs
@@ -6,18 +6,30 @@ use crate::{OneOrMany, embeddings::Embedding};
 use super::{IndexStrategy, in_memory_store::InMemoryVectorStore};
 
 /// Builder for creating an [InMemoryVectorStore] with custom configuration.
-pub struct InMemoryVectorStoreBuilder<D: Serialize> {
+pub struct InMemoryVectorStoreBuilder<D>
+where
+    D: Serialize,
+{
+    /// Embeddings of the documents.
     embeddings: HashMap<String, (D, OneOrMany<Embedding>)>,
+
+    /// Index strategy for the vector store.
     index_strategy: IndexStrategy,
 }
 
-impl<D: Serialize + Eq> Default for InMemoryVectorStoreBuilder<D> {
+impl<D> Default for InMemoryVectorStoreBuilder<D>
+where
+    D: Serialize + Eq,
+{
     fn default() -> Self {
         Self::new()
     }
 }
 
-impl<D: Serialize + Eq> InMemoryVectorStoreBuilder<D> {
+impl<D> InMemoryVectorStoreBuilder<D>
+where
+    D: Serialize + Eq,
+{
     /// Create a new builder with default settings.
     /// Default index strategy is BruteForce.
     pub fn new() -> Self {
--- a/rig-core/src/vector_store/lsh.rs
+++ b/rig-core/src/vector_store/lsh.rs
@@ -1,7 +1,10 @@
 use fastrand::Rng;
 use std::collections::HashMap;
 
-/// Random Projection LSH
+/// Locality Sensitive Hashing (LSH) with random projection.
+/// Uses random hyperplanes to hash similar vectors into the same buckets for efficient
+/// approximate nearest neighbor search. See <https://www.pinecone.io/learn/series/faiss/locality-sensitive-hashing-random-projection/>
+/// for details on how LSH works.
 #[derive(Clone, Default)]
 pub struct LSH {
     hyperplanes: Vec<Vec<f32>>,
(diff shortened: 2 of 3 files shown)


joshua-mo-143:
Hi @Dev79844, great work on this! I'll be reviewing sometime during the week.

joshua-mo-143 (changes requested):
Some general comments:
- Overall, pretty good work.
- I think we need a builder for `InMemoryVectorStore` and provide `IndexStrategy::BruteForce` as the default (which will prevent breaking changes from needing to adjust the current `from` methods).
- I am not hugely sold on the enum because it causes strange implementation behaviour (like having to fall back to brute forcing in `vector_search_lsh` when there's no LSH index), but as it's mostly internal implementation we can figure this out later on. `InMemoryVectorStore` is not really supposed to be used for huge production stuff anyway.

Dev79844:
Hey @joshua-mo-143!
After adding builder, should `index_strategy` argument be removed from the `from` methods?

joshua-mo-143:
> Hey @joshua-mo-143! After adding builder, should `index_strategy` argument be removed from the `from` methods?

Yeah index strategy should be removed.

The idea is that we provide good defaults for users to fall back to, and then when they want to use the higher-power options they can switch to it (so in this case we'd default to `IndexStrategy::BruteForce` to prevent breaking changes in behaviour).

rig-core/src/vector_store/builder.rs
```diff
@@ -0,0 +1,94 @@
+use serde::Serialize;
+use std::collections::HashMap;
+
+use crate::{OneOrMany, embeddings::Embedding};
+
+use super::{IndexStrategy, in_memory_store::InMemoryVectorStore};
+
+/// Builder for creating an [InMemoryVectorStore] with custom configuration.
+pub struct InMemoryVectorStoreBuilder<D: Serialize> {
```
joshua-mo-143: Please use full syntax (eg, using `where` clause)

rig-core/src/vector_store/builder.rs
```diff
@@ -0,0 +1,94 @@
...
+    embeddings: HashMap<String, (D, OneOrMany<Embedding>)>,
+    index_strategy: IndexStrategy,
+}
+
+impl<D: Serialize + Eq> Default for InMemoryVectorStoreBuilder<D> {
```
joshua-mo-143: Please use full syntax (eg, using `where` clause)

rig-core/src/vector_store/builder.rs
```diff
@@ -0,0 +1,94 @@
...
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl<D: Serialize + Eq> InMemoryVectorStoreBuilder<D> {
```
joshua-mo-143: Please use full syntax (eg, using `where` clause)

rig-core/src/vector_store/lsh.rs
```diff
@@ -0,0 +1,122 @@
...
+        // Collect candidates from all tables
+        for table_idx in 0..self.lsh.num_tables {
+            let hash = self.lsh.hash(embedding, table_idx);
+
+            if let Some(ids) = self.tables[table_idx].get(&hash) {
+                candidates.extend(ids.iter().cloned());
+            }
+        }
+
+        dbg!(&candidates);
```
joshua-mo-143: Please remove

rig-core/src/vector_store/lsh.rs
```diff
@@ -0,0 +1,122 @@
...
+            if dot >= 0.0 {
+                hash |= 1u64 << i;
+            }
+        }
+
+        hash
+    }
+}
+
+/// LSH Index for document IDs
```
joshua-mo-143: Please add a docs link explaining what LSH is and how it works

rig-core/src/vector_store/lsh.rs
```diff
@@ -0,0 +1,122 @@
+use fastrand::Rng;
+use std::collections::HashMap;
+
+/// Random Projection LSH
```
joshua-mo-143: Please add a docs link explaining what LSH is and how it works

rig-core/src/vector_store/lsh.rs
```diff
@@ -0,0 +1,122 @@
...
+    num_tables: usize,
+    num_hyperplanes: usize,
+}
+
+impl LSH {
+    pub fn new(dim: usize, num_tables: usize, num_hyperplanes: usize) -> Self {
+        let mut rng = Rng::new();
+        let mut hyperplanes = Vec::new();
+
+        // Generate random hyperplanes for all tables
```
joshua-mo-143: Please delete this comment (it's unnecessary since anyone working with LSH knows that you basically need to generate random hyperplanes anyway)

rig-core/src/vector_store/lsh.rs
```diff
@@ -0,0 +1,122 @@
...
+        for _ in 0..(num_tables * num_hyperplanes) {
+            let mut plane = vec![0.0; dim];
+
+            // Generate random values in [-1, 1]
```
joshua-mo-143: If you're writing a comment here it should be why the code is doing what it does and not what.

It's not obvious that the reason why the values are generated between `-1` and `1` is for uniformity across the hyperplanes

rig-core/src/vector_store/lsh.rs
```diff
@@ -0,0 +1,122 @@
...
+            for val in plane.iter_mut() {
+                *val = rng.f32() * 2.0 - 1.0;
+            }
+
+            // Normalize to unit vector
```
joshua-mo-143: Please providing reasoning here rather than just telling us what the code is doing

rig-core/src/vector_store/mod.rs
```diff
@@ -213,3 +215,20 @@ where
             .collect())
     }
 }
+
+// enum for index strategy
```
joshua-mo-143: nit: Should be "The indexing strategy used for [`super::InMemoryVectorStore`]." or something to that effect

PR #923, opened 2025-10-13
feat(rig-990): allow configuring optional lancedb features

### LINKED ISSUE #921: Allow configuring optional lancedb features  to rig-lancedb
Description: Currently, rig-lancedb depends on lancedb with its default features enabled. This introduces an unwanted dependency on openssl, which complicates cross-compilation workflows. Our project uses rustls for TLS support and does not require openssl.

Expected Behavior: We would like rig-lancedb to either:

Disable default-features when depending on lancedb, or

Expose a way for downstream users to configure lancedb features explicitly.

Suggested Change: In Cargo.toml, update the dependency to something like:

```toml
lancedb = { version = "x.y", default-features = false, features = ["desired-feature"] }
```
Alternatively, provide a feature flag in rig-lancedb that controls whether default-features are enabled.

Environment:

Target: Cross-compilation (e.g., x86_64-unknown-linux-musl)

TLS: rustls

Current rig-lancedb version: x.y.z

Additional Context: Managing openssl in cross-compilation environments is error-prone and unnecessary for users relying on rustls. Making lancedb features configurable would improve flexibility and compatibility.

joshua-mo-143:
Fixes #921.

@cody-why Please let me know if this fixes your issue!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -215,15 +215,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "argminmax"
-version = "0.6.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "70f13d10a41ac8d2ec79ee34178d61e6f47a29c2edfe7ef1721c7383b0359e65"
-dependencies = [
- "num-traits",
-]
-
-[[package]]
 name = "argon2"
 version = "0.5.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


cody-why:
Thank you for submitting this issue!

Unfortunately,  The openssl dependency you're encountering actually comes from another crate: rig-core. In your setup, if you're using a local path override like:
```Toml
rig-core = { path = "../rig-core", version = "0.21.0" }
```

you'll still inherit its default features, which include reqwest with openssl.

To resolve this, you should explicitly disable default features and enable reqwest-rustls like so:
```
rig-core = { version = "0.21", features = ["reqwest-rustls"], default-features = false }
```

joshua-mo-143:
Ok I think this should work now

PR #947, opened 2025-10-17
feat(gemini): Support streaming thinking

scrogson:
The key change in rig-core/src/providers/gemini/streaming.rs:145-177:
  - Before: Only processed the first part using choice.content.parts.first()
  - After: Iterates through all parts using for part in &choice.content.parts

This allows Gemini streaming responses to properly handle multiple parts in a single chunk
(e.g., multiple text chunks, multiple tool calls, or mixed content).

### DIFF
--- a/rig-core/src/providers/gemini/streaming.rs
+++ b/rig-core/src/providers/gemini/streaming.rs
@@ -119,7 +119,7 @@ impl CompletionModel<reqwest::Client> {
             while let Some(event_result) = event_source.next().await {
                 match event_result {
                     Ok(Event::Open) => {
-                        tracing::trace!("SSE connection opened");
+                        tracing::debug!("SSE connection opened");
                         continue;
                     }
                     Ok(Event::Message(message)) => {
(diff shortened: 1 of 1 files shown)


rig-core/src/providers/gemini/streaming.rs
```diff
@@ -142,37 +142,42 @@ impl CompletionModel<reqwest::Client> {
...
+                                },
+                                part => {
+                                    tracing::warn!(?part, "Unsupported response type with streaming");
+                                }
                             }
-                            None => tracing::trace!(reason = ?choice.finish_reason, "There is no part in the streaming content"),
+                        }
+
+                        if choice.content.parts.is_empty() {
+                            tracing::trace!(reason = ?choice.finish_reason, "There is no part in the streaming content");
```
joshua-mo-143, line 180: I am wondering if this should be a `debug` event since we don't really use trace-level messages in Rig.

joshua-mo-143 (approved):
lgtm

PR #952, opened 2025-10-21
feat(rig-976): support filters for `VectorSearchRequest`

### LINKED ISSUE #890: feat: support filters for `VectorSearchRequest`

## Feature Request
We need a way to support generic filtering for fields in `VectorSearchRequest` (for example, searching for keys where the TTL might have expired).

Required operations:
- `eq`
- `gt`
- `lt`
- `and`
- `or`

Once implemented, we then need to adapt it to each vector store provider's way of filtering.

### Motivation

This is blocking #373 

### Implementation
Something like the following:
```rust
enum FilterExpr {
    /// Equal to
    Eq(String, serde_json::String),
    /// Greater than
    Gt(String, i64),
    /// Less than
    Lt(String, i64),
    /// And combinator (must satisfy all filters)
    And(FilterExpr, FilterExpr),
    /// Or combinator (must satisfy one of the filters)
    Or(FilterExpr, FilterExpr)
}

### PR DESCRIPTION
**Pull request by FayCarsons**
Fixes #890

### CODE REVIEW

File: rig-core/src/vector_store/request.rs [hunk 1]
```diff
@@ -34,18 +36,177 @@ impl VectorSearchRequest {
...
+pub trait SearchFilter {
+    type Value;
+
+    fn eq(key: String, value: Self::Value) -> Self;
+    fn gt(key: String, value: Self::Value) -> Self;
+    fn lt(key: String, value: Self::Value) -> Self;
+    fn and(self, rhs: Self) -> Self;
+    fn or(self, rhs: Self) -> Self;
+    fn not(self) -> Self;
+}
```
FayCarsons, line 82: This is the core of it, a tagless final encoding of filter expressions. This is nice because we can (mostly) define backend-specific filters as i.e. `struct FooFilter(serde_json::Value)` or `struct BarFilter(String)` and produce the final query string/object directly as opposed to having an intermediate sum type. It also allows backend-specific extensions like `exists` or `is_not_null` via adding methods to that backend's filter representation.

rig-core/src/vector_store/request.rs
```diff
@@ -34,18 +36,177 @@ impl VectorSearchRequest {
...
+    }
+
+    fn or(self, rhs: Self) -> Self {
+        Self::Or(self.into(), rhs.into())
+    }
+
+    fn not(self) -> Self {
+        Self::Not(self.into())
+    }
+}
```
FayCarsons, line 125: We also have a canonical representation, mostly for `VectorStoreIndexDyn`, that we can interpret into any other `impl SearchFilter` to cut down on boxing, or serialize for storage.

rig-core/src/tool/server.rs
```diff
@@ -0,0 +1,433 @@
...
+                                    tracing::warn!("Tool implementation not found in toolset: {}", doc);
+                                }
+                            }
+                            Ok(acc)
+                        })
+                        .await
+                        .map_err(|e| CompletionError::RequestError(Box::new(e)))?
+        } else {
+            Vec::new()
+        };
```
FayCarsons, line 191: Here, because our vector store is `dyn VectoreStoreIndexDyn` which stores all filters as the canonical `Filter<serde_json::Value>` we use the `interpret` method to translate it to a concrete filter type at runtime.

rig-lancedb/src/lib.rs
```diff
@@ -114,6 +117,65 @@ pub enum SearchType {
...
+        Array(xs) => format!(
+            "({})",
+            xs.into_iter()
+                .map(escape_value)
+                .collect::<Vec<_>>()
+                .join(", ")
+        ),
+        // FIXME: This means we should use i.e. a LanceDbValue or similar to prevent users from
+        // constructing objects
+        Object(_) => panic!("Objects are not supported in LanceDB filter expression"),
```
FayCarsons: We could create a `enum LanceDBValue` which is essentially `serde_json::Value` w/o `Object` for better type-safety, but that would mean this vector store cannot be used w/ `VectorStoreIndexDyn`.

rig-s3vectors/src/lib.rs
```diff
@@ -21,6 +24,49 @@ pub struct CreateRecord {
...
+
+    fn or(self, rhs: Self) -> Self {
+        Self(document!({ "$or": [ self.0, rhs.0 ]}))
+    }
+
+    fn not(self) -> Self {
+        // FIXME: This is not ideal, we should remove this method from the trait or
+        // put in in another trait
+        panic!("'not' is not available for S3SearchFilter")
+    }
```
FayCarsons: Should probably refactor so that `not` is an extension on the backends that are able to support it?

rig-lancedb/src/lib.rs
joshua-mo-143: I do kind of wonder if it would be better to just return a result here since there are some unsupported operations here and elsewhere... might need to have a think and check what the syntax would look like

FayCarsons: Realized we could also do what the Qdrant impl does, which is build up a `serde_json::Value` filter, and then "compile" it to Qdrant's filter types when making a request. Then we're still compatible with `VectorStoreIndexDyn` but at least get sane error behavior.

I do think it could be worth a second look in the future to figure better compile-time safety guarantees, but maybe not worth it right now

joshua-mo-143: I think let's go with the `serde_json::Value` filter.

As long as we can make it work for all use cases, I think this might be better for now as it unblocks some major feature roadblocks. Once we've figured out how to handle this in a much more sound manner then we can definitely get around to implementation

joshua-mo-143:
Awesome work Fay!

I think at the moment following on from some of the immediate discussion on here we should probably opt for compiling to `serde_json::Value` filters and then once this has been merged in we can probably talk about next steps for this

joshua-mo-143 (changes requested):
Maybe a good idea to not check in the SQLite database file (and additionally include instructions indicating that it needs to be pre-created, if that is the case).

Otherwise, great work. This is hopefully one of the last big updates we'll need do for our vector search integrations for a long while (... at least until we need to implement Graph RAG anyway).

FayCarsons:
@joshua-mo-143 My bad, didn't catch that. Should we create a gitignore so that it doesn't get included in the future?

joshua-mo-143:
> @joshua-mo-143 My bad, didn't catch that. Should we create a gitignore so that it doesn't get included in the future?

Yeah that's fine with me

FayCarsons:
@joshua-mo-143 Not super sure how to mark this as resolved, but added the gitignore and took out the db file

PR #958, opened 2025-10-23
fix(agent): Apply tool_choice to completion request

sinkingsugar:
## Summary

This PR fixes a bug where the `Agent` struct's `tool_choice` field was not being applied to the completion request.

## Changes

- Added `tool_choice` to the completion request builder in `Agent::completion()` method
- Follows the same pattern as other optional fields like `preamble`, `temperature`, and `max_tokens`

## Context

The `Agent` struct includes a `tool_choice` field (line 69) that allows forcing the underlying LLM to use a tool before providing a response. However, this field was never being passed to the completion request builder, meaning it had no effect.

This fix ensures that when an agent is configured with a `tool_choice`, it is properly applied to all completion requests.

### DIFF
--- a/rig-core/src/agent/completion.rs
+++ b/rig-core/src/agent/completion.rs
@@ -112,6 +112,11 @@ where
         } else {
             completion_request
         };
+        let completion_request = if let Some(tool_choice) = &self.tool_choice {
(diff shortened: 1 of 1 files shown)


joshua-mo-143 (approved):
lgtm

PR #964, opened 2025-10-23
Dependent packages no longer force unnecessary features on rig-core

nielsgreen:
(This is actually two pull requests)

- When using a package like `rig-sqlite`, `rig-core` would be forced to pull in reqwest. With this pull request that is no longer the case.

- It also allows using `rig-fastembed` without having to pull in `hf-hub`, while leaving the default behavior unchanged, so `rig-fastembed` can now be used without pulling in networking functionality.

This has the added benefit that it makes rig more usable in scenario's where the programmer wants more control over what networking libraries are used. (Useful for cross compilation, and probably other things.)
Feel free to edit.

### DIFF
--- a/rig-fastembed/Cargo.toml
+++ b/rig-fastembed/Cargo.toml
@@ -20,9 +20,11 @@ anyhow = { workspace = true }
 tokio = { workspace = true, features = ["full"] }
 
 [features]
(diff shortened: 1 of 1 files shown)


joshua-mo-143 (approved):
lgtm, thanks!

PR #973, opened 2025-10-24
fix(rig-1009): Incorrect struct shape (OpenAI)

### LINKED ISSUE #972: bug: Incorrect JSON parsing causing stream events to not be emitted (OpenAI)

## Bug Report
It looks like there is a buffering/deadlock issue when using `agent.stream_chat()`.

Streaming responses through completion models and `agent.stream_prompt()` both work correctly, so this appears to be some sort of issue with deadlocking or buffering as the stream does return the final item.

## Reproduction

```rust
use futures::StreamExt;
use rig::message::{Message, UserContent};
use rig::{OneOrMany, prelude::*};

use rig::providers;
use rig::streaming::StreamingChat;

#[tokio::main]
    // Create OpenAI client
    let client = providers::openai::Client::from_env();

    // Create agent with a single context prompt
    let comedian_agent = client
        .agent("gpt-4o")
        .preamble("You are a comedian here to entertain the user using humour and jokes.")
        .build();

    let vec_message = vec![
        Message::User {
            content: OneOrMany::one(UserContent::Text("Tell me a joke!".into())),
        },
        Message::Assistant {
            id: None,
            content: OneOrMany::one(rig::message::AssistantContent::Text(
                "Why did the chicken cross the road?\n\nTo get to the other side!".into(),
            )),
        },
    ];

    let mut stream = comedian_agent
        .stream_chat("Entertain me!", vec_message)
        .await;

    while let Some(chunk) = stream.next().await {
        println!("Got chunk: {:?}", chunk);
    }

    Ok(())
}
```

Output (`Got content` is a logging thing I added to see where it might be going wrong):
```
Got content
Got chunk: Ok(FinalResponse(FinalResponse { response: "", aggregated_usage: Usage { input_tokens: 57, output_tokens: 21, total_tokens: 78 } }))
```

## Expected behavior

Should be receiving all the chunks

joshua-mo-143:
Fixes #972.

Why this is incorrect and won't parse I have no idea.

### DIFF
--- a/rig-core/examples/agent_stream_chat.rs
+++ b/rig-core/examples/agent_stream_chat.rs
@@ -1,0 +1,32 @@
+use rig::agent::stream_to_stdout;
+use rig::message::Message;
+use rig::prelude::*;
+use rig::streaming::StreamingChat;
+
+use rig::providers;
+
+#[tokio::main]
+async fn main() -> Result<(), anyhow::Error> {
+    // Create OpenAI client
+    let client = providers::openai::Client::from_env();
+
+    // Create agent with a single context prompt
+    let comedian_agent = client
+        .agent("gpt-4o")
+        .preamble("You are a comedian here to entertain the user using humour and jokes.")
+        .build();
+
+    let messages = vec![
+        Message::user("Tell me a joke!"),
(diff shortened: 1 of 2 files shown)


joshua-mo-143:
@cody-why This *should* solve your issue FYI

PR #977, opened 2025-10-24
Add flexible provider with environment-based configuration

Ammar-Alnagar:
The new flex provider enables users to configure API key, base URL,
    and available models through environment variables (FLEX_API_KEY,
    FLEX_BASE_URL, FLEX_MODELS). It includes full support for completion
    and transcription APIs, streaming responses, and a comprehensive
    example to demonstrate usage. The provider integrates with existing
    Rig traits and supports both environment-based and direct client
    creation.


rig-core/examples/flex_example.rs
```diff
@@ -0,0 +1,111 @@
...
+
+use rig::client::{CompletionClient, VerifyClient};
+use rig::completion::CompletionModel;
+use rig::message::Message;
+use rig::providers::flex;
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    // Load environment variables from .env file
+    dotenvy::dotenv().ok();
```
joshua-mo-143: Please do not use `dotenvy`.

We generally don't use or encourage use of `dotenvy` in the rig codebase, and adding it specifically for this one example isn't really something that we can justify adding it for.

rig-core/src/providers/flex.rs
```diff
@@ -0,0 +1,111 @@
+//! Example demonstrating the usage of the flex provider
```
joshua-mo-143: Is this file supposed to be here? It literally just looks like you copy and pasted the example from examples to this file

joshua-mo-143 (changes requested):
Hi Ammar! Some general comments since some of these don't really fit into just one file and concern the PR in general rather than particular lines:

- Did you use AI for this PR? It looks like the entire structure of Cargo.toml has been changed for no real reason.
- In the same vein: is there a reason you've disabled doctest? You haven't justified the reason for doing so
- Why is `src/providers/flex.rs` using `#[tokio::main]`? It should be entirely library code with the exception of tests. Did you accidentally overwrite it or something?
- Please do not use `dotenvy` - we don't encourage using it in this codebase as it is liable to cause issues with accidentally committing a `.env` file to a branch. We generally advise careful usage of `.envrc` (through `direnv`) as it allows directly exporting to your environment variables automatically, rather than needing to add another dependency.

rig-core/src/providers/flex.rs
Ammar-Alnagar: apologies , it seems that there have been some connectivity issues and the latest changes werent included , yes that file  was intedned to be the main implementation

rig-core/examples/flex_example.rs
Ammar-Alnagar: sure thing , i was trying to simulate workflows from other llm systems where uses can define their secrets in the env but you do raise valid points and concerns and i will revert to using the same approach as the rest

Ammar-Alnagar:
> Hi Ammar! Some general comments since some of these don't really fit into just one file and concern the PR in general rather than particular lines:
> 
> * Did you use AI for this PR? It looks like the entire structure of Cargo.toml has been changed for no real reason.
> * In the same vein: is there a reason you've disabled doctest? You haven't justified the reason for doing so
> * Why is `src/providers/flex.rs` using `#[tokio::main]`? It should be entirely library code with the exception of tests. Did you accidentally overwrite it or something?
> * Please do not use `dotenvy` - we don't encourage using it in this codebase as it is liable to cause issues with accidentally committing a `.env` file to a branch. We generally advise careful usage of `.envrc` (through `direnv`) as it allows directly exporting to your environment variables automatically, rather than needing to add another dependency.


1-yes , i used an agent to review my changes and make sure i didn't change anything else besides the files i need , i am not too sure why the Cargo.toml in rig-core has been recognized though.
2-I believe doctest was disabled from the main branch and i did not modify it besides the unintentional reordering.
3-yes that was connectivity issue preventing the orignial correct file from being included in the PR , should be fixed now.
4-done  and dandy , i though it was alright since there was some other implementations of it such as rig-postgres.

rig-core/src/providers/flex.rs
```diff
@@ -0,0 +1,891 @@
...
+///
+/// # Example
+/// ```
+/// // Set environment variable: FLEX_MODELS=gpt-4o,gpt-4-turbo,llama-3
+/// use rig::providers::flex;
+///
+/// let models = flex::get_models_from_env();
+/// println!("{:?}", models); // ["gpt-4o", "gpt-4-turbo", "llama-3"]
+/// ```
+pub fn get_models_from_env() -> Vec<String> {
```
joshua-mo-143, line 895: Is this common convention in Python? I'm kind of curious since it doesn't seem that you would normally need to use this and typically in the case that you were using a third party provider you'd typically just go and look at their model list to find out what you need

Ammar-Alnagar: Its not a universal thing to my knowledge,  but something i picked up when working with 3rd party providers that have constantly changing models like chutes AI , or if you are serving using vllm or custom infrence engines (the later being more common in my case).

joshua-mo-143: Ah I see, okay. Ok I think let's keep this in then

rig-core/src/providers/flex.rs
```diff
@@ -0,0 +1,891 @@
+//! Flexible API client and Rig integration that allows users to configure
```
joshua-mo-143: Needs some additional information up here about that users should use this integration for OpenAI-like APIs (that are not "completely" OpenAI compatible).

joshua-mo-143 (changes requested):
Ok I think just one last change actually, that's it. Pretty much just adding some clarification on when to use the provider since it's not super obvious

rig-core/src/providers/flex.rs
```diff
@@ -0,0 +1,906 @@
...
+                gen_ai.output.messages = tracing::field::Empty,
+            )
+        } else {
+            tracing::Span::current()
+        };
+
+        let body = serde_json::to_vec(&request)?;
+        let req = self
+            .client
+            .req(Method::POST, "/chat/completions")?
```
cvauclair, line 492: If the goal is to be flexible and support any endpoint that is OpenAI-compliant-ish, that path should also be a parameter instead of hard-coded to `/chat/completions`

cvauclair: `/chat/completions` can be the default if not provided

Ammar-Alnagar:
@cvauclair @joshua-mo-143 
Thanks for the review.
After digging deeper into the crates’ docs and source, I found this PR isn’t as necessary as i have initiallythought, so I’ll close it.

cvauclair:
No stress, thank you for looking into this!

kskarthik:
This is not implemented ?! It's useful :)

joshua-mo-143:
> This is not implemented ?! It's useful :)

Being useful doesn't necessarily mean it needs to be added to the library - and that's okay!

With `rig-core` primarily aiming to be a lean library (at least until we split the crates), we want to ensure that realistically only features which will receive real usage will get added. Additionally, pretty much all of the internal traits we use are public - so you can use them whichever way you'd like instead of being tied down by the library.

As far as providers go: with the small team that we have (internally speaking it's literally just me and one other person who is currently part time), it is ideal that we only add new providers when it's obvious that there is added value there. If the provider integration is not 100% required to do whatever the user is doing, then it is not hugely likely that it will be accepted. This also extends to the OpenAI Completions API integration which has a million and one other APIs that also encompass the exact same API request-response specification.

PR #989, opened 2025-10-25
Fix bedrock tool calls with zero arguments

scrogson:
This fixes a bug with bedrock failing to call tools with no arguments.
The stream would EOF after the text streaming and halt.

I also added support for setting `tool_choice`.


scrogson:
Done. I removed some pointless tests which didn't actually test the implementation as I'm not sure how to go about setting up an integration test for the empty tool call args.

joshua-mo-143 (approved):
lgtm

Issue #988, opened 2025-10-25
feat: Usage details for all non-completion operations


joshua-mo-143:
- [x] I have looked for existing issues (including closed) about this

## Feature Request

Well, usage is basically already returned by default for completions. We should probably implement it for the other methods for the purpose of usage tracking.

liamwh:
Is there any progress on this one?

liamwh:
Made PR: #1447

sergiomeneses:
Made PR: #1791

PR #992, opened 2025-10-25
feat: extractor usage details

joshua-mo-143:
Part of #988

### DIFF
--- a/rig-core/src/extractor.rs
+++ b/rig-core/src/extractor.rs
@@ -36,7 +36,7 @@ use serde_json::json;
 
 use crate::{
     agent::{Agent, AgentBuilder, AgentBuilderSimple},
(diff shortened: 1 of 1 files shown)


gold-silver-copper:
Closing this as superseded by #1447.

This PR changed `extract()` / `extract_with_chat_history()` to return an `ExtractResponse<T>`, which would break existing callers. Current `main` already has the same extractor usage-tracking intent in a compatible form:

- `ExtractionResponse<T>`
- `extract_with_usage()`
- `extract_with_chat_history_with_usage()`
- `ExtractionResponse` re-exported from `rig-core`
- provider coverage for extractor usage behavior

I verified the current behavior locally with `cargo test --test openai extractor_usage`. The broader #988 can stay open for other non-completion usage tracking work, but this extractor-specific PR is no longer needed.

PR #1010, opened 2025-10-28
fix: implement Serialize for LanceDBFilter to enable LanceDbVectorIndex usage in dynamic_context

cody-why:
Previously, invoking:
```rust
    client
        .completion_model(&self.openai_model)
        .completions_api()
        .into_agent_builder()
        .temperature(self.temperature)
        .preamble(&self.preamble)
        .dynamic_context(top_k, vector_index)
        .build()
```
would fail with:
```
error[E0277]: the trait bound `LanceDBFilter: serde::Serialize` is not satisfied
    |
    |             .dynamic_context(top_k, vector_index)
    |              ---------------        ^^^^^^^^^^^^ the trait `Serialize` is not implemented for `LanceDBFilter`
    |              |
    |              required by a bound introduced by this call
```

To fix this, we implemented `serde::Serialize` and `serde::Deserialize` for `LanceDBFilter`, allowing it to be properly serialized when passed into the agent builder.

This change enables seamless integration of LanceDbVectorIndex with dynamic context filtering.

update dependencies lancedb v0.22  arrow-array v56

### DIFF
--- a/rig-lancedb/tests/integration_tests.rs
+++ b/rig-lancedb/tests/integration_tests.rs
@@ -5,7 +5,9 @@ use fixture::{Word, as_record_batch, schema, words};
 use lancedb::index::vector::IvfPqIndexBuilder;
 use rig::{
     client::EmbeddingsClient,
+    completion::Prompt,
     embeddings::{EmbeddingModel, EmbeddingsBuilder},
+    prelude::CompletionClient,
     providers::openai,
     vector_store::{VectorStoreIndex, request::VectorSearchRequest},
 };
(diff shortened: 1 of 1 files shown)


cody-why:
The feature feat: add example for vector search local ANN agent has been submitted. 

Also, I'm not entirely sure if .dynamic_context(top_k, vector_store_index) is the optimal usage.

joshua-mo-143:
It doesn't need to be optimal, we just need to be able to compile it to check that it works. Otherwise, it's time consuming to build a minimum reproducible example for debugging.

cody-why:
Got it — I’ve already submitted a test to verify it compiles.

My project does use Agent, so I think dynamic_context is the right approach here. But if there’s a better way to integrate the vector store, I’m open to suggestions!

Issue #1018, opened 2025-10-30
Proposal: Add option to log request body in HTTP client


cody-why:
- [x] I have looked for existing issues (including closed) about this

Hi team 👋

1. Log request body in HTTP client Would it be useful to support logging the request body in the HTTP client? This can be helpful for debugging or tracing payloads, especially in agent workflows. If this sounds reasonable, I’d be happy to submit a PR for it.

2. gen_ai.system_instructions in span! I noticed that gen_ai.system_instructions is often included in the span! macro, and in practice, this tends to be a fairly long preamble across many providers. Would it make sense to exclude it from span! to reduce noise in traces? Just curious to hear your thoughts — not a strong opinion either way.

Thanks for considering!

joshua-mo-143:
Hmmmm, I do think it would be useful to be able to configure these to either be disabled or enabled. I'll look into how we can do it since there are some things that need to be clarified on the tracing APIs.

PR #1019, opened 2025-10-31
feat: add http_client request logging #1018

cody-why:
Description: This merge introduces HTTP request logging capabilities to the http_client module. It adds support for previewing request bodies at the TRACE level, with configurable maximum preview length via max_log_body_preview. Sensitive headers like Authorization are redacted before logging. The implementation includes:

HttpLogConfigExt trait for client builders

set_max_log_body_preview function

Escape helper for safe byte logging

log_request and log_headers functions for structured trace output

This enhancement improves observability and debugging of outbound HTTP traffic.


joshua-mo-143:
Thanks for submitting this!

Will need to potentially have internal discussion on Monday to see if this is within scope for Rig (it should be, but there may be other things that have not been considered yet) - once that's all cleared I'll do the review

rig-core/src/http_client/mod.rs
```diff
@@ -246,3 +299,51 @@ impl HttpClientExt for reqwest::Client {
         }
     }
 }
+
+fn log_request(parts: &Parts, body: &Bytes) {
+    if tracing::enabled!(Level::TRACE) {
+        // Redact sensitive headers (e.g., Authorization) for logging
+        let mut redacted_headers = parts.headers.clone();
+        redacted_headers.remove("Authorization");
```
joshua-mo-143: You will also need to remove any headers that contain "-key" (as this is often used to hold API keys, etc).

rig-core/src/http_client/mod.rs
```diff
@@ -246,3 +299,51 @@ impl HttpClientExt for reqwest::Client {
         }
     }
 }
+
+fn log_request(parts: &Parts, body: &Bytes) {
+    if tracing::enabled!(Level::TRACE) {
+        // Redact sensitive headers (e.g., Authorization) for logging
```
joshua-mo-143: Can be removed.

It's pretty obvious what this is doing

rig-core/src/http_client/mod.rs
```diff
@@ -75,6 +78,46 @@ pub fn with_bearer_auth(req: Builder, auth: &str) -> Result<Builder> {
     Ok(req.header("Authorization", auth_header))
 }
 
+static LOG_HTTP_BODY_MAX: AtomicUsize = AtomicUsize::new(8 * 1024);
```
joshua-mo-143: Not really a fan of this static global. Is there a way we can do this per-type impl?

rig-core/src/http_client/escape.rs
```diff
@@ -0,0 +1,41 @@
+use std::fmt;
```
joshua-mo-143, line 1: Please rename this file to `escape.rs` rather than `util.rs`.

Utility/helper files can generally lead to code smells if not well-maintained. I would like to be optimistic about this, but realistically it'll be a matter of when it will be a code smell and not if - so the best solution is to eliminate the problem entirely.

rig-core/src/http_client/mod.rs
```diff
@@ -202,10 +252,13 @@ impl HttpClientExt for reqwest::Client {
     {
         let (parts, body) = req.into_parts();
 
+        let body_bytes: Bytes = body.into();
+        log_request(&parts, &body_bytes);
```
joshua-mo-143, line 285: Is it possible to make an on/off toggle for this?

rig-core/src/http_client/mod.rs
```diff
@@ -159,6 +206,9 @@ impl HttpClientExt for reqwest::Client {
         U: WasmCompatSend + 'static,
     {
         let (parts, body) = req.into_parts();
+
+        log_headers(&parts);
```
joshua-mo-143, line 239: Is it possible to make an on/off toggle for this?

joshua-mo-143 (changes requested):
Some changes requested, check my comments.

I am not *entirely* sure if this is what we want given that theoretically there is nothing stopping you from just having this in your custom impl (... and `reqwest` has its own traces), but I think this could still serve a purpose given that pretty much every request has to pass through the HttpClientExt impl anyway.

cody-why:
Thanks for the feedback! I’ve pushed a new commit that introduces the requested configuration switch so the logging can be toggled on/off. I also agree with your point that reqwest already provides its own traces, so this addition may not be strictly necessary. If the team feels it’s out of scope, I’m fine with not merging it.

joshua-mo-143:
Closing for now as I think this remains out of scope. Users can implement logging pretty much wherever they want in their own implementation.

PR #1021, opened 2025-10-31
Some compatible OpenAI service responses may not have `tools` field

windoze:
Some compatible OpenAI service responses may not have `tools` field, e.g. LM Studio, this will cause deserialization failure.

Make this field optional to fix this issue.

### DIFF
--- a/rig-core/src/providers/openai/responses_api/mod.rs
+++ b/rig-core/src/providers/openai/responses_api/mod.rs
@@ -755,6 +755,7 @@ pub struct CompletionResponse {
     /// The model output (messages, etc will go here)
     pub output: Vec<Output>,
     /// Tools
+    #[serde(default)]
     pub tools: Vec<ResponsesToolDefinition>,
     /// Additional parameters
     #[serde(flatten)]


joshua-mo-143:
Hi, thanks for opening this PR! I was not aware that other providers had now started implementing their own version of OpenAI Responses.

Can you provide an example of this failing? It seems that LMStudio simply refer to OpenAI's Responses API partially for this rather than actually providing full documentation on this.

windoze:
I've got 'CompletionError: JsonError: missing field `tools` at line 30 column 1' error when I use LM Studio API server, the response is
```
{
  "id": "resp_099a2298a07f6207a5bd3648dd4bce2dc6a1a817d061ddb8",
  "object": "response",
  "created_at": 1761913150,
  "status": "completed",
  "model": "qwen3-next-80b-a3b-instruct-mlx",
  "output": [
    {
      "id": "msg_ulpvptya11o13ul8dq6jm",
      "type": "message",
      "role": "assistant",
      "status": "completed",
      "content": [
        {
          "type": "output_text",
          "text": "<Actual response text......>"
        }
      ]
    }
  ],
  "usage": {
    "input_tokens": 1034,
    "output_tokens": 1156,
    "total_tokens": 2190,
    "output_tokens_details": {
      "reasoning_tokens": 0
    }
  },
  "previous_response_id": null
}
```

Obviously the `tools` field doesn't exist, despite how the LM Studio document says.

windoze:
Great, thank you so much for this handy project, I used it for many projects and it's very convenient.

PR #1028, opened 2025-11-03
refactor(rig-1031): remove `worker` dependency

### LINKED ISSUE #1027: refactor: Remove `#[worker::send]` from all providers

## Feature Request

Now that we are using `WasmCompatSend`, we should theoretically be able to get rid of `#[worker::send]` from all relevant provider methods. This was tested locally with OpenAI and it still compiles, so I suspect that this should work with other providers too.

The idea is that by doing so, we can get rid of the `worker` crate dependency and therefore the feature.

### Acceptance criteria
- The `worker` dependency must be deleted from `rig-core`
- The `worker` feature in `rig-core` should be renamed to `wasm`
- We need to ensure that a CF worker module with Rig in still compiles with the `wasm` feature (to `wasm32-unknown-unknown`).
- The CI needs to be changed to use the `wasm` feature rather than `worker`.

joshua-mo-143:
Fixes #1027, and additionally does the following:

- Changes all references to WASM conditional compilation to use the `wasm` feature (edit: and additionally check for wasm32 target arch).

### DIFF
--- a/rig-core/src/agent/prompt_request/streaming.rs
+++ b/rig-core/src/agent/prompt_request/streaming.rs
@@ -20,11 +20,11 @@ use crate::{
     tool::ToolSetError,
 };
 
-#[cfg(not(feature = "wasm"))]
+#[cfg(not(all(feature = "wasm", target_arch = "wasm32")))]
 pub type StreamingResult<R> =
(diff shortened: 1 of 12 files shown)


joshua-mo-143:
Todo:
- Check that we can still compile a CF worker module using Rig

joshua-mo-143:
Closing due to re-architected packages causing too many issues, will re-open a new PR once it's ready.

PR #1030, opened 2025-11-03
Fix Gemini responses lacking content

ribelo:
### Problem

- Gemini occasionally returns candidates with `finishReason` set (e.g., `MALFORMED_FUNCTION_CALL`) but no `candidate.content` field.
- When this happens, `rig-core`’s Gemini adapter fails deserializing into `ContentCandidate { content: Content, … }` with `JsonError: missing field content`, causing the entire call to error out in `serde` rather than surfacing a provider error.

### Evidence

- Real responses observed from the Gemini API for PDF + tool-calling scenarios include candidates without content and only `finishMessage`, e.g.:

  ```json
  {
    "candidates": [{
      "finishReason": "MALFORMED_FUNCTION_CALL",
      "finishMessage": "Malformed function call: …",
      "index": 0
    }],
    "usageMetadata": { … }
  }
  ```
- Similar reports in the wild:
    - Google AI forum: [“Getting FinishReason.MALFORMED_FUNCTION_CALL when function calling arguments contain large amount of text”](https://discuss.ai.google.dev/t/getting-finishreason-malformed-function-call-when-function-calling-arugments-contain-large-amount-of-text-content/69488)
    - Pydantic AI issue: [“MALFORMED_FUNCTION finishReason in Gemini candidate”](https://github.com/pydantic/pydantic-ai/issues/631)

### Summary of Changes

- Make `ContentCandidate.content` optional (`Option<Content>`).
- When content is absent, return a clear `CompletionError::ResponseError` that includes `finish_reason` and `finish_message`, instead of a `serde` failure.
- Add a debug log of the raw Gemini response when completion deserialization fails to aid diagnosis.
- Update streaming and transcription paths to handle optional content, and adjust tests accordingly.
- No change in behavior for well-formed responses; for malformed/safety cases we now fail gracefully with actionable errors rather than panicking in `serde`.

### Validation

- `cargo test -p rig-core --features "reqwest-rustls,image,audio,derive"` passes locally.
### Why this approach

- The provider sometimes omits content in error/edge cases; making the field optional aligns the adapter with real API behavior and prevents hard failures.
- Surfacing `finish_reason`/`finish_message` provides actionable context to callers.
- Logging the raw body on parse errors makes future provider changes easier to diagnose.

### DIFF
--- a/rig-core/src/providers/gemini/completion.rs
+++ b/rig-core/src/providers/gemini/completion.rs
@@ -126,7 +126,18 @@ where
                 .await
                 .map_err(CompletionError::HttpError)?;
 
-            let response: GenerateContentResponse = serde_json::from_slice(&response_body)?;
+            let response_text = String::from_utf8_lossy(&response_body).to_string();
+            tracing::debug!("Received raw response from Gemini API: {}", response_text);
+
+            let response: GenerateContentResponse = serde_json::from_slice(&response_body)
+                .map_err(|err| {
+                    tracing::error!(
+                        error = %err,
+                        body = %response_text,
+                        "Failed to deserialize Gemini completion response"
+                    );
+                    CompletionError::JsonError(err)
+                })?;
 
             match response.usage_metadata {
                 Some(ref usage) => tracing::info!(target: "rig",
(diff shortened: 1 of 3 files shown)


joshua-mo-143 (approved):
lgtm

Issue #980, opened 2025-10-24
feat: Add backend specific vector search filters


FayCarsons:
- [x] I have looked for existing issues (including closed) about this

## Feature Request
Now that we have basic filtering operations for all vector store backends, we should add backend-specific operations like `gte`, `exists`, and `member` where applicable. 

For example, in the MongoDB implementation:
```rust
#[derive(Clone, Debug)]
pub struct MongoDbSearchFilter(Document);

impl SearchFilter for MongoDbSearchFilter {
    type Value = Bson;

    fn eq(key: String, value: Self::Value) -> Self {
        Self(doc! { key: value })
    }

    fn and(self, rhs: Self) -> Self {
        Self(doc! { "$and": [ self.0, rhs.0 ]})
    }

    // and so on...
}

impl MongoDbSearchFilter {
    pub fn gte(key: String, value: <Self as SearchFilter>::Value) -> Self {
        Self(doc! { key: { "$gte": value } })
    }

    pub fn lte(key: String, value: <Self as SearchFilter>::Value) -> Self {
        Self(doc! { key: { "$lte": value } })
    }

    pub fn not(self) -> Self {
        Self(doc! { "$not": self.0 })
    }
}
```

Currently there are a handful of extensions, but it should be more consistent and exhaustive wrt to a given backend's features

PR #1032, opened 2025-11-04
feat(rig-1014): add backend specific vector search filters

FayCarsons:
implements #980

Also improved MongoDB filter impl performance by moving filtering to the request vs aggregation pipeline.


joshua-mo-143 (approved):
lgtm

Issue #568, opened 2025-07-08
feat: Add support for Vertex AI


shaileshahuja:
## Model Provider Integration Request
Vertex AI (Gemini wrapper)?

### Resources
https://cloud.google.com/vertex-ai/generative-ai/docs
https://cloud.google.com/vertex-ai/generative-ai/docs/start/quickstart

bharattech:
Hi Josh,

i would like to work on this issue

joshua-mo-143:
Taking over this ticket as there has been no obvious update and this is relatively high priority for business adoption.

Aditya-PS-05:
@joshua-mo-143 I would love to work.

joshua-mo-143:
@Aditya-PS-05 Assigned 🚀 

Some stuff you'll need to keep in mind:
- You will probably need to make use of `yup-oauth2` as this ticket requires the google application details credentials stuff
- This integration probably needs to be done in a side-crate as I am not entirely sure `yup-oauth2` is WASM friendly and we'd much rather keep providers that require extra dependencies out of `rig-core`

Aditya-PS-05:
@joshua-mo-143 making great progress.

joshua-mo-143:
Closed because the PR to add this in has been merged.

PR #1040, opened 2025-11-11
feat: rig-vertexai

kafonek:
Hello Rig team. Thanks for putting this project out in the ecosystem. I'm a first time contributor. I used `rig` in a prototype at work, where by happenstance it worked best to use vertex. I saw issue https://github.com/0xPlaygrounds/rig/issues/568 and wanted to contribute a potential implementation. 

I've "click tested" this insomuch that I've used this branch in an effective in-house prototype that does multi-turn tool calling to gemini 2.5 flash. I scoped this PR to not include Image or Streaming or anything else beyond text-based tool calling support. I've tried to follow patterns in other Provider members of the workspace (particularly Bedrock). The examples work on my setup.

In issue 568 there was a recommendation to use `yap-oauth2`, which I haven't tried. We use the (google-maintained) google-cloud-rust crates extensively especially since they were released past 1.0, and I think those fit idiomatically. 

I'm happy to address feedback, also welcome maintainers to commit changes directly. If you have another PR working on this integration, feel free to close this one and use it solely as comparison / inspiration. Thanks again for the open source contribution.

### DIFF
--- a/rig-vertexai/examples/completion_vertexai.rs
+++ b/rig-vertexai/examples/completion_vertexai.rs
@@ -24,7 +24,7 @@ async fn main() -> Result<(), anyhow::Error> {
     let mut response_text = String::new();
     for content in response.choice.iter() {
         if let rig::message::AssistantContent::Text(rig::message::Text { text }) = content {
-            response_text.push_str(&text);
+            response_text.push_str(text);
         }
     }
 


joshua-mo-143:
Hi @kafonek, thank you for this PR! Looks great. I initially suggested `yup-oauth2` as that was mostly the only crate I really knew of that was suitable for the auth issue, but if the google cloud crates work for you guys then I am sure they will work for others as well.

I'll get to a review as soon as I am able to do so.

kafonek:
Last set of commits are feedback from a colleague who works regularly on the google genai libraries, including google-cloud-rust. They also mentioned that `GEMINI_API_BASE_URL` `https://generativelanguage.googleapis.com` in the non-vertex gemini provider should probably be `aiplatform.googleapis.com`, can open a separate issue/PR for that later.

rig-vertexai/src/client.rs
```diff
@@ -0,0 +1,243 @@
...
+    {
+        Client::new()
+    }
+
+    fn from_val(_: ProviderValue) -> Self
+    where
+        Self: Sized,
+    {
+        panic!(
+            "Vertex AI uses Application Default Credentials (ADC). Use `Client::from_env()` for default credentials, or `Client::new().with_credentials(...).build()` for custom credentials."
```
joshua-mo-143, line 224: This is not ideal but I guess we can't really fix this in the space of this PR

rig-vertexai/src/types/json_utils.rs
```diff
@@ -0,0 +1,79 @@
+//! Utilities for converting between serde_json::Value and wkt::Struct (protobuf Struct)
+//!
+//! Note: wkt::Struct is a type alias for serde_json::Map<String, serde_json::Value>
```
joshua-mo-143: Where does `wkt` come from here?

rig-vertexai/src/types/json_utils.rs
```diff
@@ -0,0 +1,79 @@
...
+
+use rig::completion::CompletionError;
+
+pub type Struct = serde_json::Map<String, serde_json::Value>;
```
joshua-mo-143: Not entirely happy with this naming since just naming it `Struct` leaves a lot of ambiguity without a prior understanding of what `wkt::Struct` is.

This might be fixed by adding aforementioned documentation or probably just renaming it to something like `JsonMap`.

kafonek: Good feedback. Let me see what I can refactor for better readability here. The `Struct` name came from google-cloud-rust convention, https://github.com/googleapis/google-cloud-rust/blob/main/src/wkt/src/rstruct.rs#L15. I agree it's not a good fit here. We're essentially converting between the rig-core types (newtyped in `rig-vertexai`) and the well known types (wkt) that the google-cloud crates want for their API request and response args.

kafonek: I removed the `json_utils.rs` entirely. For converting vertexai to rig tool calling it's a simple cast from `serde_json::Map` to `serde_json::Value::Object`. Going the other way it looked like it was easiest to collect the tool result contents into a `serde_json::Map` with one key. So either `{"output": tool_content}` or `{"output": [content1, content2]}`.

joshua-mo-143:
Looks like the `rig-core` version broke the lib because we released a new version while this PR was still in progress.

You just need to update it to 0.24 then it should compile again

joshua-mo-143:
This pretty much looks good to me. Just need to fix the clippy lint and we can get this merged in.

PR #1047, opened 2025-11-13
fix: improve Qdrant insert error handling

cody-why:
Reason: Previously, document inserts into Qdrant could fail silently — errors were sent but not surfaced, 
leading to unsuccessful writes without feedback. This change adds proper error handling and reporting.

Note: `.wait(true)` has been added to make inserts synchronous. While this may impact performance, 
it ensures timely error feedback and more reliable insert operations.

### DIFF
--- a/rig-qdrant/src/lib.rs
+++ b/rig-qdrant/src/lib.rs
@@ -85,8 +85,9 @@ where
         let collection_name = self.query_params.collection_name.clone();
 
         for (document, embeddings) in documents {
-            let json_document = serde_json::to_value(&document).unwrap();
-            let doc_as_payload = Payload::try_from(json_document).unwrap();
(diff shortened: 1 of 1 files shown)


joshua-mo-143:
Thanks for the PR!

What happens if you remove `.wait(true)` and get an error?

cody-why:
If .wait(true) is removed, then self.client.upsert_points runs asynchronously and immediately returns success. For example, when I insert data with incorrect dimensions, it reports success without surfacing any error, even though the insert actually fails. This means the operation doesn’t provide timely feedback about unsuccessful writes.

joshua-mo-143:
Ok lgtm, I'll check with the Qdrant team to see if this is intentional or not

PR #1049, opened 2025-11-14
Require 'static lifetime for CompletionModel

Rishirandhawa:
Adds a 'static bound to the CompletionClient::CompletionModel associated type so implementations can be turned into trait objects (Box/Arc<dyn ...>). This addresses compiler lifetime/bound errors when converting CompletionModels into dynamic trait objects for CompletionClientDyn/AgentBuilder.
Notes: Most provider implementations already satisfy this because they bound their inner generic HTTP client with 'static

### DIFF
--- a/rig-core/src/client/completion.rs
+++ b/rig-core/src/client/completion.rs
@@ -17,7 +17,9 @@ use std::sync::Arc;
 /// Clone is required for conversions between client types.
 pub trait CompletionClient: ProviderClient + Clone {
     /// The type of CompletionModel used by the client.
-    type CompletionModel: CompletionModel;
+    // require the associated CompletionModel to be 'static so it can be used
+    // behind trait objects (Box/Arc<dyn ... + 'static>) safely
+    type CompletionModel: CompletionModel + 'static;
 
     /// Create a completion model with the given name.
     ///


rig-core/src/client/completion.rs
```diff
@@ -17,7 +17,9 @@ use std::sync::Arc;
 /// Clone is required for conversions between client types.
 pub trait CompletionClient: ProviderClient + Clone {
     /// The type of CompletionModel used by the client.
-    type CompletionModel: CompletionModel;
+    // require the associated CompletionModel to be 'static so it can be used
```
joshua-mo-143, line 20: Please use the original docstring here and move this detail to the trait docstrings rather than having it here

joshua-mo-143 (changes requested):
Only one change required then lgtm

Sytten:
I mean we already have CompletionClientDyn for this, unless you want to get rid of the dyn version @joshua-mo-143 I would probably not impose that constraint on the trait.

joshua-mo-143:
> I mean we already have CompletionClientDyn for this, unless you want to get rid of the dyn version @joshua-mo-143 I would probably not impose that constraint on the trait.

Ah, yeah - completely forgot about that.

@Rishirandhawa Is there any specific reason why you need this specific change rather than CompletionClientDyn?

joshua-mo-143:
Closing in the interest of repo hygiene since there's no response and there was no prior discussion.

PR #1050, opened 2025-11-14
feat(rig-985): Consolidate provider clients

### LINKED ISSUE #911: feat: Consolidate provider clients

## Feature Request
Currently, all ~20 providers have unique client implementations resulting in a lot of repeated code, more surface for logic errors, and difficulty in refactoring. Clients could be consolidated into one `Client<Ext>` type, with some machinery to account for any custom behavior. This should have little to no effect on the public API, and could potentially simplify `DynClientBuilder` etc.

### Proposal
The `Client` type could look like this:
```rust
pub struct Client<H, Ext> {
  base_url: String,
  http_client: H,
  ext: Ext,
}
```

Where `Ext` is some type acting as state specific to a provider, and implements a trait `CustomizeRequest` with a method that is `fn(Self, http_client::Request) -> http_client::Request` which can be used to add credentials, provider-specific headers, etc.

Some small details need to be worked out:
- How do we get things like the URI used in `VerifyClient`? Could be done with associated consts, but would require reworking the trait or adding another.
- How do we handle differences in embedding APIs?
- Are NDJSON + SSE support added via this `Ext` field, or do we need separate machinery for that?  

With all of that worked out, the final result should look something like this:

```rust
struct AnthropicClientExt {
  api_key: String,
  anthropic_version: String,
  anthropic_betas: Option<String>,
}

impl CustomizeRequest for AnthropicClientExt {
  fn with_custom(self, req: Request) {
    req
      .header("X-Api-Key", self.api_key)
      .header("anthropic-version", self.anthropic_version)
  }
}

// Comes with get, post, builder, etc defined for us
type AnthropicClient<H> = Client<H, AnthropicClientExt>;
```

FayCarsons:
Fixes #911 

A draft for the moment because there are still a couple of small things to fix, but 99% of it is in place

### DIFF
--- a/rig-core/src/client/mod.rs
+++ b/rig-core/src/client/mod.rs
@@ -58,13 +58,9 @@ pub trait ProviderClient {
 
     /// Create a client from the process's environment.
     /// Panics if an environment is improperly configured.
-    fn from_env() -> Self
-    where
-        Self: Sized;
+    fn from_env() -> Self;
 
-    fn from_val(input: Self::Input) -> Self
-    where
-        Self: Sized;
+    fn from_val(input: Self::Input) -> Self;
 }
 
 use crate::completion::{GetTokenUsage, Usage};
@@ -81,21 +77,24 @@ impl GetTokenUsage for FinalCompletionResponse {
     }
 }
 
+/// A trait for API keys. This determines whether the key is inserted into a [Client]'s default
+/// headers (in the `Some` case) or handled by a given provider extension (in the `None` case)
 pub trait ApiKey: Sized {
     fn into_header(self) -> Option<http_client::Result<(HeaderName, HeaderValue)>> {
         None
     }
 }
 
-pub struct SimpleKey(String);
+/// An API key which will be inserted into a `Client`'s default headers as a bearer auth token
+pub struct BearerAuth(String);
 
-impl ApiKey for SimpleKey {
+impl ApiKey for BearerAuth {
     fn into_header(self) -> Option<http_client::Result<(HeaderName, HeaderValue)>> {
         Some(make_auth_header(self.0))
     }
 }
 
-impl<S> From<S> for SimpleKey
+impl<S> From<S> for BearerAuth
 where
     S: Into<String>,
 {
@@ -112,14 +111,16 @@ pub struct Nothing;
 impl ApiKey for Nothing {}
 
 impl TryFrom<String> for Nothing {
-    type Error = String;
+    type Error = &'static str;
 
     fn try_from(_: String) -> Result<Self, Self::Error> {
-        Ok(Nothing)
+        Err(
+            "Tried to create a Nothing from a string - this should not happen, please file an issue",
+        )
     }
 }
 
-// So that i.e Ollama can ignore auth
+// So that Ollama can ignore auth
 #[derive(Clone)]
 pub struct Client<Ext = Nothing, H = reqwest::Client> {
     base_url: String,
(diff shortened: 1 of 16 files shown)


rig-core/src/providers/azure.rs
```diff
@@ -34,93 +42,118 @@ use serde_json::json;
...
         Self {
-            auth: auth.into(),
-            api_version: None,
-            azure_endpoint: endpoint,
-            http_client: Default::default(),
+            endpoint: Default::default(),
+            api_version: DEFAULT_API_VERSION.into(),
         }
     }
 }
```
FayCarsons: @joshua-mo-143 Did everything we talked about in standup today, I think this is the last thing which needs to be addressed. 

Not sure if we can have a default for the endpoint, or if we want to somehow enforce that its been set before building.

joshua-mo-143: We need to enforce that it's been set while building (so you probably need an `Option<String>` here).

FayCarsons:
Ok I think the actual last thing which may need to be addressed here is naming. I think the `Provider`, `ProviderBuilder` and `ProviderClient` traits are not named very well and could potentially confuse users

joshua-mo-143:
> Ok I think the actual last thing which may need to be addressed here is naming. I think the `Provider`, `ProviderBuilder` and `ProviderClient` traits are not named very well and could potentially confuse users

I think if we add some docstrings to explain what they are/how they're used it *should* be OK. I don't think there is anything really more appropriate we can name them, so the next best thing is to add docstrings explaining exactly what they do.

You will probably need to change the docstrings for `ProviderClient` as it doesn't require all the traits that it did before.

rig-core/src/providers/gemini/completion.rs
```diff
@@ -53,14 +54,49 @@ use super::Client;
...
+        Gemini15Flash => "gemini-1.5-flash",
+        /// `gemini-1.5-pro` completion model
+        Gemini15Pro => "gemini-1.5-pro",
+        /// `gemini-1.5-pro-8b` completion model
+        Gemini15Pro8B => "gemini-1.5-pro-8b",
+        /// `gemini-1.0-pro` completion model
+        Gemini10Pro => "gemini-1.0-pro",
+    }
+}
+pub use CompletionModels::*;
```
joshua-mo-143, line 90: At first glance, this is *sort of* an anti-pattern because we shouldn't really be trying to import/`use` individual completion models... I don't really see how else we can really do this though (and agent still takes &str anyway), so I guess we can let this slide until we know better

FayCarsons: Yeah it's not the most idiomatic but it's what I figured would require the least work wrt to migration. Didn't want to expand scope too much here wrt `agent` and a few other places where this could be integrated better, it was mostly a way to glue some things together w/ the benefit of more type safety in the long run.

Worth noting models still hold strings internally, so a `with_custom_model` or similar method can be added.

FayCarsons: Added doc comments to `client/mod.rs` btw

cody-why:
Thank you for the update. We urgently need support for a custom model.

patonw:
I know it's deprecated now, but has anyone tried to use the new `DynClientBuilder`?

It seems like this PR broke it in a rather obvious way: providers are inserted into the map by just provider name. Lookups are done with `provider:model` though. Am I just missing something fundamental here?

Also there doesn't seem to be a reason why the provider arguments need to be static in the factory methods (factory, agent, completion, etc). Having that constraint prevents using anything parsed at runtime, like config files or program arguments to build a client.

Issue #1051, opened 2025-11-14
bug: Openrouter is not using the SSE event stream


Sytten:
## Bug Report
For some reason PR #815 added a new method for the streaming, but didn't remove the old one and then in #955 we deleted it. 

Unsure why this was done or if it was intentional, we should most likely use the sse for this right?

PR #1052, opened 2025-11-14
Fix providers

Sytten:
This is got a bit out of control. I found a lot of issues during the last days.

- Support reasoning in anthropic non-streaming api
- Support reasoning in openrouter
- Cleanup dead code
- Changed logging of request/response to trace and uniformize the target
- Moved openrouter to SSE parser
- Refactored openai message parsing to allow more granular re-use
- Ensure the streaming FinalResponse is always sent even on error
- Uniformize the tracing targets
- Ensure Otel doesn't do useless compute for disabled span

Fixes https://github.com/0xPlaygrounds/rig/issues/1051

### DIFF
--- a/rig-core/src/telemetry/mod.rs
+++ b/rig-core/src/telemetry/mod.rs
@@ -55,6 +55,10 @@ impl SpanCombinator for tracing::Span {
     where
         U: GetTokenUsage,
     {
+        if self.is_disabled() {
+            return;
+        }
+
         if let Some(usage) = usage.token_usage() {
             self.record("gen_ai.usage.input_tokens", usage.input_tokens);
             self.record("gen_ai.usage.output_tokens", usage.output_tokens);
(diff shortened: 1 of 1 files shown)


rig-core/src/providers/openai/responses_api/streaming.rs
```diff
@@ -340,8 +340,8 @@ where
                 }
             }
 
-            // // Ensure event source is closed when stream ends
-            // event_source.close();
+            // Ensure event source is closed when stream ends
+            event_source.close();
```
Sytten, line 344: @joshua-mo-143 You commented this in #955, was that intentional?
I figured it would be safe to re-enable.

joshua-mo-143:
Hey @Sytten, thanks for the PR!

We are currently doing a provider reconsolidation #1050 - at the internal sync today I'll align the team on a merge strategy since we should probably be handling this a bit more carefully than usual

Sytten:
@joshua-mo-143 Make sense, as far as I see the work in that PR is mostly done for the http layer. I didn't touch that layer really so it should be easier to rebase on top.

rig-core/src/providers/openai/responses_api/streaming.rs
joshua-mo-143: yeah that's fine, think it was an accident

Issue #1054, opened 2025-11-15
feat: Re-think streaming events


Sytten:
- [x] I have looked for existing issues (including closed) about this

## Feature Request
I think we should completely re-think events and use the model provided by the Vercel AI SDK:
```typescript
type LanguageModelV2StreamPart = {
    type: 'text-start';
    providerMetadata?: SharedV2ProviderMetadata;
    id: string;
} | {
    type: 'text-delta';
    id: string;
    providerMetadata?: SharedV2ProviderMetadata;
    delta: string;
} | {
    type: 'text-end';
    providerMetadata?: SharedV2ProviderMetadata;
    id: string;
} | {
    type: 'reasoning-start';
    providerMetadata?: SharedV2ProviderMetadata;
    id: string;
} | {
    type: 'reasoning-delta';
    id: string;
    providerMetadata?: SharedV2ProviderMetadata;
    delta: string;
} | {
    type: 'reasoning-end';
    id: string;
    providerMetadata?: SharedV2ProviderMetadata;
} | {
    type: 'tool-input-start';
    id: string;
    toolName: string;
    providerMetadata?: SharedV2ProviderMetadata;
    providerExecuted?: boolean;
} | {
    type: 'tool-input-delta';
    id: string;
    delta: string;
    providerMetadata?: SharedV2ProviderMetadata;
} | {
    type: 'tool-input-end';
    id: string;
    providerMetadata?: SharedV2ProviderMetadata;
} | LanguageModelV2ToolCall | LanguageModelV2ToolResult | LanguageModelV2File | LanguageModelV2Source | {
    type: 'stream-start';
    warnings: Array<LanguageModelV2CallWarning>;
} | ({
    type: 'response-metadata';
} & LanguageModelV2ResponseMetadata) | {
    type: 'finish';
    usage: LanguageModelV2Usage;
    finishReason: LanguageModelV2FinishReason;
    providerMetadata?: SharedV2ProviderMetadata;
} | {
    type: 'raw';
    rawValue: unknown;
} | {
    type: 'error';
    error: unknown;
};
```

### Motivation
Right now different providers behave differently, some will send a final message like anthropic with the full content.
This is because the events have historically been used in different context.
It is also very unclear if an event `Text` is the whole previous text or just the new delta. IMO it should just be the delta. We can create an accumulation abstraction but it should not be the default.

### Proposal
I propose we refactor the enum of streaming event to something like:
```rust
#[derive(Debug, Clone)]
[...]

PR #1055, opened 2025-11-15
Reasoning Delta

Sytten:
While not a full fix for #1054, this provides temporary relief for mixed provider setup so the user can at least differentiate between the two events.

All providers are expect to send Delta events, but not all will send the full reasoning at the end.

Merge after #1052

### DIFF
--- a/rig-bedrock/src/streaming.rs
+++ b/rig-bedrock/src/streaming.rs
@@ -107,10 +107,9 @@ impl CompletionModel {
                                         }
 
                                         if !text.is_empty() {
-                                            yield Ok(RawStreamingChoice::Reasoning {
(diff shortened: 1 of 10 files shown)


Sytten:
@joshua-mo-143 I rebased, ready for review.
Unsure if you want to merge this or not, I still thinks its an improvement toward a better event system but happy to refactor further

Sytten:
@joshua-mo-143 Gentle ping, just want to know what your take is on that :)

joshua-mo-143:
Hey @Sytten - thanks for your patience on this.

Looks good to me. I think we could probably really do with disambiguating the reasoning deltas and full reasoning items.

PR #1060, opened 2025-11-18
fix(gemini-request): add `#[serde(default)]` for missing `generation_config` field

### LINKED ISSUE #1059: bug: Gemini provider requires unnecessary boilerplate for generationConfig
- [ √] I have looked for existing issues (including closed) about this

## Bug Report
Currently, the Gemini provider requires users to explicitly provide an empty `generationConfig` object, even when no specific configuration is needed. If the `generationConfig` is omitted, the request will fail during deserialization. This leads to unnecessary boilerplate code for the user.

## Reproduction
To create a basic agent, a user currently must write the following code:
```rust
let google_client = gemini::Client::from_env();
google_client.agent(gemini::completion::GEMINI_2_5_FLASH_PREVIEW_05_20)
    .additional_params(json!({
        "generationConfig": {}
    }))
    .build();
```

## Expected behavior
It should be possible to create the agent without providing the empty `generationConfig` object. The API should handle the default case gracefully. The expected code should be:

```rust
let google_client = gemini::Client::from_env();
google_client.agent(gemini::completion::GEMINI_2_5_FLASH_PREVIEW_05_20)
    .build();
```

## Screenshots
N/A

## Additional context
This issue can be fixed by adding the `#[serde(default)]` attribute to the `generation_config` field. This will make the provider more robust and improve the developer experience.

I have a Pull Request prepared to fix this.

Numeration:
Fixes #1059

### DIFF
--- a/rig-core/src/providers/gemini/completion.rs
+++ b/rig-core/src/providers/gemini/completion.rs
@@ -407,6 +407,7 @@ pub mod gemini_api_types {
     #[serde(rename_all = "camelCase")]
     pub struct AdditionalParameters {
         /// Change your Gemini request configuration.
+        #[serde(default)]
         pub generation_config: GenerationConfig,
         /// Any additional parameters that you want.
         #[serde(flatten, skip_serializing_if = "Option::is_none")]


joshua-mo-143:
lgtm - thanks!

PR #1067, opened 2025-11-20
refactor(rig-777): proper request modelling for every provider

### LINKED ISSUE #496: feat: request modelling for every single provider

joshua-mo-143:
Fixes #496


rig-core/src/providers/together/completion.rs
```diff
@@ -135,31 +134,36 @@ pub use CompletionModels::*;
...
-        if let Some(docs) = completion_request.normalized_documents() {
+        if let Some(docs) = req.normalized_documents() {
             let docs: Vec<openai::Message> = docs.try_into()?;
             full_history.extend(docs);
         }
-        let chat_history: Vec<openai::Message> = completion_request
+
+        let chat_history: Vec<openai::Message> = req
             .chat_history
+            .clone()
```
Sytten: Do you need to clone the chat history?

joshua-mo-143: Oh oops, not sure how a clone got in there. I'll fix this in a bit

PR #1069, opened 2025-11-20
fix(rig-1050): Inconsistent model/agent initialisation methods

### LINKED ISSUE #1066: bug: Inconsistent model/agent initialisation methods

## Bug Report

There are currently inconsistent model/agent initialisation methods between clients on `main` branch.

We should probably align on one and make all of them the same, since consistency in a unified API is pretty important.

FayCarsons:
Fixes #1066

### DIFF
--- a/rig-core/src/providers/groq.rs
+++ b/rig-core/src/providers/groq.rs
@@ -22,7 +22,7 @@ use crate::client::{
 use crate::completion::GetTokenUsage;
 use crate::http_client::sse::{Event, GenericEventSource};
 use crate::http_client::{self, HttpClientExt};
-use crate::json_utils::merge;
+use crate::json_utils::{empty_or_none, merge};
 use crate::providers::openai::{AssistantContent, Function, ToolType};
 use async_stream::stream;
 use futures::StreamExt;


rig-core/src/providers/openrouter/completion.rs
```diff
@@ -398,12 +408,16 @@ where
     type StreamingResponse = StreamingCompletionResponse;
 
     type Client = Client<T>;
-    type Models = String;
+    type Models = CompletionModels;
```
FayCarsons: I'm not 100% on the approach we want to take with azure, since they do provide a base set of models but it seems like their main offering is custom models which you set a custom name for

rig-fastembed/src/lib.rs
```diff
@@ -127,6 +127,10 @@ impl embeddings::EmbeddingModel for EmbeddingModel {
         Self::new(&model, dims.unwrap_or_default())
     }
 
+    fn make_custom(_: &Self::Client, _: &str, _: Option<usize>) -> Self {
+        panic!("User defined models must be provided via `new_from_user_defined`")
+    }
+
```
FayCarsons: This sucks, but idk what the alternative is

rig-core/src/client/completion.rs
```diff
@@ -85,7 +103,11 @@ impl CompletionModel for CompletionModelHandle<'_> {
     type Client = ();
     type Models = String;
 
-    fn make(_client: &Self::Client, _model: impl Into<Self::Models>) -> Self {
+    fn make(_: &Self::Client, _: impl Into<Self::Models>) -> Self {
```
joshua-mo-143: Please add docstrings to indicate that usage of this particular method (edit: in this context) should panic

rig-core/src/client/completion.rs
```diff
@@ -85,7 +103,11 @@ impl CompletionModel for CompletionModelHandle<'_> {
...
+        panic!("Cannot create a completion model handle from a client")
+    }
+
+    fn make_custom(_: &Self::Client, _: &str) -> Self {
```
joshua-mo-143: Please add docstrings to indicate that usage of this particular method (edit: in this context) should panic

rig-core/src/client/image_generation.rs
```diff
@@ -70,6 +86,12 @@ mod image {
             )
         }
 
+        fn make_custom(_: &Self::Client, _: &str) -> Self {
```
joshua-mo-143: Please add docstrings to indicate that usage of this particular method should panic in this context.

To be honest, I am sort of thinking we may want to split this into a separate trait if possible... it doesn't make sense to have this method in this trait if we're just going to force it to panic. Not sure if this should be addressed in this PR or another one though.

rig-fastembed/src/lib.rs
joshua-mo-143: If you add docstrings to show that this will panic, I think it will be OK. I think.

rig-core/src/client/audio_generation.rs
```diff
@@ -67,6 +84,12 @@ mod audio {
             )
         }
 
+        fn make_custom(_: &Self::Client, _: &str) -> Self {
```
joshua-mo-143: Please add docstrings to show that this function will panic (as well as *why* it should panic).

rig-core/src/providers/openrouter/completion.rs
joshua-mo-143: Hmmmm. I *think* this should be OK for now given that we have added the escape hatch.

Until we actually get any feedback, we don't and can't know any better, so this will probably have to do.

joshua-mo-143 (changes requested):
See comments. Some docstrings around the panics would be great (since not everyone's going to read the source code and will probably be relying on LSP).

Additionally, I think we need to update the `CompletionClient` trait to also be able to spawn an AgentBuilder from a custom completion model name.

Sytten:
Fixes #1074

erlendp:
How does this work with agents? Can't seem to get it working with the following:

        let model = self.client.custom_completion_model("...");

        let agent = self
            .client
            .agent(model) // <-- Failing to convert here
            .additional_params(serde_json::to_value(cfg)?)
            .build();

I'm also not convinced this is the best way forward with the API - what was wrong with a simple string - doesn't make sense to me that every new model requires a re-compilation of this library to avoid going down the 'custom' path, which appears to have far worse ergonomics...

joshua-mo-143:
> How does this work with agents? Can't seem to get it working with the following:
> 
>         let model = self.client.custom_completion_model("...");
> 
>         let agent = self
>             .client
>             .agent(model) // <-- Failing to convert here
>             .additional_params(serde_json::to_value(cfg)?)
>             .build();
> 
> I'm also not convinced this is the best way forward with the API - what was wrong with a simple string - doesn't make sense to me that every new model requires a re-compilation of this library to avoid going down the 'custom' path, which appears to have far worse ergonomics...

Hi, I've brought this up in my last comment (spawning a model basically using a String). 

The original reason as to why we've ended up here is because of refactoring the DynClientBuilder to not precariously rely on using 5 or 6 different traits and forcing every provider, whether it actually supports the behaviour or not. This PR simply aims to re-align the initialisation methods. 

The Rig team is internally syncing today so we'll be trying to figure out the best solution to this.

We'll be taking some extra time to iron everything out before next release as the change that this follows up from was extremely large, so there were bound to be 1 or 2 things we overlooked

joshua-mo-143 (approved):
lgtm

PR #1077, opened 2025-11-24
feat(gemini-request): add response_json_schema to GenerationConfig

### LINKED ISSUE #1076: feat: Add `response_json_schema` to Gemini `GenerationConfig`

## Feature Request
Add support for the **responseJsonSchema** field in the Gemini provider configuration.

### Motivation
To support the **responseJsonSchema** field as defined in the [Gemini API documentation](https://ai.google.dev/api/generate-content#generationconfig). This allows users to define structured outputs using [JSON Schema](https://json-schema.org/) (supporting `$defs`, `$ref`, etc.) on Gemini models.

### Proposal
Add a `response_json_schema` field to the `GenerationConfig` struct:
```rust
pub response_json_schema: Option<Value>,
```
*I have already implemented this and will submit a PR shortly.*

### Alternatives
N/A

Numeration:
Fixes #1076

### DIFF
--- a/rig-core/src/providers/gemini/completion.rs
+++ b/rig-core/src/providers/gemini/completion.rs
@@ -1309,17 +1309,17 @@ pub mod gemini_api_types {
         /// types: application/json: Schema for JSON response. Refer to the JSON text generation guide for more details.
         #[serde(skip_serializing_if = "Option::is_none")]
         pub response_schema: Option<Schema>,
-        /// Internal or alternative representation for `response_json_schema`.
-        #[serde(
-            skip_serializing_if = "Option::is_none",
-            rename = "_responseJsonSchema"
-        )]
-        pub _response_json_schema: Option<Value>,
         /// Optional. The output schema of the generated response.
         /// This is an alternative to responseSchema that accepts a standard JSON Schema.
         /// If this is set, responseSchema must be omitted.
         /// Compatible MIME type: application/json.
         /// Supported properties: $id, $defs, $ref, type, properties, etc.
+        #[serde(
+            skip_serializing_if = "Option::is_none",
+            rename = "_responseJsonSchema"
+        )]
+        pub _response_json_schema: Option<Value>,
+        /// Internal or alternative representation for `response_json_schema`.
         #[serde(skip_serializing_if = "Option::is_none")]
         pub response_json_schema: Option<Value>,
         /// Number of generated responses to return. Currently, this value can only be set to 1. If


rig-core/src/providers/gemini/completion.rs
```diff
@@ -1309,6 +1309,13 @@ pub mod gemini_api_types {
         /// types: application/json: Schema for JSON response. Refer to the JSON text generation guide for more details.
         #[serde(skip_serializing_if = "Option::is_none")]
         pub response_schema: Option<Schema>,
+        /// Optional. The output schema of the generated response.
+        /// This is an alternative to responseSchema that accepts a standard JSON Schema.
+        /// If this is set, responseSchema must be omitted.
+        /// Compatible MIME type: application/json.
+        /// Supported properties: $id, $defs, $ref, type, properties, etc.
+        #[serde(skip_serializing_if = "Option::is_none")]
+        pub response_json_schema: Option<Value>,
```
joshua-mo-143, line 1324: This should be also tagged with `rename = "_responseJsonSchema"` since that appears to be what the actual field name is in the Gemini API reference.

Numeration: Thanks for the review!

I was a bit confused at first because the official documentation seems contradictory. In the [API reference](https://ai.google.dev/api/generate-content#generationconfig), it labels `responseJsonSchema` as an "internal detail," but then immediately says, "Use `responseJsonSchema` rather than this field," which looks like a typo.

At the same time, the official [REST API example](https://ai.google.dev/gemini-api/docs/structured-output?example=recipe#rest_2) also uses `responseJsonSchema` (without the underscore). These two points lead me to believe that responseJsonSchema is indeed the correct public field.

Nevertheless, to ensure robustness, I will update the code as you suggested and add a `_responseJsonSchema` field to align with the API reference. Thanks for catching this.

joshua-mo-143: Yeah that's understandable. I will never not hate Gemini APIs :joy:

rig-core/src/providers/gemini/completion.rs
```diff
@@ -1309,6 +1309,19 @@ pub mod gemini_api_types {
...
+        #[serde(
+            skip_serializing_if = "Option::is_none",
+            rename = "_responseJsonSchema"
+        )]
+        pub _response_json_schema: Option<Value>,
+        /// Optional. The output schema of the generated response.
+        /// This is an alternative to responseSchema that accepts a standard JSON Schema.
+        /// If this is set, responseSchema must be omitted.
+        /// Compatible MIME type: application/json.
+        /// Supported properties: $id, $defs, $ref, type, properties, etc.
```
joshua-mo-143, line 1316: You need to swap the `_response_json_schema` and `response_json_schema` docstrings around

joshua-mo-143 (changes requested):

rig-core/src/providers/gemini/completion.rs
Numeration: Got it.

PR #1083, opened 2025-11-26
Fix/open ai structured output

TheoOiry:


TheoOiry:
This is a fail sorry for that 😥​

PR #1100, opened 2025-11-30
feat(rig-1072): allow configuring telemetry

### LINKED ISSUE #1097: feat: allow configuring message telemetry

## Feature Request

Allow configuring whether or not the actual message contents of a given request/response should be displayed in logs.

This is pretty important for ensuring that PII stays private.

Related: #1056 

### Motivation

Related to preserving PII and other information that should be encrypted

### Proposal

Each provider/model should have its own telemetry configuration which can be adjusted by the user.

### Alternatives

There isn't really any other ways to solve this issue.  I guess we could use a global mut but there is really no real reason to use it here.

joshua-mo-143:
Fixes #1097

### DIFF
--- a/rig-core/src/agent/builder.rs
+++ b/rig-core/src/agent/builder.rs
@@ -5,6 +5,7 @@ use tokio::sync::RwLock;
 use crate::{
     completion::{CompletionModel, Document},
     message::ToolChoice,
+    telemetry::TelemetryConfiguration,
     tool::{
         Tool, ToolSet,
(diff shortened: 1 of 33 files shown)


joshua-mo-143:
Marked low prio because we're still figuring out the best approach for this.

joshua-mo-143:
Closing as it's not very clear that this is the definitive route we want to approach this from, and the recent telemetry de-bloating has done quite a lot already.

PR #1103, opened 2025-12-01
updated the mod.rs to with the helper function to handle adding the json

yamamotox01:
header

helper function found from lined 96 to 105


joshua-mo-143:
Hey @yamamotox01, unfortunately during the internal Rig team sync today the final verdict was that we probably shouldn't be doing it this way and should instead try to add it into the `HttpClientExt` trait implementation, which means I'll need to close this PR. 

If there's any other PRs you want to work on I'm happy to assign you to them since they have mostly all been accepted for adding.

Superceded by #1106

PR #1105, opened 2025-12-01
Deprecate `DynClientBuilder`

FayCarsons:
Soft deprecate `DynClientBuilder` and add an example of enum dispatch over a couple of provider agents

### DIFF
--- a/rig-core/examples/enum_dispatch.rs
+++ b/rig-core/examples/enum_dispatch.rs
@@ -89,14 +89,14 @@ async fn main() {
         .unwrap();
 
     let oai_response = openai_agent
-        .prompt("How much does 4oz of parmesean cheese weigh")
+        .prompt("How much does 4oz of parmesan cheese weigh")
         .await
         .unwrap();
 
     println!("Helpful: {oai_response}");
 
     let anthropic_response = anthropic_agent
-        .prompt("How much does 4oz of parmesean cheese weigh")
+        .prompt("How much does 4oz of parmesan cheese weigh")
         .await
         .unwrap();
 


joshua-mo-143 (approved):
lgtm

patonw:
I'm curious, what is the rationale for deprecating this?
Enum dispatch has always been an option from the start. This builder is a convenience with negligible cost.

joshua-mo-143:
> I'm curious, what is the rationale for deprecating this? Enum dispatch has always been an option from the start. This builder is a convenience with negligible cost.

Hi!

The current rationale for soft deprecating this is that it's quite difficult to understand how exactly the builder works because the way it's been built is very convoluted. It requires using recursive traits, effectively meaning you had to implement all possible client behaviour types for any given client (completions, embedding, etc...). This created increased coupling across the codebase, which initially wasn't a big deal but with how many more providers we have now it is starting to become a significant burden with it affecting other parts of the codebase.

It's also confusing for new users and contributors to understand *why* you have to implement all those traits for new providers. At this point, just trying to add or take away new things on it is a huge pain in the ass (not to mention that if we ever added any new traits for capabilities we'd have to deal with a load of cascading changes).

From a user standpoint, yes - the builder basically has a negligible cost. However from the internal team POV, it's a difficult piece of code to onboard new team members onto specifically because of the aforementioned issues.

Admittedly, it may have been a bit premature for a deprecation notice. I'll bring it up in the internal sync next week since this does still seem to be a useful feature even if it is difficult to maintain

Sytten:
I personally use the system a lot so this is highly disruptive for me but I understand that it might be a PITA to maintain it.

Ideally you keep supporting this, but at minimum @joshua-mo-143 I would ask for a macro to automatically implement `CompletionModel` over the enum so at least it would be painless to migrate.

Just a sample of how I use it:
```rust
pub trait CompletionModelDynExt: CompletionModelDyn {
    fn reasoning(
        &self,
        reasoning: AIReasoningRequest,
        max_tokens: Option<u64>,
    ) -> Result<Option<serde_json::Value>>;

    fn supports_reasoning(&self) -> bool;

    fn structured_output(
        &self,
        structured_output: AIStructuredOutput,
    ) -> Result<Option<serde_json::Value>>;

    fn supports_structured_output(&self) -> bool;
}

pub trait IntoRig<T> {
    fn into_rig(self, model: &dyn CompletionModelExt) -> Result<T>;
}

impl IntoRig<rig::completion::CompletionRequest> for AICompletionRequest {
    fn into_rig(
        self,
        model: &dyn CompletionModelExt,
    ) -> Result<rig::completion::CompletionRequest> {
        let mut request = rig::completion::CompletionRequest {
            preamble: self.system,
            chat_history: rig::one_or_many::OneOrMany::many(
                self.messages
                    .into_iter()
                    .filter_map(|message| message.into_rig(model).transpose())
                    .collect::<Result<Vec<rig::completion::Message>>>()?,
            )
            .map_err(|_| RigError::EmptyList)?,
            documents: vec![],
            tools: self
                .tools
                .into_iter()
                .map(|tool| tool.into_rig(model))
                .collect::<Result<Vec<rig::completion::ToolDefinition>>>()?,
            temperature: self.temperature,
            max_tokens: self.max_tokens,
            tool_choice: None,
            additional_params: None,
        };

        if let Some(reasoning) = self.reasoning {
            if let Some(reasoning_value) = model.reasoning(reasoning, self.max_tokens)? {
                request.additional_params.merge(reasoning_value);
            }
        }

        if let Some(structured_output) = self.structured_output {
            if let Some(structured_output_value) = model.structured_output(structured_output)? {
                request.additional_params.merge(structured_output_value);
            }
        }

        Ok(request)
    }
}
```

joshua-mo-143:
> I personally use the system a lot so this is highly disruptive for me but I understand that it might be a PITA to maintain it.
> 
> Ideally you keep supporting this, but at minimum @joshua-mo-143 I would ask for a macro to automatically implement `CompletionModel` over the enum so at least it would be painless to migrate.
> 
> Just a sample of how I use it:
> 
> <snip>

Ok, before this gets out of hand since this update is proving to be relatively disruptive: there **will** be a replacement. We have already discussed this internally and are fine with adding a replacement, the primary thing is just making the abstraction easier to maintain because at the moment it's basically a giant ball of type unsafety with no safety net (that also forces other parts of the codebase to depend on it) for the sake of convenience.

We will also additionally be writing documentation in the new Rig playbook specifically on the dynamic model client builder and how users can implement it themselves should they need/want to.

Issue #1056, opened 2025-11-16
feat: Otel is too verbose and expensive


Sytten:
- [ ] I have looked for existing issues (including closed) about this

## Feature Request
Make the Otel records configurable.

### Motivation
Otel is adding a lot of crap to spans which reflect in all the http debug logs after that, quite annoying and can leak PII (specially the messages). It is also costly to do a bunch of JSON serialization all the time (I did partially fix that in #1052)

### Proposal
I am not a tracing expert, I don't really know if it is possible to disable recording fields. Otherwise it could be a feature flag.

### Alternatives
I guess the consumer can filter those?

joshua-mo-143:
Yeah we should probably add a feature flag or way to configure this. It's been low priority for a while but I think a way to configure this would not go amiss

Sytten:
I will need to dig more on how to do that. I think we might need a dynamic configuration for it per model or something if we want targeted thing.

joshua-mo-143:
Had a look into this.

As far as I'm concerned from what I have discovered, the message logging in GenAI SemConv is optional - so we should be safe to remove the message logging from spans without any worry as it adds an extremely high amount of cardinality. We should probably prefer the messages/requests to be in logs, instead.

This is probably the millionth breaking change we've made, but it's likely a correct one given that not *everyone* (read: almost nobody at all) needs the entire message history in their span telemetry.

PR #1108, opened 2025-12-02
fix(streaming): use .instrument() instead of span.enter() to prevent span leak

heeen:
## Summary

This PR fixes a tracing span leak issue in the streaming implementation where `invoke_agent` span context leaks to unrelated concurrent tasks.

Related to #1056 (though this is a different bug - context leakage vs verbosity).

### The Problem

Using `span.enter()` inside `async_stream::stream!` (line 203 in `streaming.rs`) holds the span guard across yield points. This causes thread-local span context to leak to other concurrent tasks running on the same thread, resulting in unrelated log messages being incorrectly prefixed with the `invoke_agent{...}` span.

For example, logs from completely unrelated modules would appear like:
```
invoke_agent{gen_ai.operation.name="invoke_agent" ...}: some_other_module: Unrelated log message
```

### The Fix

Replace `span.enter()` with `.instrument(agent_span)` from `tracing-futures`. The `Instrument` trait was already imported but not being used for the outer `agent_span`.

This change properly propagates span context through the async task system without thread-local pollution.

### Changes

- Removed `let _guard = agent_span.enter();` inside the `async_stream::stream!` macro
- Wrapped the stream with `.instrument(agent_span)` after creation
- Added `test_span_leak` example that verifies the fix

### Testing

The new `test_span_leak` example runs a background logger concurrent with a streaming request and verifies that background logs are not polluted with the `invoke_agent` span.

Run with:
```bash
ANTHROPIC_API_KEY=<key> cargo run --example test_span_leak
```

### References

- [tracing documentation on async code](https://docs.rs/tracing/latest/tracing/span/struct.Span.html#in-asynchronous-code)
- [rust-clippy#8722](https://github.com/rust-lang/rust-clippy/issues/8722) - Discussion about adding a clippy lint for this anti-pattern
- This is a well-known pitfall with `span.enter()` in async contexts

### DIFF
--- a/rig-core/src/agent/prompt_request/streaming.rs
+++ b/rig-core/src/agent/prompt_request/streaming.rs
@@ -199,8 +199,13 @@ where
 
         let cancel_signal = CancelSignal::new();
 
-        Box::pin(async_stream::stream! {
-            let _guard = agent_span.enter();
+        // NOTE: We use .instrument(agent_span) instead of span.enter() to avoid
+        // span context leaking to other concurrent tasks. Using span.enter() inside
+        // async_stream::stream! holds the guard across yield points, which causes
+        // thread-local span context to leak when other tasks run on the same thread.
+        // See: https://docs.rs/tracing/latest/tracing/span/struct.Span.html#in-asynchronous-code
+        // See also: https://github.com/rust-lang/rust-clippy/issues/8722
+        let stream = async_stream::stream! {
             let mut current_prompt = prompt.clone();
             let mut did_call_tool = false;
 
@@ -435,8 +440,9 @@ where
                     prompt: last_prompt_error.clone().into(),
                 }).into());
             }
+        };
 
-        })
+        Box::pin(stream.instrument(agent_span))
     }
 }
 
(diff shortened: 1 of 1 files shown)


joshua-mo-143:
Hi, thanks for your PR!

Can you add your example as a test in the streaming prompt module? This fits more as a test rather than an example. We are currently trying to minimise how many examples get added. As you can see from the examples folder, there's a gargantuan amount of them.

The test should be annotated with a macro like this to avoid being ran on CI (it'll still get run through `clippy`, so no need to worry about that):

```
#[tokio::test]
#[ignore = "This requires an API key"]
async fn test() {
    ..
}
```

heeen:
Done! Moved the test from an example to a proper test in the streaming module:

```rust
#[tokio::test(flavor = "current_thread")]
#[ignore = "This requires an API key"]
async fn test_span_context_isolation() {
    // ...
}
```

The test verifies that span context doesn't leak to concurrent tasks during streaming.

joshua-mo-143:
Looks like this is missing `cargo fmt`

heeen:
I think that latest failed CI run is unrelated

PR #1115, opened 2025-12-02
feat(rig-1076): Providers should route all requests through `client::Client`

### LINKED ISSUE #1107: feat: Providers should route all requests through `client::Client`
Currently, some providers use the `http_client()` getter method on `client::Client` to directly access the internal http client and make requests, while most use i.e. `client::Client::send`. This has the potential to cause bugs which are difficult to track down, such as missing headers or other behavior which has been provided by `client::Client`.

The `http_client()` getter should be removed and the corresponding field made private so that all requests go through `client::Client`'s methods, for consistency and maintainability.

FayCarsons:
fixes #1107


joshua-mo-143:
lgtm

PR #1117, opened 2025-12-04
fix: OpenAI Chat Completions API tool result format and strict mode support

lleukkun:
## Summary

Fix tool result serialization for OpenAI Chat Completions API and add support for strict mode tool schemas.

### Tool Result Format Fix
- Change `ToolResult.content` from `OneOrMany<ToolResultContent>` to `String` in both OpenAI and OpenRouter providers
- The Chat Completions API expects tool message content as a plain string, not an array of content blocks
- This fixes errors like `Input should be a valid string` when using OpenRouter with models like DeepSeek

### Strict Mode Support
- Add `FunctionDefinition` struct with optional `strict` field
- Move existing `sanitize_schema` from `responses_api` to `openai/mod.rs` with `pub(crate)` visibility so it can be shared by both completion and responses_api modules
- Add `strict_tools` field to `CompletionModel` (default: false)
- Add `with_strict_tools()` builder method to opt-in
- Add `OpenAIRequestParams` and `OpenRouterRequestParams` structs for passing strict_tools through request building

## Usage

```rust
// OpenAI
let completion_model = client
    .completion_model("gpt-4o")
    .completions_api()
    .with_strict_tools();

// OpenRouter
let completion_model = client
    .completion_model("openai/gpt-4o")
    .with_strict_tools();
```

## Test plan

- [x] All existing tests pass (171 tests)
- [x] Clippy passes with no warnings
- [x] Manually tested with OpenRouter + DeepSeek - tool calls now work correctly

### DIFF
--- a/rig-core/src/providers/openai/completion/mod.rs
+++ b/rig-core/src/providers/openai/completion/mod.rs
@@ -538,10 +538,10 @@ impl TryFrom<OneOrMany<message::AssistantContent>> for Vec<Message> {
                     message::AssistantContent::Text(text) => texts.push(text),
                     message::AssistantContent::ToolCall(tool_call) => tools.push(tool_call),
                     message::AssistantContent::Reasoning(_) => {
-                        unimplemented!("The OpenAI Completions API doesn't support reasoning!");
+                        panic!("The OpenAI Completions API doesn't support reasoning!");
                     }
                     message::AssistantContent::Image(_) => {
-                        unimplemented!(
+                        panic!(
                             "The OpenAI Completions API doesn't support image content in assistant messages!"
                         );
                     }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9728,7 +9728,7 @@ checksum = "57397d16646700483b67d2dd6511d79318f9d057fdbd21a4066aeac8b41d310a"
 
 [[package]]
 name = "rig-bedrock"
-version = "0.3.7"
+version = "0.3.8"
 dependencies = [
  "anyhow",
  "async-stream",
@@ -9737,7 +9737,7 @@ dependencies = [
  "aws-smithy-types",
  "base64 0.22.1",
  "reqwest 0.12.24",
- "rig-core 0.25.0",
+ "rig-core 0.26.0",
  "rig-derive",
  "schemars 1.0.4",
  "serde",
@@ -9774,7 +9774,7 @@ dependencies = [
 
 [[package]]
 name = "rig-core"
-version = "0.25.0"
+version = "0.26.0"
 dependencies = [
  "anyhow",
  "as-any",
@@ -9828,7 +9828,7 @@ dependencies = [
  "indoc",
  "proc-macro2",
  "quote",
- "rig-core 0.25.0",
+ "rig-core 0.26.0",
  "serde",
  "serde_json",
  "syn 2.0.106",
@@ -9838,13 +9838,13 @@ dependencies = [
 
 [[package]]
 name = "rig-eternalai"
-version = "0.3.13"
+version = "0.3.14"
 dependencies = [
  "anyhow",
  "async-stream",
  "ethers",
  "reqwest 0.12.24",
- "rig-core 0.25.0",
+ "rig-core 0.26.0",
  "schemars 1.0.4",
  "serde",
  "serde_json",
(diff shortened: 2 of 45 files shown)


rig-core/src/providers/openai/completion/mod.rs
```diff
@@ -267,21 +267,106 @@ pub enum ToolType {
...
+                strict: None,
+            },
         }
     }
 }
 
+/// Recursively ensures all object schemas respect OpenAI strict mode restrictions.
+/// When strict mode is enabled, all objects must have `additionalProperties: false`
+/// and all properties must be in the `required` array.
+fn sanitize_schema_for_strict(schema: &mut serde_json::Value) {
```
joshua-mo-143: This function already exists in the OpenAI Responses module.

Instead of adding another version of this, maybe we should move the already-existing function to the `mod.rs` file?

rig-core/src/providers/openai/completion/mod.rs
```diff
@@ -267,21 +267,106 @@ pub enum ToolType {
...
+    let serde_json::Value::Object(obj) = schema else {
+        return;
+    };
+
+    let is_object_schema = obj.get("type")
+        == Some(&serde_json::Value::String("object".to_string()))
+        || obj.contains_key("properties");
+
+    // Add additionalProperties: false for objects
```
joshua-mo-143: Please remove these non-docstring comments.

If we can't figure out what is happening from the code itself, we shouldn't be adding it to the codebase. The only reason we should add non-docstring comments is to provide context for *why* something has been implemented a given way (as opposed to something else)

rig-core/src/providers/openai/completion/mod.rs
lleukkun: yes, I'll do the change!

rig-core/src/providers/openai/completion/mod.rs
lleukkun: Aye, will clean up.

rig-core/src/providers/openai/completion/mod.rs
```diff
@@ -145,7 +145,7 @@ pub enum Message {
     #[serde(rename = "tool")]
     ToolResult {
         tool_call_id: String,
-        content: OneOrMany<ToolResultContent>,
+        content: String,
```
joshua-mo-143: This does not seem quite correct: https://platform.openai.com/docs/api-reference/chat/create#chat_create-messages-tool_message

I guess *technically* the correct solution to this would be to create an enum that's either just a string, or a vec of structs that have the string type (you could do this with a one-variant enum that has type tags on it).

joshua-mo-143 (changes requested):
Looks mostly ok to me, have commented on something that I seem to have missed out on from the first review round

rig-core/src/providers/openai/completion/mod.rs
lleukkun: Yes, I'll have to think this over again. There are also some openai responses api encrypted reasoning content issues but I'll make another PR about them. These APIs are surprisingly complex considering that the model IO is basically just text...

joshua-mo-143: Yeah, that's the trick unfortunately haha. I do wish that they'd stop changing something every week or 2 but such is the nature of the LLM space

lleukkun: Oh dear. This ballooned. I have now ~300 line additional diff that fixes openai responses api tool calls when encrypted reasoning blocks are requested, and adds serverside tool calls (for web search) to both anthropic and openai. I will still need to fix this actual issue here, but this will have to wait until tomorrow.

lleukkun: @joshua-mo-143 I've pushed a commit to reintroduce array support. Defaults to string, but arrays can be chosen using with_tool_result_array_content() in the CompletionModel. At least seems to work with openrouter + deepseek-v3.2 (using strings, with deepinfra as provider).

rig-core/src/providers/openai/completion/mod.rs
```diff
@@ -252,6 +252,55 @@ impl From<String> for ToolResultContent {
...
 }
 
+#[derive(Debug, Deserialize, Clone, PartialEq)]
+#[serde(untagged)]
+pub enum ToolResultContentValue {
+    Array(Vec<ToolResultContent>),
+    String(String),
+}
+
+impl Serialize for ToolResultContentValue {
```
joshua-mo-143: Unnecessary manual impl. Given that the type is already untagged, deriving Serialize automatically *should* give the correct response

lleukkun: done. merged main and resolved the merge conflict by introducing identical oneOf -> anyOf change to the relocated method.

Issue #708, opened 2025-08-16
Error when using multi-turn: Expected the last role to be User or Tool (or Assistant with prefix True).


sopaco:
`` rs
let sql_executor = SqlExecutor::default();
        let text_saver = TextSaver::default();

        let llm_client = providers::mistral::Client::from_env();

        // Create agent with tools
        let agent = llm_client
            .agent("devstral-small-latest")
            .preamble(format!(include_str!("prompts/analysis_trace_sys.tpl"), platform).as_str())
            .max_tokens(100000)
            .temperature(0.1)
            .tool(sql_executor)
            .tool(text_saver)
            .build();

        // Agent is now ready to use with the configured tools
        let kb_all_about_perfetto_sql = include_str!("./KB/all-about-perfetto-sql.md");

        // Create the initial prompt
        let initial_prompt = format!(
            include_str!("prompts/analysis_trace.tpl"),
            trace_path, out_path, kb_all_about_perfetto_sql
        );

        // Start the agentic loop
        // self.agentic_loop(&agent, initial_prompt, 0).await?;
        let response = agent.prompt(&initial_prompt).multi_turn(1).await?;
```

My code is as above, and it reports an error when running:

``` log
Error: CompletionError: ProviderError: {"object":"error","message":"Expected last role User or Tool (or Assistant with prefix True) for serving but got assistant","type":"invalid_request_message_order","param":null,"code":"3230"}

Caused by:
    ProviderError: {"object":"error","message":"Expected last role User or Tool (or Assistant with prefix True) for serving but got assistant","type":"invalid_request_message_order","param":null,"code":"3230"}

```

`rig-core: 0.17.1`

sopaco:
It does indeed appear that the most recent entry in chat_history does not contain a tool message. However, based on the execution flow in prompt_request, it seems that the tool message was actually retrieved.

<img width="1644" height="570" alt="Image" src="https://github.com/user-attachments/assets/ec247e11-6415-426c-b222-197cb6ddef59" />

<img width="1567" height="1020" alt="Image" src="https://github.com/user-attachments/assets/dc54f104-625e-4163-8657-6ffdff8b4ad3" />

<img width="915" height="304" alt="Image" src="https://github.com/user-attachments/assets/83e4e361-3c36-465e-867b-9e8ebacf6978" />

0xMochan:
[...]

PR #1130, opened 2025-12-08
feat: Tool message role for Mistral

aleksmeshr:
https://github.com/0xPlaygrounds/rig/issues/708

- Define the `Tool` message role for Mistral
- Translates `User` messages having a `ToolResult` content into a message with the Tool role.
- That fixes the error `Expected last role User or Tool (or Assistant with prefix True) for serving but got assistant` reported by Mistral when a User message with a ToolResult was provided


joshua-mo-143:
Lgtm - ~~have you tested this?~~ will test tomorrow then merge

Issue #1126, opened 2025-12-06
bug: DeepSeek provider fails with "can only flatten structs and maps (got a float)" error


wangeguo:
### Description
The DeepSeek provider fails to serialize completion requests due to incorrect use of `#[serde(flatten)]` attribute on the `temperature` field which is of type `Option<f64>`.

### Error Message

```
CompletionError: JsonError: can only flatten structs and maps (got a float)
```

### Root Cause

In `rig-core/src/providers/deepseek.rs`, line 436-437:

```rust
#[serde(flatten, skip_serializing_if = "Option::is_none")]
temperature: Option<f64>,
```

The `#[serde(flatten)]` attribute is only valid for struct or map types. Using it on a scalar type like `f64` causes serde to fail during serialization.

### Suggested Fix

Remove the `flatten` attribute from the `temperature` field:

```diff
-    #[serde(flatten, skip_serializing_if = "Option::is_none")]
+    #[serde(skip_serializing_if = "Option::is_none")]
     temperature: Option<f64>,
```

### Steps to Reproduce

1. Create a DeepSeek client with rig-core
2. Make any completion request with a temperature parameter
3. Observe the serialization error

### Environment

- rig-core version: 0.26.0
- Rust version: 1.91.0

joshua-mo-143:
Hi, thanks for opening the issue - looks like an oversight from a previous merge.

Will fix this immediately

PR #1141, opened 2025-12-10
fix: OpenRouter completion request broken due to unnecessary flatten

ronaldslc:
Fixes https://github.com/0xPlaygrounds/rig/issues/1126 for OpenRouter


joshua-mo-143:
Hi, thanks for your PR!

Please remove the Cargo.lock changes. We plan on updating the Cargo.lock separately to avoid bundling it with small changes (and therefore if something goes wrong while updating the lockfile, we won't have to revert the other additional useful changes that were made).

PR #1154, opened 2025-12-11
ToolCall Signature and additional parameters

Sytten:
- Added a signature for the ToolCall
- Added additional parameters for ToolCall (this allows extending in the future for other providers)
- Added builder methods since we are starting to have many parameters and it is getting annoying to change the code in many places


joshua-mo-143:
lgtm, will need to discuss internally on Monday and make sure the other Rig team members are aware that this is something that we *probably* want to be merged but other than that this seems fine, given that this appears to be blocking sending reasoning signatures with gemini

PR #1157, opened 2025-12-11
feat(rig-1093): helper methods for making single-text messages

### LINKED ISSUE #1153: feat: helper methods for making single-text user/assistant messages

## Feature Request

Currently, creating user or assistant messages that have a single text in them is quite a bit more clunky than it should be:

```rust
let message = Message::User {
    content: OneOrMany::one(UserContent::text("foo"))
};
```

We could really do something that's more like this:

```
let message = Message::user_text("foo");
```

### Motivation

Improve DX

### Proposal

See the aforementioned code snippet.

joshua-mo-143:
Fixes #1153

### DIFF
--- a/rig/rig-core/src/completion/message.rs
+++ b/rig/rig-core/src/completion/message.rs
@@ -44,6 +44,29 @@ pub enum Message {
     },
 }
 
+impl Message {
+    /// A helper method to create a user message with a single piece of text in it.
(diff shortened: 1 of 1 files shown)


gold-silver-copper:
This currently exists as `Message::user(...)` and `Message::assistant(...)`

PR #1159, opened 2025-12-12
fix(groq): rename StreamingOptions to StreamOptions

yazaldefilimone:


joshua-mo-143 (approved):
lgtm, not sure how I missed this

PR #1166, opened 2025-12-16
Respect custom http headers for outgoing client requests

simonjanssen:
I am operating in a env where I need to set custom headers for my LLM providers (e.g. user-agent, etc.). Specifically, I need to set a custom `Accept` header for streaming updates via OpenAI clients.

Setting default headers for my reqwest client had no effect as the `ProviderClient` treats http client and http headers separately for http request building.

In the `ClientBuilder` [these headers default to an empty map](https://github.com/0xPlaygrounds/rig/blob/133cad1671e703f42449e52650a993a11a433fd6/rig/rig-core/src/client/mod.rs#L450). As far as I can tell there is currently no public setter method to set anything but the default headers. 

What I've added/changed:
- added a demo `http_header` setter method for `ClientBuilder` to add additional, custom http headers
- `GenericEventSource::new()` reflects custom `Accept` headers set this way instead of overwriting

Tested the proposed changes with this setup which worked just fine:
```rust
let client = openai::CompletionsClient::<reqwest::Client>::builder()
    .base_url(my_base_url)
    .http_headers(my_custom_headers) // added
    .http_client(my_reqwest_client)
    .api_key(my_api_key)
    .build()?;
```

### DIFF
--- a/rig/rig-core/src/client/mod.rs
+++ b/rig/rig-core/src/client/mod.rs
@@ -521,10 +521,7 @@ where
 
     /// Set the HTTP headers used in this client
     pub fn http_headers(self, headers: HeaderMap) -> Self {
-        Self {
-            headers,
-            ..self
-        }
+        Self { headers, ..self }
     }
 
     pub(crate) fn headers_mut(&mut self) -> &mut HeaderMap {


joshua-mo-143 (changes requested):
Needs `cargo fmt` then lgtm

simonjanssen:
> Needs `cargo fmt` then lgtm

done

joshua-mo-143 (approved):
lgtm, thanks

Issue #1189, opened 2025-12-24
bug: Panic caused by Option::Unwrap() called on None value in Ollama Provider EmbeddingModel


corysabol:
- [x] I have looked for existing issues (including closed) about this

## Bug Report

**Rig version**: v0.27.0
**System information:**
```
Software:

      System Version: macOS 15.6 (24G84)
      Kernel Version: Darwin 24.6.0

Hardware:
      Model Name: MacBook Air
      Chip: Apple M4
      Total Number of Cores: 10 (4 performance and 6 efficiency)
      Memory: 32 GB
```

When constructing an `EmbeddingModel` using the Ollama client a panic is triggered.

**Error message**
```
thread 'main' panicked at /Users/corysabol/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/rig-core-0.27.0/src/providers/ollama.rs:211:47:
called `Option::unwrap()` on a `None` value
```

**Problematic code**
https://github.com/0xPlaygrounds/rig/blob/7d0f19d56fcc280d154d4026aacb3f8f9fb2fe10/rig/rig-core/src/providers/ollama.rs#L210-L212
Line 211 in `providers/ollama.rs` triggers a panic due to calling `.unwrap()` on a `None` value.
In this case the `None` value is `dims`.

The panic occurs when constructing an embedding model as follows:
```rust
let embedding_model = ollama_client.embedding_model("nomic-embed-text");
```

It is possible to work around this issue by using `embedding_model_with_ndims` instead:
```rust
let embedding_model = ollama_client.embedding_model_with_ndims("nomic-embed-text", 768);
```

## Reproduction

**main.rs**
```rust
use rig::client::Nothing;
use rig::prelude::*;
use rig::providers::ollama::Client;

#[tokio::main]
async fn main() -> Result<(), anyhow::Error> {
    let ollama_client = Client::from_val(Nothing);
    let embedding_model = ollama_client.embedding_model("nomic-embed-text");

    Ok(())
}
```
**Cargo.toml**
```toml
[package]
name = "rig_ollama_embedding_model_bug"
version = "0.1.0"
edition = "2024"

[dependencies]
anyhow = "1.0.100"
rig-core = "0.27.0"
tokio = { version = "1.48.0", features = ["full"] }
```

## Expected behavior

[...]

PR #1199, opened 2025-12-29
fix: ollama embedding_model dims Option panic.

KanonWY:
use ollama embedding_model， Missing model dimensions.
```rust
    let ollama_client = ollama::Client::from_env();
   // this will panic.
    let model = ollama_client.embedding_model("nomic-embed-text");
```
add function `model_dimensions_from_identifier` to give a default dimension.
maybe fix [1189](https://github.com/0xPlaygrounds/rig/issues/1189)

### DIFF
--- a/rig/rig-core/src/providers/ollama.rs
+++ b/rig/rig-core/src/providers/ollama.rs
@@ -147,6 +147,14 @@ enum ApiResponse<T> {
 pub const ALL_MINILM: &str = "all-minilm";
 pub const NOMIC_EMBED_TEXT: &str = "nomic-embed-text";
 
+fn model_dimensions_from_identifier(identifier: &str) -> Option<usize> {
+    match identifier {
+        ALL_MINILM => Some(384),
+        NOMIC_EMBED_TEXT => Some(768),
+        _ => None,
+    }
+}
+
 #[derive(Debug, Serialize, Deserialize)]
 pub struct EmbeddingResponse {
     pub model: String,
@@ -208,7 +216,11 @@ where
     type Client = Client<T>;
 
     fn make(client: &Self::Client, model: impl Into<String>, dims: Option<usize>) -> Self {
-        Self::new(client.clone(), model, dims.unwrap())
+        let model = model.into();
+        let dims = dims
+            .or(model_dimensions_from_identifier(&model))
+            .unwrap_or_default();
+        Self::new(client.clone(), model, dims)
     }
 
     const MAX_DOCUMENTS: usize = 1024;


rig/rig-core/src/providers/ollama.rs
```diff
@@ -147,6 +147,14 @@ enum ApiResponse<T> {
...
 pub const NOMIC_EMBED_TEXT: &str = "nomic-embed-text";
 
+fn model_dimensions_from_identifier(identifier: &str) -> Option<usize> {
+    match identifier {
+        ALL_MINILM => Some(384),
+        NOMIC_EMBED_TEXT => Some(768),
+        _ => None,
+    }
+}
+
```
joshua-mo-143, line 157: I think this part is largely unmaintainable, so we should avoid adding this in. It is usually quite trivial to find the dimension sizes

joshua-mo-143 (changes requested):
Looks OK. See my comment

rig/rig-core/src/providers/ollama.rs
KanonWY: yes, I think this part is largely unmaintainable too. A better approach is to use `embedding_model_with_ndims`. I added it this way because I noticed that other modules did the same. [openai/embedding.rs:row59](https://github.com/0xPlaygrounds/rig/blob/main/rig/rig-core/src/providers/openai/embedding.rs)

joshua-mo-143: Yeah OpenAI only has 2/3 models so it's fine. Unfortunately for us, ollama seems to have an absolute tonne of them which means we can't reliably keep track of all of them

codename-irvin: Understand the concern with maintainability, but something should probably be done as the API is currently broken. Calling `ollama_client.embedding_model(...)` is guaranteed to fail.

joshua-mo-143: > Understand the concern with maintainability, but something should probably be done as the API is currently broken. Calling `ollama_client.embedding_model(...)` is guaranteed to fail.

Hmmm, yeah I guess. To be honest I was mostly just holding off from merging because I wasn't sure how long I should wait for a response.

joshua-mo-143:
Merging as this unblocks users with no apparent side-effects (besides the embedding model being initialised with `None` dimensions).

If we get any issues, we can just open a new ticket

rig/rig-core/src/providers/ollama.rs
codename-irvin: Another possible solution is that if ollama_client can't properly support the `embedding_model` method, then calling it could produce a compiler error via `compile_error!`. The error message could urge consumers to use `embedding_model_with_ndims`.

But, maintaining feature parity with other clients is ideal. Otherwise, the abstraction is a bit leaky.

Issue #1206, opened 2026-01-04
bug: Inconsistent Content-Type header between `send` and `send_streaming`


Phoenix500526:
- [x] I have looked for existing issues (including closed) about this

## Bug Report

`send_streaming` does not set the `Content-Type: application/json` header, unlike `send`. This inconsistency can cause issues （503 error） with some API provider gateways (like OneRouter) that strictly require this header. 

## Reproduction

Looking at the current implementation in `client/mod.rs`:

```rust
// send() sets Content-Type
fn send<T>(
    &self,
    mut req: Request<T>,
) -> impl Future<Output = http_client::Result<http_client::Response>> + WasmCompatSend
where
    T: Into<Bytes>,
{
    req.headers_mut().insert(
        http::header::CONTENT_TYPE,
        http::HeaderValue::from_static("application/json"),
    );
    self.http_client.send(req)
}

// send_streaming() does NOT set Content-Type
fn send_streaming<T>(
    &self,
    req: Request<T>,  // not mut
) -> impl Future<Output = http_client::Result<http_client::StreamingResponse>> + WasmCompatSend
where
    T: Into<Bytes>,
{
    self.http_client.send_streaming(req)  // no Content-Type header
}
```

## Expected behavior

Both `send` and `send_streaming` should consistently set the `Content-Type: application/json` header, as streaming requests also send JSON request bodies.

## Fix

```rust
fn send_streaming<T>(
    &self,
    mut req: Request<T>,
) -> impl Future<Output = http_client::Result<http_client::StreamingResponse>> + WasmCompatSend
where
    T: Into<Bytes>,
{
    req.headers_mut().insert(
        http::header::CONTENT_TYPE,
        http::HeaderValue::from_static("application/json"),
    );

    self.http_client.send_streaming(req)
}
```

## Additional context

While most API providers may work without an explicit Content-Type header (inferring it from the request body), having consistent behavior between `send` and `send_streaming` is important for:

1. **API compatibility**: Some providers may require explicit Content-Type headers
2. **Code consistency**: Both methods handle JSON request bodies identically
3. **Best practices**: Explicitly declaring Content-Type is standard practice for HTTP APIs

joshua-mo-143:
Already covered by #1196

joshua-mo-143:
Closed as duplicate of #1174

PR #1207, opened 2026-01-04
fix(client): add Content-Type header to send_streaming

Phoenix500526:
Unlike `send`, `send_streaming` was not setting Content-Type header, causing 503 errors on strict API providers like OpenRouter.

Closes: #1206


joshua-mo-143:
Closing because this is a duplicate of #1196

PR #1211, opened 2026-01-04
fix: some completion providers send usage chunks with 0 completion choices causing 0 reported usage

dxvid-pts:
Fixes: https://github.com/0xPlaygrounds/rig/issues/1212

Proposed fix:

  - In rig-core/src/providers/openai/completion/streaming.rs, capture data.usage before the choices.first() early-continue, e.g.

    if let Some(usage) = data.usage.as_ref() {
        final_usage = Some(usage.clone());
    }
  - Add regression test: deserialize a chunk with choices: [] and usage != null and ensure the final streamed usage is not lost.


dxvid-pts:
pipeline should be fixed now

PR #1221, opened 2026-01-05
fix(embeddings): respect token limits in EmbeddingsBuilder batching

### LINKED ISSUE #462: bug: EmbeddingBuilder::build() Exceeds OpenAI Token Limit Due to Lack of Token-Based Chunking

## Bug Report

When using `EmbeddingBuilder::build()` with OpenAI's `text-embedding-3-small` model, I encountered the following error:

```json
Error: ProviderError: {
    "error": {
        "message": "Requested 510944 tokens, max 300000 tokens per request",
        "type": "max_tokens_per_request",
        "param": null,
        "code": "max_tokens_per_request"
    }

````

This occurred despite the implementation's provision to chunk requests based on `M::MAX_DOCUMENTS`. Upon investigation, it appears that while the method chunks documents by count, it doesn't account for the total token count per batch. Given that OpenAI's API imposes a maximum token limit per request, this oversight can lead to such errors when the cumulative tokens of the documents in a batch exceed the allowed limit.

## Reproduction

```rust
let mut builder = EmbeddingsBuilder::new(model.clone());
for (i, chunk) in chunks.into_iter().enumerate() {
    builder = builder.document(Document {
        id: format!("chunk_{}", i),
        content: chunk,
    })?;
}
let embeddings = builder.build().await?;
```

Where `chunks` are strings of roughly 2000 characters each from a 500 pages PDF.

## Expected behavior

The builder should automatically group documents into batches that respect OpenAI's total token limit per request (e.g., 300,000 tokens), avoiding API errors due to oversized requests.

## Additional context

In my case, the issue was resolved by manually splitting the documents into two batches and calling `build()` separately on each. However, this workaround defeats the purpose of automatic batching, and a proper fix would be to support token-aware batching logic inside the `EmbeddingBuilder`.

godnight10061:
Fixes #462

## Summary
`EmbeddingsBuilder::build()` previously batched only by `M::MAX_DOCUMENTS`. For providers like OpenAI embeddings, requests can also fail when the combined input exceeds the provider’s per-request token budget.

## Changes
- Add `EmbeddingModel::max_tokens_per_request()` (default `None`) so providers can expose a per-request token budget.
- Batch by both `M::MAX_DOCUMENTS` and `max_tokens_per_request()` (uses `text.len()` as a conservative proxy for tokens to avoid adding a tokenizer dependency).
- Set OpenAI embedding models to `Some(300_000)` (based on the provider error in the issue).
- Add a regression test covering token-budget batching.

## Test
- `cargo test -p rig-core --lib`

### DIFF
--- a/rig/rig-core/src/embeddings/builder.rs
+++ b/rig/rig-core/src/embeddings/builder.rs
@@ -111,15 +111,48 @@ where
             texts.push((i, doc_texts));
         }
 
+        // Flatten the texts while keeping track of the document index.
+        let mut flat_texts = Vec::new();
+        for (i, doc_texts) in texts.into_iter() {
+            for text in doc_texts {
+                flat_texts.push((i, text));
+            }
+        }
+
+        let max_documents = M::MAX_DOCUMENTS;
+        let max_tokens = self.model.max_tokens_per_request().unwrap_or(usize::MAX);
+
+        // Group them into batches.
+        let mut batches = Vec::new();
+        let mut current_batch = Vec::new();
+        let mut current_tokens = 0;
+
+        for (i, text) in flat_texts {
+            // Simple KISS estimate: bytes = tokens (upper bound)
+            let text_tokens = text.len();
+
+            // Check if adding this text would exceed the limit
+            if !current_batch.is_empty()
+                && (current_batch.len() >= max_documents
+                    || current_tokens + text_tokens > max_tokens)
+            {
+                batches.push(current_batch);
+                current_batch = Vec::new();
+                current_tokens = 0;
+            }
+
+            current_tokens += text_tokens;
+            current_batch.push((i, text));
+        }
+        if !current_batch.is_empty() {
+            batches.push(current_batch);
+        }
+
         // Compute the embeddings.
-        let mut embeddings = stream::iter(texts.into_iter())
-            // Merge the texts of each document into a single list of texts.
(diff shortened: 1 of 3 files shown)


gold-silver-copper:
Thanks for this, and sorry it sat so long without a decision.

Closing because the issue it targets is no longer open: #462 was closed as `not planned` on 2026-03-09, after the discussion there settled on a different approach. @ProHaller proposed a split-and-retry on `build()` — split the batch when the provider rejects it — and the reasoning was that it avoids paying for tokenization on every request. This PR's `max_tokens_per_request()` + `text.len()` proxy takes the opposite trade: it estimates up front on every build, and the estimate is necessarily conservative, so it will split batches that would have been accepted.

Nothing here was wrong, and the regression test is a good one — it's just that the direction moved and the branch has been sitting against a `main` that has since moved a long way (embeddings now go through `EmbeddingResponse` and the `OpenAIEmbeddingsCompatible` seam, so the OpenAI-side change would need reworking regardless).

If you'd like to pick up the split-and-retry shape instead, that would be very welcome — it's still unimplemented, and reopening #462 with that plan would be the right starting point.

ProHaller:
Thank you for the feedback. It's probably the safer implementation anyway.

gold-silver-copper:
@ProHaller hai :D 

sorry for the slop comment, was just closing some older issues

PR #1222, opened 2026-01-05
add tool name to tool call delta streaming events

bebiksior:
Adds tool name to `ToolCallDelta` streaming events so apps can display which tool is being called during streaming

Uses a `ToolCallDeltaContent` enum with `Name(String)` and `Delta(String)` variants since providers send either the name (first event) or argument deltas (subsequent events)

```rust
pub enum ToolCallDeltaContent {
    Name(String),
    Delta(String),
}

RawStreamingChoice::ToolCallDelta {
    id: String,
    content: ToolCallDeltaContent,
}
```

### DIFF
--- a/rig-integrations/rig-bedrock/src/streaming.rs
+++ b/rig-integrations/rig-bedrock/src/streaming.rs
@@ -6,7 +6,7 @@ use rig::completion::GetTokenUsage;
 use rig::streaming::StreamingCompletionResponse;
 use rig::{
     completion::CompletionError,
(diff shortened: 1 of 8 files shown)


Sytten:
@joshua-mo-143 FYI @bebiksior here is working for caido, I support this PR. This is a breaking change though so let us know if you prefer to have a nullable name instead but this new enum is IMO better to convey the intent of the delta.

joshua-mo-143 (approved):
lgtm!

PR #1230, opened 2026-01-09
Feat/gemini interactions api

iskng:
- Add Gemini Interactions API support, including SSE streaming and deep research example.
  - Add Google Search, URL context, and code execution tooling + helpers for Interactions.
  - Add citation helpers and exchange grouping (with best‑effort fallback when call_id is missing).
  - Update examples to cover interactions tools and neutral prompts.

### DIFF
--- a/rig/rig-core/src/providers/gemini/client.rs
+++ b/rig/rig-core/src/providers/gemini/client.rs
@@ -147,6 +147,7 @@ impl<H> Capabilities<H> for GeminiInteractionsExt {
     type Completion = Capable<super::interactions_api::InteractionsCompletionModel<H>>;
     type Embeddings = Capable<super::embedding::EmbeddingModel>;
     type Transcription = Capable<super::transcription::TranscriptionModel>;
+    type ModelListing = Nothing;
 
     #[cfg(feature = "image")]
     type ImageGeneration = Nothing;
--- a/rig/rig-core/src/providers/gemini/interactions_api/mod.rs
+++ b/rig/rig-core/src/providers/gemini/interactions_api/mod.rs
@@ -482,6 +482,7 @@ impl TryFrom<Interaction> for completion::CompletionResponse<Interaction> {
             choice,
             usage,
             raw_response: response,
+            message_id: None,
         })
     }
 }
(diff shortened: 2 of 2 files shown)


iskng:
Added this because you can only use the deep research model from google with their interactions api.

rig/rig-core/examples/gemini_deep_research.rs
```diff
@@ -0,0 +1,278 @@
...
+    let params = deep_research_params();
+    let request = model
+        .completion_request(prompt)
+        .additional_params(serde_json::to_value(&params)?)
+        .build();
+
+    if use_streaming {
+        println!("== Deep Research (streaming) ==");
+        let mut state = StreamState::default();
+        let mut attempt = 0usize;
```
joshua-mo-143: `GenericEventSource` already handles retries

rig/rig-core/src/providers/gemini/client.rs
```diff
@@ -13,14 +13,27 @@ use std::fmt::Debug;
...
+#[derive(Debug, Default, Clone)]
+pub struct GeminiInteractionsExt {
+    api_key: String,
+}
+
+/// Builder marker for the Gemini Interactions client.
+#[derive(Debug, Default, Clone)]
+pub struct GeminiInteractionsBuilder;
+
+/// Wrapper type for Gemini API keys.
```
joshua-mo-143, line 34: Is there any reason why this needs a separate builder? We can just add an `.interactions_api()` converter method from the Gemini client typesig

rig/rig-core/src/providers/gemini/interactions_api/streaming.rs
```diff
@@ -0,0 +1,339 @@
...
+
+                        let data = serde_json::from_str::<InteractionSseEvent>(&message.data);
+
+                        let Ok(data) = data else {
+                            let err = data.unwrap_err();
+                            tracing::debug!(
+                                "Failed to deserialize interactions SSE event: {err}"
+                            );
+                            continue;
+                        };
```
joshua-mo-143: Can this be compressed? It seems redundant to have two separate statements for this.

rig/rig-core/src/providers/gemini/client.rs
iskng: `rig/rig-core/src/providers/gemini/client.rs:

  impl<H> Client<H> {
      /// Create an Interactions API client from this GenerateContent client.
      pub fn interactions_api(self) -> InteractionsClient<H> {
          let api_key = self.ext().api_key.clone();
          self.with_ext(GeminiInteractionsExt { api_key })
      }
  }`

We still need a distinct builder marker because Provider::Builder/ProviderBuilder::Output are one‑to‑one; dropping GeminiInteractionsBuilder would remove InteractionsClient::new/builder/from_env and force a conversion path (API change). Interactions also uses different auth handling (header vs query param), so it still needs its own Provider impl.

dropped the public InteractionsClientBuilder alias and steerd to .interactions_api().

joshua-mo-143: Oh right, I see.

OK let's keep this then

joshua-mo-143:
You can just set the capability as `Nothing` for model listing. We haven't rolled out the capability yet.

joshua-mo-143 (changes requested):
lgtm have tried locally - just needs to pass CI

joshua-mo-143:
Huh, it looks like I forgot to merge this before the release.

Apologies - this'll be merged now!

PR #1233, opened 2026-01-09
bug(1111): fix JSON value binding in rig-postgres

### LINKED ISSUE #1190: bug: PgSearchFilter fails to bind non-string serde_json::Value parameters in PostgreSQL queries
---

#### 🐞 Problem

When using `PostgresVectorStore`, filtering by metadata with non-string values (e.g., integers, booleans, floats) via `PgSearchFilter` causes runtime errors during query execution.

This happens because the current implementation passes `serde_json::Value` directly to `sqlx::Query::bind()`. However, **`sqlx` does not support binding `serde_json::Value` directly**—it requires concrete Rust types like `i64`, `f64`, `&str`, or `bool`.

Example that triggers the error:
```rust
let filter = PgSearchFilter::eq("user_id".into(), 123.into()); // Value::Number
// or
let filter = PgSearchFilter::eq("is_public".into(), true.into()); // Value::Bool
```

At runtime, PostgreSQL returns an error:
```
Vector search error: Datastore error: error returned from database: operator does not exist: Number= jsonb
```

#### 🔧 Proposed Fix

Introduce a helper macro `bind_json_params!` that pattern-matches on `serde_json::Value` and binds the appropriate native SQL type:

- `String` → `&str`
- `Number` → `i64` (if integer) or `f64` (if float)
- `Bool` → `bool`
- `Null` → `Option::<&str>::None`
- Fallback: bind as-is (for safety)

Then replace the manual `.fold(... .bind(param))` calls in both `top_n` and `top_n_ids` with this macro.

#### 📄 Patch (ready to apply)

```
diff --git "a/rig-integrations/rig-postgres/src/lib.rs" "b/rig-integrations/rig-postgres/src/lib.rs"
index 3436c324..c91d503c 100644
--- "a/rig-integrations/rig-postgres/src/lib.rs"
+++ "b/rig-integrations/rig-postgres/src/lib.rs"
@@ -13,6 +13,26 @@ use serde_json::Value;
 use sqlx::PgPool;
 use uuid::Uuid;
 
+macro_rules! bind_json_params {
+    ($builder:expr, $params:expr) => {{
+        $params.iter().fold($builder, |mut __b, param| match param {
+            serde_json::Value::String(s) => __b.bind(s.as_str()),
+            serde_json::Value::Number(n) => {
+                if let Some(i) = n.as_i64() {
+                    __b.bind(i)
+                } else if let Some(f) = n.as_f64() {
+                    __b.bind(f)
+                } else {
+                    __b.bind(n.to_string())
+                }
+            }
+            serde_json::Value::Bool(b) => __b.bind(*b),
+            serde_json::Value::Null => __b.bind(Option::<&str>::None as Option<&str>),
+            _ => __b.bind(param),
+        })
+    }};
+}
+
 pub struct PostgresVectorStore<Model: EmbeddingModel> {
     model: Model,
     pg_pool: PgPool,
@@ -50,7 +70,7 @@ impl Display for PgVectorDistanceFunction {
     }
 }
 
-#[derive(Clone, Default)]
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
 pub struct PgSearchFilter {
     condition: String,
     values: Vec<serde_json::Value>,
@@ -366,9 +386,10 @@ where
             .bind(embedded_query)
             .bind(req.samples() as i64);
 
-        let builder = params
-            .iter()
-            .fold(builder, |builder, param| builder.bind(param));
+        // let builder = params
+        //     .iter()
+        //     .fold(builder, |builder, param| builder.bind(param));
+        let builder = bind_json_params!(builder, params);
 
         let rows = builder
             .fetch_all(&self.pg_pool)
@@ -412,9 +433,10 @@ where
             .bind(embedded_query)
             .bind(req.samples() as i64);
 
-        let builder = params
-            .iter()
-            .fold(builder, |builder, param| builder.bind(param));
+        // let builder = params
+        //     .iter()
+        //     .fold(builder, |builder, param| builder.bind(param));
+        let builder = bind_json_params!(builder, params);
 
         let rows: Vec<SearchResultOnlyId> = builder
             .fetch_all(&self.pg_pool)

```

✅ This patch has been tested locally and correctly handles common JSON types (`string`, `number`, `boolean`, `null`) in metadata filters.

#### ✅ Impact

- **Backward compatible**: No changes to public API.
- **Safe**: Falls back to original behavior for unexpected types.
- **Improves robustness**: Enables full use of structured metadata in PostgreSQL vector searches.

FayCarsons:


joshua-mo-143 (approved):
lgtm

Issue #1234, opened 2026-01-09
feat: Add `AgentBuilder::tools` to add static tools dynamically


silentEAG:
- [x] I have looked for existing issues (including closed) about this

## Feature Request
Add a `tools` method to `AgentBuilder` and `AgentBuilderSimple` (parameter type: `Vec<Box<dyn ToolDyn>>`) to support configurable, bulk importing of tools.

### Motivation
If external code needs to integrate the rig agent and dynamically update the tool configuration, it requires `Box<dyn Trait>` for dynamic dispatch. In the current implementation, for `static_tool`, the code only supports importing tools one by one via `tool(mut self, tool: impl Tool + 'static) -> Self`, which cannot effectively meet this requirement.

### Proposal
Add `AgentBuilder::tools` and `AgentBuilderSimple::tools` methods to update an agent’s configured static tools using `Vec<Box<dyn ToolDyn>>`.

joshua-mo-143:
Closing as the relevant PR to solve this issue has been merged (#1236)

PR #1236, opened 2026-01-09
Add `AgentBuilder::tools` for adding static tools dynamically

silentEAG:
Feat #1234


silentEAG:
> `cargo fmt` then lgtm

done

Issue #1075, opened 2025-11-23
feat: List models


Sytten:
- [x] I have looked for existing issues (including closed) about this

## Feature Request
I would like to dynamically list models provided by a provider. This could be a new Client trait, almost all providers have `/models` endpoint we can call to get that list.

### Motivation
When working with customer providers like LiteLLM, the list of models is not static. It depends on what the user configured.

### Proposal
```rust
pub trait ListModelsClient {
    /// Verify the configuration.
    fn models(&self) -> impl Future<Output = Result<Vec<Model>, VerifyError>> + WasmCompatSend;
}
```

joshua-mo-143:
Looks good to me, would be great to have this in Rig!

(apologies for random tag labelling, wasn't sure what we should use but I guess just commenting works for now since there aren't too many issues yet)

Sytten:
Great I will try to find the time to code this in the coming weeks!

eibrahim95:
Hey, This looks straight forwarded enough, I would like to take a stab at this as my first contribution here, if you are currently not working on it @Sytten 

Thanks

Sytten:
I am not

eibrahim95:
Okay @Sytten @joshua-mo-143 I've linked a PR (interestingly, the id of the pr is 1234 :D), 
https://github.com/0xPlaygrounds/rig/pull/1243/files This adds the capability and the related traits, and the struct model.

If all goes well, I'll start creating PRs for different providers.

Thanks

gold-silver-copper:
Closed by #1243

PR #1243, opened 2026-01-12
feat: Add model listing capability

eibrahim95:
Putting the ground work for:  #1075

Adds ModelListingClient and ModelLister traits for discovering available provider models. Introduces Model, ModelList, and ModelListingError types to represent responses. Updates provider capabilities to include model listing.

If all goes well with this PR, I'll start creating a PR for each provider that supports model listing to implement the model listing capability.

### DIFF
--- a/rig/rig-core/src/model/listing.rs
+++ b/rig/rig-core/src/model/listing.rs
@@ -45,7 +45,6 @@ use std::fmt;
 /// };
 /// ```
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
-#[serde(rename_all = "camelCase")]
 pub struct Model {
     /// The unique identifier for the model (required)
     pub id: String,
@@ -167,7 +166,6 @@ impl fmt::Display for Model {
 /// # Fields
 ///
 /// - `data`: The complete list of available models
-/// - `total`: Total number of models
 ///
 /// # Example
 ///
@@ -185,13 +183,9 @@ impl fmt::Display for Model {
 /// }
 /// ```
 #[derive(Debug, Clone, Serialize, Deserialize)]
-#[serde(rename_all = "camelCase")]
 pub struct ModelList {
     /// The complete list of available models
     pub data: Vec<Model>,
-
-    /// Total number of models available
-    pub total: usize,
 }
 
 impl ModelList {
@@ -211,11 +205,9 @@ impl ModelList {
     ///     Model::from_id("gpt-3.5-turbo"),
     /// ]);
     /// assert_eq!(list.len(), 2);
-    /// assert_eq!(list.total, 2);
     /// ```
     pub fn new(data: Vec<Model>) -> Self {
-        let total = data.len();
-        Self { data, total }
+        Self { data }
     }
 
     /// Returns true if the list is empty.
@@ -295,16 +287,6 @@ impl<'a> IntoIterator for &'a ModelList {
 ///
 /// This enum represents the various error conditions that may arise when
 /// attempting to retrieve the list of available models from an LLM provider.
-///
-/// # Variants
-///
-/// - `ApiError`: The provider returned an error response with a status code
-/// - `RequestError`: Failed to send the request to the provider
-/// - `ParseError`: Failed to parse the provider's response
-/// - `AuthError`: Authentication failed (invalid API key, etc.)
-/// - `RateLimitError`: Rate limit was exceeded
-/// - `ServiceUnavailable`: The provider service is temporarily unavailable
-/// - `UnknownError`: An unexpected error occurred
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub enum ModelListingError {
     /// The provider returned an error response with a status code
@@ -465,7 +447,6 @@ mod tests {
     fn test_model_list_new() {
         let list = ModelList::new(vec![Model::from_id("gpt-4")]);
         assert_eq!(list.len(), 1);
-        assert_eq!(list.total, 1);
     }
 
     #[test]
@@ -544,16 +525,13 @@ mod tests {
     fn test_model_list_serde() {
         let list = ModelList {
             data: vec![Model::from_id("gpt-4")],
-            total: 100,
         };
 
         let json = serde_json::to_string(&list).unwrap();
         assert!(json.contains("gpt-4"));
-        assert!(json.contains("total"));
 
         let deserialized: ModelList = serde_json::from_str(&json).unwrap();
         assert_eq!(deserialized.len(), 1);
-        assert_eq!(deserialized.total, 100);
     }
 
     #[test]


rig/rig-core/src/client/model_listing.rs
```diff
@@ -0,0 +1,173 @@
...
+    /// This implementation should handle fetching all pages if the provider
+    /// supports pagination, returning complete results in a single call.
+    ///
+    /// # Returns
+    ///
+    /// A `ModelList` containing all available models.
+    fn list_all(
+        &self,
+    ) -> impl std::future::Future<Output = Result<ModelList, ModelListingError>> + WasmCompatSend;
+}
```
joshua-mo-143, line 128: Are there any parameters we're missing here? What about if users just want to get results from the first page (for instance)?

rig/rig-core/src/model/listing.rs
```diff
@@ -0,0 +1,578 @@
...
+    ///     Model::from_id("gpt-3.5-turbo"),
+    /// ]);
+    ///
+    /// for model in list.iter() {
+    ///     println!("Model: {}", model.display_name());
+    /// }
+    /// ```
+    pub fn iter(&self) -> std::slice::Iter<'_, Model> {
+        self.data.iter()
+    }
```
joshua-mo-143, line 265: Why not just implement `std::iter::Iterator`? (edit: ref: https://doc.rust-lang.org/std/iter/trait.Iterator.html)

rig/rig-core/src/model/listing.rs
```diff
@@ -0,0 +1,578 @@
...
+    ///
+    /// let list = ModelList::new(vec![
+    ///     Model::from_id("gpt-4"),
+    ///     Model::from_id("gpt-3.5-turbo"),
+    /// ]);
+    /// assert_eq!(list.len(), 2);
+    /// ```
+    pub fn len(&self) -> usize {
+        self.data.len()
+    }
```
joshua-mo-143, line 245: General feedback: you store a `total` variable, then use `self.data.len()` to get the number of models on a page. This makes no sense and requires cleanup

rig/rig-core/src/model/listing.rs
```diff
@@ -0,0 +1,578 @@
...
+///
+/// # Variants
+///
+/// - `ApiError`: The provider returned an error response with a status code
+/// - `RequestError`: Failed to send the request to the provider
+/// - `ParseError`: Failed to parse the provider's response
+/// - `AuthError`: Authentication failed (invalid API key, etc.)
+/// - `RateLimitError`: Rate limit was exceeded
+/// - `ServiceUnavailable`: The provider service is temporarily unavailable
+/// - `UnknownError`: An unexpected error occurred
```
joshua-mo-143: We can probably remove this part. The error names are pretty self explanatory

rig/rig-core/src/client/model_listing.rs
eibrahim95: Actually, the documentation is wrong, I don't want to do pagination at all for simplicity, usually model list is not that big, when i first implemented this there was pagination, and then i removed it, and didn't clean up documentation well.

Do you think we need pagination here?

rig/rig-core/src/model/listing.rs
eibrahim95: Yeah, because of the early implementation I did that had pagination.
Nice Catch!
I think I'll stick with no pagination, and remove that total variable

rig/rig-core/src/client/model_listing.rs
eibrahim95: oh! I just rememberd, the documentation is saying that the individual provider implementation, should iterate on all pages, and returning the union, with that we are making our trait paginationless.
Meaning we are not exposing pagination of providers to the final users of rig.

rig/rig-core/src/model/listing.rs
eibrahim95: Sorry I still don't have a lot of experience with Rust,
I did some research and this would require having some state that hold the current item, so i can implement the next() method, don't you think it is better to make use of already implemented method of Vec?

Or is there anything i'm missing?

rig/rig-core/src/model/listing.rs
eibrahim95: Done

rig/rig-core/src/model/listing.rs
eibrahim95: Done

rig/rig-core/src/model/listing.rs
joshua-mo-143: It's more idiomatic to implement Iterator because there might be other methods/types outside of the Rig crate that require types that implement Iterator.

joshua-mo-143: Just coming back on this: it looks like using this version should be fine. We can probably leave it alone for now unless there is an absolute pressing need to impl `Iterator`.

rig/rig-core/src/client/model_listing.rs
joshua-mo-143: > oh! I just rememberd, the documentation is saying that the individual provider implementation, should iterate on all pages, and returning the union, with that we are making our trait paginationless. Meaning we are not exposing pagination of providers to the final users of rig.

Sorry can you just remind me quickly which documentation this is referring to? I am a bit lost on context here

eibrahim95: my documentation:
```
    /// List all available models from the provider.
    ///
    /// This implementation should handle fetching all pages if the provider
    /// supports pagination, returning complete results in a single call.
    ///
    /// # Returns
    ///
    /// A `ModelList` containing all available models.
    ```

Comment (joshua-mo-143) [reply rig/rig-core/src/client/model_listing.rs:128]: Oh, sorry - yes. This should be fine, I think.

**Comment by eibrahim95** [comment]
Hey @joshua-mo-143 Is there a plan to merge this soon?

File: rig/rig-core/src/model/listing.rs [hunk 5]
```diff
@@ -0,0 +1,556 @@
...
+    pub name: Option<String>,
+
+    /// A detailed description of the model's capabilities
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub description: Option<String>,
+
+    /// The type of model (e.g., "chat", "completion", "embedding")
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(rename = "type")]
+    pub r#type: Option<String>,
```
ehiggs, line 63: TIL about raw identifiers. And they're used a lot in the wild it seems.

https://github.com/search?q=%22r%23type%22+language%3Arust&type=code

rig/rig-core/src/model/listing.rs
```diff
@@ -0,0 +1,556 @@
...
+    /// The organization or entity that owns the model
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub owned_by: Option<String>,
+
+    /// The maximum context window size for the model
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub context_length: Option<u32>,
+}
+
+impl Model {
```
ehiggs, line 78: Maybe this should be `ModelListEntry`. Elsewhere in the code the `Model` is a `String`. e.g. in the ollama provider:

```
#[derive(Debug, Serialize, Deserialize)]
pub struct CompletionResponse {
    pub model: String,
```

```
#[derive(Debug, Serialize, Deserialize)]
pub struct EmbeddingResponse {
    pub model: String,
```

Everything is namespaced; but when people hit alt-enter or alt-. (depending on their IDE) they will be given options to import Model and if there are fewer options then it's easier for people to fall into the pit of success.

rig/rig-core/src/model/listing.rs
```diff
@@ -0,0 +1,556 @@
...
+///     Model::from_id("gpt-3.5-turbo"),
+/// ]);
+///
+/// println!("Found {} models", list.len());
+/// for model in list.iter() {
+///     println!("- {}", model.display_name());
+/// }
+/// ```
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ModelList {
```
ehiggs, line 186: The methods below don't provide any value above `type ModelList = Vec<Model>` Then you don't need to implement `IntoIterator` and all these other methods.

rig/rig-core/src/model/listing.rs
ehiggs: @joshua-mo-143 I would say don't even make this a struct. I mention above that there's no value here above making a new type defintion to name the Vec<Model> as ModelList.

rig/rig-core/src/model/listing.rs
ehiggs: @eibrahim95 a question about the usage here. The Model here is returned from the provider's client api and that is the only use case, right? If so, is there a time when these fields are really `None`? I get that it's convenient to test using `Model::from_id` as a one liner to get the tests, but you can also make a function to make `Model` objects in the test submodule and populate the fields with actual data.

joshua-mo-143:
@eibrahim95 I think some valid points have been raised here by @ehiggs.

Looks OK to me, just needs maybe some amendments according to the comments made and lgtm

PR #1252, opened 2026-01-14
feat: add custom vector store backend example

FayCarsons:


rig/rig-core/Cargo.toml
```diff
@@ -51,7 +51,17 @@ pin-project-lite = "0.2.16"
...
+  "json",
+  "multipart",
+  "charset",
+  "http2",
+] }
+redis = { version = "1.0.2", optional = true, features = [
+  "tokio-comp",
+  "aio",
+  "vector-sets",
+] }
```
joshua-mo-143: This should be a dev dependency

joshua-mo-143 (changes requested):
General comments:
- I think "vector_store" is a bit too ambiguous as an example name - maybe "vector_store_custom"?

That's pretty much about it besides comments

FayCarsons:
Will do !

PR #1253, opened 2026-01-14
feat: add default max depth to agents

0xMochan:
Adds a `.default_max_depth` to `Agent` and it's various builders. `PromptRequest` will look at this (if defined) and use it by default. Otherwise, works the same as before (starts at 0). This number can still be overridden by `.multi_turn`.

### DIFF
--- a/rig/rig-core/src/agent/builder.rs
+++ b/rig/rig-core/src/agent/builder.rs
@@ -65,6 +65,8 @@ where
     tool_server_handle: Option<ToolServerHandle>,
     /// Whether or not the underlying LLM should be forced to use a tool before providing a response.
     tool_choice: Option<ToolChoice>,
+    /// Default maximum depth for recursive agent calls
+    default_max_depth: Option<usize>,
 }
 
 impl<M> AgentBuilder<M>
@@ -84,6 +86,7 @@ where
             dynamic_context: vec![],
             tool_server_handle: None,
             tool_choice: None,
+            default_max_depth: None,
         }
     }
 
@@ -151,6 +154,7 @@ where
             temperature: self.temperature,
             tools,
             tool_choice: self.tool_choice,
+            default_max_depth: self.default_max_depth,
         }
     }
 
@@ -174,6 +178,7 @@ where
             temperature: self.temperature,
             tools,
             tool_choice: self.tool_choice,
+            default_max_depth: self.default_max_depth,
         }
     }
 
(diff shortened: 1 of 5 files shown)


rig/rig-core/src/agent/prompt_request/mod.rs
```diff
@@ -73,7 +73,7 @@ where
         Self {
             prompt: prompt.into(),
             chat_history: None,
-            max_depth: 0,
+            max_depth: agent.default_max_depth.unwrap_or(0),
```
joshua-mo-143: nit: Use `unwrap_or_default()` here (the default for usize is already 0)

rig/rig-core/src/agent/prompt_request/streaming.rs
```diff
@@ -125,7 +125,7 @@ where
         Self {
             prompt: prompt.into(),
             chat_history: None,
-            max_depth: 0,
+            max_depth: agent.default_max_depth.unwrap_or(0),
```
joshua-mo-143: nit: Use `unwrap_or_default()` here (the default for usize is already 0)

joshua-mo-143 (approved):
Looks OK to me.

I do think we really need to get on top of controlling how many examples we have... we're getting close to almost 100 example files. That's too many

0xMochan:
> Looks OK to me.
> 
> I do think we really need to get on top of controlling how many examples we have... we're getting close to almost 100 example files. That's too many

Yea, tbh i was thinking of just editting an existing one. also some of the examples use models that don't work anymore lol

0xMochan:
btw, is this naming ok?

joshua-mo-143:
> btw, is this naming ok?

seems fine to me

rig/rig-core/src/agent/builder.rs
```diff
@@ -266,6 +273,11 @@ where
         self
     }
 
+    pub fn default_max_depth(mut self, default_max_depth: usize) -> Self {
+        self.default_max_depth = Some(default_max_depth);
+        self
+    }
+
```
joshua-mo-143, line 281: sorry completely forgot to add - can you add a docstring to clarify that "max depth" is basically just the maximum number of turns or steps that an agent will use for multi-step tool calling?

rig/rig-core/src/agent/builder.rs
```diff
@@ -506,6 +523,11 @@ where
...
```
joshua-mo-143, line 532: same as other review comment

joshua-mo-143 (changes requested):
sorry missed something in the previous review, just need to add some docstring to clarify what "max depth" is as I don't think it's very obvious to users reading docs

joshua-mo-143 (approved):

PR #1282, opened 2026-01-20
fix(agent): fix CancelSignal cancellation and reason sharing bugs

### LINKED ISSUE #1281: bug: CancelSignal::cancel_with_reason() doesn't cancel and reason not shared on clone

## Bug Report

`CancelSignal` has two bugs that prevent the Hook mechanism from working correctly:

1. **`cancel_with_reason()` doesn't set the cancelled flag** - After calling `cancel_with_reason()`, `is_cancelled()` still returns `false`
2. **`Clone` doesn't share reason** - `OnceLock<String>` is cloned as independent instances, so reason set on a cloned signal is invisible to the original

## Reproduction

```rust
// Bug 1: cancel_with_reason() doesn't cancel
let sig = CancelSignal::new();
sig.cancel_with_reason("frontend_tool");

assert!(!sig.is_cancelled());  // ❌ Still false!
assert_eq!(sig.cancel_reason(), Some("frontend_tool"));  // Reason is set

// Bug 2: Clone doesn't share reason
let sig = CancelSignal::new();
let cloned = sig.clone();

cloned.cancel_with_reason("frontend_tool");
cloned.cancel();

assert!(sig.is_cancelled());  // ✅ sig is shared via Arc
assert_eq!(cloned.cancel_reason(), Some("frontend_tool"));  // ✅ Clone has reason
assert_eq!(sig.cancel_reason(), None);  // ❌ Original has no reason!
```

## Expected behavior

1. `cancel_with_reason()` should set both the reason AND the cancelled flag
2. Cloned `CancelSignal` should share the same reason storage, so setting reason on any clone is visible to all

## Screenshots

N/A

## Additional context

This bug was discovered while implementing **Frontend Tools** feature. The workflow:

1. `StreamingPromptHook::on_tool_call` intercepts frontend tool calls
2. Hook calls `cancel_sig.cancel_with_reason("frontend_tool")` to cancel the loop
3. Agent loop checks `cancel_sig.is_cancelled()` and reads `cancel_reason()`

**Current result**: Loop doesn't cancel (Bug 1), or reason is always `None` (Bug 2) because hook receives a cloned signal.

**Suggested fix**:

```diff
 pub struct CancelSignal {
     sig: Arc<AtomicBool>,
-    reason: OnceLock<String>,
+    reason: Arc<OnceLock<String>>,
 }

 pub fn cancel_with_reason(&self, reason: &str) {
     let _ = self.reason.set(reason.to_string());
+    self.cancel();
 }
```

yexiyue:
## Summary

Fix two bugs in `CancelSignal` that prevent proper cancellation handling in prompt hooks.

Fixes #1281 

## Changes

1. **`cancel_with_reason()` now calls `cancel()`** - Previously it only set the reason without setting the cancelled flag, so `is_cancelled()` still returned `false`.

2. **Wrap `reason` in `Arc`** - Change from `OnceLock<String>` to `Arc<OnceLock<String>>` so cloned signals share the same reason instance.

```diff
 pub struct CancelSignal {
     sig: Arc<AtomicBool>,
-    reason: OnceLock<String>,
+    reason: Arc<OnceLock<String>>,
 }

 pub fn cancel_with_reason(&self, reason: &str) {
     let _ = self.reason.set(reason.to_string());
+    self.cancel();
 }
```

## Test Plan

```rust
#[test]
fn test_cancel_with_reason_sets_flag() {
    let sig = CancelSignal::new();
    sig.cancel_with_reason("test");
    assert!(sig.is_cancelled());
}

#[test]
fn test_clone_shares_reason() {
    let sig = CancelSignal::new();
    let cloned = sig.clone();
    cloned.cancel_with_reason("test");
    assert_eq!(sig.cancel_reason(), Some("test"));
}
```

### DIFF
--- a/rig/rig-core/src/agent/prompt_request/mod.rs
+++ b/rig/rig-core/src/agent/prompt_request/mod.rs
@@ -160,14 +160,14 @@ where
 /// You can additionally add a reason for early termination with `CancelSignal::cancel_with_reason()`.
 pub struct CancelSignal {
     sig: Arc<AtomicBool>,
-    reason: OnceLock<String>,
+    reason: Arc<OnceLock<String>>,
 }
 
 impl CancelSignal {
     fn new() -> Self {
         Self {
             sig: Arc::new(AtomicBool::new(false)),
-            reason: OnceLock::new(),
+            reason: Arc::new(OnceLock::new()),
         }
     }
 
@@ -179,6 +179,7 @@ impl CancelSignal {
         // SAFETY: This can only be set once. We immediately return once the prompt hook is finished if the internal AtomicBool is set to true
         // It is technically on the user to return early when using this in a prompt hook, but this is relatively obvious
         let _ = self.reason.set(reason.to_string());
+        self.cancel();
     }
 
     fn is_cancelled(&self) -> bool {


joshua-mo-143:
hi @yexiyue, thanks for submitting this! looks like I somehow missed it :sweat_smile: good thing I delayed release

Issue #1260, opened 2026-01-16
feat: Allow `on_tool_call` hook to reject tool execution


Phoenix500526:
- [x] I have looked for existing issues (including closed) about this

## Feature Request

Add the ability for `on_tool_call` hooks to reject tool execution and return rejection messages to the LLM, enabling use cases like permission control and user approval workflows.              

### Motivation

Currently, the `on_tool_call` hook in both `PromptHook` and `StreamingPromptHook` traits returns `()`, making it purely observational. Once a tool is called by the LLM, it will always be executed. This limitation prevents several important use cases:                                               

1. **Permission Control**: Applications cannot implement approval workflows where certain tools require user permission before execution (e.g., file deletion, system modifications, network requests).                                                                                                                                         
2. **Conditional Tool Availability**: Hooks cannot dynamically enable/disable tools based on runtime state (e.g., rate limiting, feature flags, user roles).              
3.  **Safety Guardrails**: There's no way to programmatically block dangerous operations based on context or validation rules.                                 
4.  **Interactive Agents**: Cannot build agents that ask users for confirmation before executing sensitive operations.                                                       
                                                                                                                                          
Reference: https://github.com/0xPlaygrounds/rig/issues/1191#issuecomment-3696323238   

### Proposal

Change the return type of `on_tool_call` from `()` to `Result<(), String>`:  
                                                                                          
  ```rust                                                                               
  // Before                                                                             
  async fn on_tool_call(                                                                
      &self,                                                                            
      tool_name: &str,                                                                  
      tool_call_id: Option<String>,                                                     
      args: &str,                                                                       
      cancel_sig: CancelSignal,                                                         
  ) -> impl Future<Output = ()> + WasmCompatSend {                                      
      async {}                                                                          
  }                                                                                     
                                                                                        
  // After                                                                              
  async fn on_tool_call(                                                                
      &self,                                                                            
      tool_name: &str,                                                                  
      tool_call_id: Option<String>,                                                     
      args: &str,                                                                       
      cancel_sig: CancelSignal,                                                         
  ) -> impl Future<Output = Result<(), String>> + WasmCompatSend {                      
      async { Ok(()) }                                                                  
  }  
```                                                                
                                                                                        
Behavior:                                                                             
  - Ok(()) - Proceed with tool execution (current behavior)                             
  - Err(message) - Reject tool execution; the error message is returned to the LLM as the tool result            
                                                                                        
 
### Alternatives
[...]

PR #1284, opened 2026-01-20
feat(agent): allow on_tool_call hook to reject tool execution

Phoenix500526:
Change on_tool_call return type from () to ToolCallHookAction. Returning ToolCallHookAction::Skip{reason} will skip tool execution and return the rejection reason to the LLM as the tool result.

Closes: #1260


rig/rig-core/src/agent/prompt_request/mod.rs
```diff
@@ -511,16 +524,35 @@ where
...
+                                    )
+                                    .await;
+
                                 if cancel_sig1.is_cancelled() {
                                     return Err(ToolSetError::Interrupted);
                                 }
+
+                                if let ToolCallHookAction::Skip { reason } = action {
+                                    // Tool execution rejected, return rejection message as tool result
+                                    tracing::info!("Tool {tool_name} rejected: {reason}");
```
joshua-mo-143: nit: `tool_name` and `reason` should be structured fields within the trace.

See the following:
```rust
tracing::info!(
    tool_name = tool_name,
    reason = reason,
    "Tool call rejected"
);
```

Phoenix500526: Done

Issue #1161, opened 2025-12-12
feat: Pass tool_call_id to PromptHook::on_tool_call and on_tool_result


Phoenix500526:
- [x] I have looked for existing issues (including closed) about this

## Feature Request

Add tool_call_id parameter to PromptHook::on_tool_call and PromptHook::on_tool_result callbacks.    

### Motivation

I'm building an AI coding agent that uses rig-core's Agent abstraction with PromptHook to send real-time tool execution notifications to an IDE. The communication follows the [ACP](https://agentclientprotocol.com), which requires a tool_call_id to correlate tool start and completion events: 

```text
  // ACP session/update notification                                                                                                                                          
  interface ToolCall {                                                                                                                                                        
    tool_call_id: string;  // Required to match start and complete                                                                                                            
    tool_name: string;                                                                                                                                                        
    arguments: object;                                                                                                                                                        
  }                                                                                                                                                                           
                                                                                                                                                                              
  interface ToolResult {                                                                                                                                                      
    tool_call_id: string;  // Must match the corresponding ToolCall                                                                                                           
    result: string;                                                                                                                                                           
    success: boolean;                                                                                                                                                         
  }
```                                                                                                                                                                    
Currently, PromptHook trait methods don't expose the tool_call_id:                                                                                                          
```rust
  fn on_tool_call(&self, tool_name: &str, args: &str, cancel_sig: CancelSignal);  
  fn on_tool_result(&self, tool_name: &str, args: &str, result: &str, cancel_sig: CancelSignal); 
```
The tool_call_id (from ToolCall::id) is available internally in the agent loop but not passed to hooks, making it impossible to properly correlate tool events for external notification systems.      

### Proposal
Add tool_call_id: &str as the first parameter to both methods:                                                                                                              
```rust                                                                                                                                                           
  fn on_tool_call(&self, tool_call_id: &str, tool_name: &str, args: &str, cancel_sig: CancelSignal);                                                                          
  fn on_tool_result(&self, tool_call_id: &str, tool_name: &str, args: &str, result: &str, cancel_sig: CancelSignal);                                                          
```                                                                                                                                                                  

The ID is already available in the agent loop from ToolCall::id. The change would simply pass it through to the hook callbacks. 

This is a breaking change to the `PromptHook` trait. Existing implementations will need to update their method signatures.                                          

I'd be happy to submit a PR for this if you're interested.                                                                                                                  
                                                                                                                           

### Alternatives

[...]

PR #1285, opened 2026-01-20
fix(message): initialize ToolCall::call_id as Some(id)

Phoenix500526:
In issue #1161, we agreed to pass `tool_call_id` into `PromptHook` and `StreamingPromptHook` (`on_tool_call` / `on_tool_result`). However, `ToolCall` / `RawStreamToolCall` were always constructed with `call_id` = `None`, which meant `tool_call_id` was always None in those callbacks.

It prevented external clients (e.g., ACP) from matching tool execution results with their corresponding tool calls. This change initializes ToolCall::call_id as Some(id) so the hook callbacks always receive a usabletool_call_id.

### DIFF
--- a/rig/rig-core/Cargo.toml
+++ b/rig/rig-core/Cargo.toml
@@ -57,7 +57,7 @@ reqwest-middleware = { version = "0.4.2", optional = true, features = [
   "charset",
   "http2",
 ] }
-uuid = { workspace = true, features = ["v4"] }
+uuid = { workspace = true, features = ["v4", "js"] }
 
 [dev-dependencies]
 anyhow = { workspace = true }


joshua-mo-143:
`ToolCall::with_call_id()` exists for exactly this purpose.

edit: Where exactly are you seeing that tool calls are always constructed with call_id = None? We should fix it at those callsites rather than trying to do a lazy workaround method like this

Phoenix500526:
> `ToolCall::with_call_id()` exists for exactly this purpose.
> 
> edit: Where exactly are you seeing that tool calls are always constructed with call_id = None? We should fix it at those callsites rather than trying to do a lazy workaround method like this

I’ve already used with_call_id to specify the tool call ID for some models, such as OpenAI, Anthropic, and Gemini. For Gemini, since it doesn’t provide a unique tool call ID, I generate one using a UUID. Please take a look when you have time.

joshua-mo-143:
Superceded by #1311 - I think having a separate internal call ID is better than trying to twist the current API into being  Some

PR #1289, opened 2026-01-21
fix(providers): unify gemini and ollama EmbeddingModel exports

martsokha:
- Re-export EmbeddingModel, CompletionModel, and embedding constants (EMBEDDING_001, EMBEDDING_004) at gemini module root for consistency with other providers like cohere or mistral
- Add default type parameter `T = reqwest::Client` to ollama's EmbeddingModel for consistency with other providers
- Update gemini doc example to use the shorter path

### DIFF
--- a/rig/rig-core/src/providers/gemini/mod.rs
+++ b/rig/rig-core/src/providers/gemini/mod.rs
@@ -6,7 +6,7 @@
 //!
 //! let client = gemini::Client::new("YOUR_API_KEY");
 //!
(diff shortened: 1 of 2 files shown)


joshua-mo-143:
Hi, can you please split this into two PRs? One PR should be equal to one feature - adding them all like this makes it hard to revert singular features if anything goes wrong

martsokha:
@joshua-mo-143 #1292 #1293

PR #1299, opened 2026-01-22
feat(providers): expose cached_input_tokens in Usage struct

### LINKED ISSUE #1297: feat: Expose cached prompt tokens in usage

## Feature Request
Rig currently surfaces top-level token usage (`input_tokens`, `output_tokens`, `total_tokens`) but does not expose cached prompt token counts returned by some OpenAI-compatible providers (notably Groq) in:

- `usage.prompt_tokens_details.cached_tokens` (Chat Completions / OpenAI-compatible schema)

This field is needed to meter/bill cached input tokens correctly when prompt caching discounts apply.

### Motivation
Groq’s prompt caching reports cache hits via `usage.prompt_tokens_details.cached_tokens` and applies a discount to cached input tokens. OpenAI’s prompt caching similarly reports cache hits via `cached_tokens`.

Without this field, downstream apps cannot accurately calculate discounted cached-input costs through Rig’s abstractions, and must bypass Rig to call provider HTTP APIs directly (or implement provider-specific parsing).

### Proposal
1) Extend Rig’s token usage surface area to include cached input tokens (optional / 0 when absent).  
   *Example shape*
   - `cached_input_tokens: Option<u64>` on `rig::completion::request::Usage`, **or**
   - a nested `UsageDetails { cached_input_tokens: u64, ... }` behind an `Option`, **or**
   - another non-breaking approach that still exposes cached tokens via the public API.

2) Parse and surface cached token details for OpenAI-compatible Chat Completions responses:
   - Parse `usage.prompt_tokens_details.cached_tokens` and expose it through the generic usage type/trait.
   - This should apply to Groq (OpenAI compatibility) and OpenAI-compatible endpoints that include this field.

3) Propagate cached tokens from OpenAI Responses API usage:
   - Rig already deserializes `ResponsesUsage.input_tokens_details.cached_tokens`, but `GetTokenUsage` for Responses streaming currently drops these details when converting into `crate::completion::Usage`.
   - Please map this cached token value into the same `cached_input_tokens` field so both schemas are supported.

#### Example JSON (Groq / OpenAI-compatible Chat Completions)
```json
{
  "usage": {
    "prompt_tokens": 4641,
    "completion_tokens": 1817,
    "total_tokens": 6458,
    "prompt_tokens_details": { "cached_tokens": 4608 }
  }
}
```

#### Expected outcome
Downstream code using Rig’s standard token usage access (`GetTokenUsage`/`CompletionResponse` usage) can reliably read cached input token counts when providers supply them, enabling correct discounted pricing and usage reporting.

legietha:
Add cached_input_tokens field to the core Usage struct to expose prompt caching information from OpenAI-compatible providers like Groq.

Changes:
- Add cached_input_tokens: u64 field to completion::Usage
- Add PromptTokensDetails struct to OpenAI completion module
- Update all providers to include cached_input_tokens in GetTokenUsage
- Map cache_read_input_tokens for Anthropic provider
- Extract cached_tokens from prompt_tokens_details for Groq/DeepSeek

Fixes #1297

### DIFF
--- a/rig-integrations/rig-bedrock/src/streaming.rs
+++ b/rig-integrations/rig-bedrock/src/streaming.rs
@@ -28,6 +28,7 @@ impl GetTokenUsage for BedrockStreamingResponse {
             input_tokens: u.input_tokens as u64,
             output_tokens: u.output_tokens as u64,
             total_tokens: u.total_tokens as u64,
+            cached_input_tokens: 0, // unsupported at time of adding this
         })
     }
 }
--- a/rig-integrations/rig-bedrock/src/types/assistant_content.rs
+++ b/rig-integrations/rig-bedrock/src/types/assistant_content.rs
@@ -49,6 +49,8 @@ impl TryFrom<AwsConverseOutput> for completion::CompletionResponse<AwsConverseOu
                 input_tokens: usage.input_tokens as u64,
                 output_tokens: usage.output_tokens as u64,
                 total_tokens: usage.total_tokens as u64,
+                cached_input_tokens: usage.cache_read_input_tokens.unwrap_or_default() as u64
+                    + usage.cache_write_input_tokens.unwrap_or_default() as u64,
             })
             .unwrap_or_default();
 
--- a/rig-integrations/rig-eternalai/src/providers/eternalai.rs
+++ b/rig-integrations/rig-eternalai/src/providers/eternalai.rs
@@ -508,6 +508,7 @@ impl TryFrom<CompletionResponse> for completion::CompletionResponse<CompletionRe
                 input_tokens: usage.prompt_tokens as u64,
                 output_tokens: (usage.total_tokens - usage.prompt_tokens) as u64,
                 total_tokens: usage.total_tokens as u64,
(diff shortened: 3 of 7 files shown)


rig/rig-core/src/completion/request.rs
```diff
@@ -303,6 +303,8 @@ pub struct Usage {
     pub output_tokens: u64,
     /// We store this separately as some providers may only report one number
     pub total_tokens: u64,
+    /// The number of cached input tokens (from prompt caching). 0 if not reported by provider.
+    pub cached_input_tokens: u64,
```
joshua-mo-143, line 307: nit: If they're not reported, shouldn't this theoretically be None?

joshua-mo-143: Resolving as this isn't really worth debating. Technically if it's not reported, we can assume that they don't support this.

joshua-mo-143:
Manually fixing and merging because we need this PR internally and are pretty close to releasing a new version for Rig (and no response from user).

Thanks for the PR!

PR #1302, opened 2026-01-22
feat(1159): impl VectorStoreIndexDyn for scyllaDB integration

FayCarsons:


rig-integrations/rig-scylladb/src/lib.rs
```diff
@@ -41,8 +42,48 @@ pub struct ScyllaDbVectorStore<M: EmbeddingModel> {
...
+            let pairs = map
+                .into_iter()
+                .map(|(k, v)| Ok((CqlValue::Text(k), cql_value_from_json(v)?)))
+                .collect::<Result<Vec<_>, FilterError>>()?;
+            Ok(CqlValue::Map(pairs))
+        }
+        Value::Null => Ok(CqlValue::Empty),
+    }
+}
+
```
joshua-mo-143, line 86: please use TryFrom<serde_json::Value> for CqlValue

FayCarsons: This would be an orphan instance since neither type is defined in rig. We could wrap `CqlValue` in a newtype, but since this isn't exposed in the API afaict that may just add friction on our end. Up to you!

joshua-mo-143: Oh, yeah. Duh - forgot CqlValue is owned by scylladb 🤦

joshua-mo-143: Oh, yeah. Duh - forgot CqlValue is owned by scylladb 🤦

PR #1303, opened 2026-01-22
fix(providers): fixed azure openai embedding dimension

austinsimpsond41:
Hi All,

I hit a small issue with using the Azure OpenAI provider. I'm creating an embedding model with `emedding_model_with_ndims`, but the ndims property isn't used in the client. This PR adds the embedding model dimension to the request body.

I've also added a test that verifies the fix.

Please let me know if you need anything else from me!

Thank you!

### DIFF
--- a/rig/rig-core/src/providers/azure.rs
+++ b/rig/rig-core/src/providers/azure.rs
@@ -449,10 +449,15 @@ where
     ) -> Result<Vec<embeddings::Embedding>, EmbeddingError> {
         let documents = documents.into_iter().collect::<Vec<_>>();
 
-        let body = serde_json::to_vec(&json!({
+        let mut body = json!({
             "input": documents,
-            "dimensions": self.ndims,
-        }))?;
+        });
+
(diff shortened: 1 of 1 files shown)


rig/rig-core/src/providers/azure.rs
```diff
@@ -451,6 +451,7 @@ where
 
         let body = serde_json::to_vec(&json!({
             "input": documents,
+            "dimensions": self.ndims,
```
joshua-mo-143: It says [here](https://learn.microsoft.com/en-us/azure/ai-foundry/openai/reference?view=foundry-classic#request-body-1) the dimensions property is only supported on `text-embedding-3` and later models, you may want to add this conditionally since `self.ndims` is currently guaranteed to always exist

austinsimpsond41: Hi Joshua,

My pleasure. I checked to see how the openai provider works. 

It checks to see if the model isn't TEXT_EMBEDDING_ADA_002 and adds dimensions as a property if it's not. I'll make my PR match this behavior.

PR #1311, opened 2026-01-26
feat(tools): adds internal tool call id (breaking)

mylesberueda:
### Why
Gemini doesn't seem to generate tool call ids, so it's hard to track parallel/duplicate tool calls. This PR adds a provider-agnostic way to track tool events across different timings via generated internal id. 

### Details

Implemented to give rig a similar api to [langchain](https://github.com/langchain-ai/langchain/blob/c930062f69bbf72d0147db2e2db1940777966ffe/libs/core/langchain_core/callbacks/base.py#L681) by generating a [unique id on tool start](https://github.com/langchain-ai/langchain/blob/c930062f69bbf72d0147db2e2db1940777966ffe/libs/core/langchain_core/callbacks/manager.py#L1490)

This is a breaking change; consumers will need to update:
- Pattern matching on `ToolCallDelta` variants (add `internal_call_id`)
- Implementations of `on_tool_call_delta` (now has `internal_call_id`)

The original provider-based tool call id is preserved for round-trips (i.e. OpenAI, Anthropic).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10184,6 +10184,7 @@ dependencies = [
  "fastrand",
  "futures",
  "futures-timer",
+ "getrandom 0.2.16",
  "glob",
  "http 1.3.1",
  "hyper-util",
--- a/rig/rig-core/Cargo.toml
+++ b/rig/rig-core/Cargo.toml
@@ -47,6 +47,7 @@ tracing-futures = { version = "0.2.5", features = ["futures-03"] }
 serenity = { version = "0.12.4", optional = true }
 fastrand = "2.3.0"
 nanoid = { workspace = true }
+getrandom = { version = "0.2", optional = true }
 eventsource-stream = "0.2.3"
 pin-project-lite = "0.2.16"
 futures-timer = "3.0.3"
@@ -101,7 +102,7 @@ discord-bot = ["dep:serenity"]
 pdf = ["dep:lopdf"]
 epub = ["dep:epub", "dep:quick-xml"]
 rayon = ["dep:rayon"]
-wasm = ["dep:wasm-bindgen-futures", "futures-timer/wasm-bindgen"]
+wasm = ["dep:wasm-bindgen-futures", "futures-timer/wasm-bindgen", "getrandom/js"]
 rmcp = ["dep:rmcp"]
 socks = ["reqwest/socks"]
 reqwest-tls = ["reqwest/default"]


mylesberueda:
Closes #1312

joshua-mo-143:
Hi, thanks for the PR!

In terms of implementation details, since this is supposed to replace the provider ID/`tool_call_id` if it doesn't exist it might be better to have an enum like this:

```rust
#[derive(Deserialize, Serialize, Debug, Clone)]
#[serde(rename_all = "snake_case", tag = "tool_call_id_type")]
pub enum ToolCallId {
    Provider(String),  // From OpenAI, Anthropic, etc.
    Internal(String),  // Generated via nanoid
}

impl ToolCallId {
    get_id(&self) -> &str {
        match self {
            Self::Provider(string) -> &string,
            Self::Internal(string) -> &string
        }
    }
}
```

mylesberueda:
Hi Joshua, thanks for the review!

The internal_call_id is meant to be a stable correlation key for users tracking tool call events across timing boundaries such as parallel tool calls - not a replacement for the provider id, which I believe is needed to let the provider know which tool call we're responding to.

The problem: provider ids aren't reliable for streaming correlation:

- OpenAI: The id field only appears in the first chunk - subsequent chunks have id=None. Internally, OpenAI uses index for accumulation, but that resets each response and doesn't cross the delta -> call -> result boundary. OpenAI's impl shows that [id can be None](https://github.com/openai/openai-python/blob/dc68b90655912886bd7a6c7787f96005452ebfc9/src/openai/types/chat/chat_completion_chunk.py#L55) in their SDK.
- Gemini: Doesn't send tool call ids at all, and subsequent calls to the same tool overwrite previous results

My use case: I'm (basically) streaming rig events to a UI. When a `ToolCallDelta` arrives, I need to know which `ToolCall` it belongs to, and later correlate the `ToolResult` back to both.

How I see them:

- provider id: api protocol, tells LLM which tool call finished
- internal id: user correlation, informs users across tool events i.e. delta -> call -> result

Open to naming suggestions too - correlation_id? tracking_id? internal_call_id made sense when I wrote it, but now that I read it again it's unclear what this is used for.

Thanks again!

joshua-mo-143:
Hmmm, I suppose that makes sense. I'll bring it up in the internal sync today!

joshua-mo-143:
Edited PR to explicitly say this will be a breaking change.

Needs conflict resolution (and CI/review to pass), then we can add this in!

mylesberueda:
Thanks! I've rebased, added `internal_call_id` to bedrock and xai, and updated permission_control.rs with the new tool call signature (basically all the `just ci` stuff).

joshua-mo-143:
lgtm, thanks!

PR #1314, opened 2026-01-27
fix: avoid duplicate role in responses input

aurexav:
## Summary
- avoid duplicate "role" fields when serializing Responses API input items
- add regression test for InputItem serialization

## Testing
- cargo test -p rig-core --test openai_responses_input_item


joshua-mo-143:
needs to pass CI before I can review

joshua-mo-143:
lgtm, thanks!

Issue #1315, opened 2026-01-27
bug: deepseek-reasoner v3.2 compatibility issue


chenyh1990:
- [x] I have looked for existing issues (including closed) about this

## Bug Report

When using `deepseek-reasoner` to creata an agent, "Missing `reasoning_content` field" error occurs.

## Reproduction

```rust
        let client = deepseek::Client::from_env();
        let fin_agent = client
            .agent(deepseek::DEEPSEEK_REASONER)
            .preamble(PREAMPLE)
            .max_tokens(4 * 1024)
            .tools(tools)
            .build();

        let resp = fin_agent
            .prompt("Stock symbol of XXX in Hong Kong Stock Exchange")
            .await
            .unwrap();
        println!("Response: {}", resp);
```

Got this error:

```
called `Result::unwrap()` on an `Err` value: CompletionError(HttpError(InvalidStatusCodeWithMessage(400, "{\"error\":{\"message\":\"Missing `reasoning_content` field in the assistant message at message index 2. For more information, please refer to https://api-docs.deepseek.com/guides/thinking_mode#tool-calls\",\"type\":\"invalid_request_error\",\"param\":null,\"code\":\"invalid_request_error\"}}")))
```

## Expected behavior

Should work as deepseek-chat model.

## Additional context

Seems related to a server-side update in v3.2: https://github.com/langgenius/dify/issues/28988

fengwei-dev:
I’ve recently found that this issue is not limited to the `deepseek-reasoner` model. Following the release of DeepSeek-V4 (including both the Flash and Pro versions), similar problems have also occurred. After my testing, Kimi-2.6 is affected by the same issue as well.

Below are some related community discussion threads:
https://forum.cursor.com/t/kimi-2-6-custom-model-provider-returned-error-error-message-thinking-is-enabled-but-reasoning-content-is-missing/159625
https://github.com/MoonshotAI/kimi-cli/issues/1616
https://x.com/karminski3/status/2049926904120267144
https://github.com/QwenLM/qwen-code/issues/3619

As we can see, various tools are currently working urgently to fix this problem. Until a permanent solution is available, the only workaround is to disable the thinking mode.

Given this situation, could we extend the compatibility logic originally implemented for DeepSeek to **all OpenAI-compatible providers**? That is, pass through and return the `reasoning_content` field to the model regardless of whether it is empty or not.

PR #1317, opened 2026-01-28
fix(rig-1167): fix deepseek-reasoner v3.2 invoke

chenyh1990:
Fix rig-1167 (#1315)


joshua-mo-143 (approved):
lgtm - have you tested this works?

edit: I guess this doesn't work. You'll need to pass CI before a second review

PR #1328, opened 2026-01-30
feat(rig-1155): mock integration provider

### LINKED ISSUE #1290: feat(tests): add mock integration provider

## Feature Request

As much as mocks can be an anti-pattern, we need a way to competently test that integrations work without using up tokens (that is to say, we're frugal). The goal is to be able to test agent features without needing to connect to a third party provider and consume credits to check whether something works.

By creating a mock integration provider we can fully test all of our non-provider functionalities without actually having to use a provider. Which would be great. The model request/response schema should look similar to the OpenAI Chat Completions API (except no API key is required).

The provider should have an option to conditionally enable returning fixed tool function calls.

### Motivation

Test regressions cheaply/freely

### Proposal

```rust
struct MockClient {
    // .. etc
}

impl CompletionClient for MockClient {
    // .. etc
}

impl CompletionModel for MockModel {
    // .. etc
}
```

joshua-mo-143:
Fixes #1290

### DIFF
--- a/rig/rig-core/src/providers/mock.rs
+++ b/rig/rig-core/src/providers/mock.rs
@@ -8,8 +8,8 @@
 //!
 //! # Example
 //! ```rust,ignore
-//! use rig::providers::mock::{self, Client, ErrorMode};
-//! use rig::client::{CompletionClient, Nothing};
+//! use crate::providers::mock::{self, Client, ErrorMode};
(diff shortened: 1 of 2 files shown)


joshua-mo-143:
Closing for now. Not sure if we should merge this or not

PR #1357, opened 2026-02-04
feat(extractor): add with_agent_builder method for custom agent configuration

atellou:
### Summary
This PR improves the [ExtractorBuilder](cci:2://file:///Users/andrestello/Repos/external/rig/rig/rig-core/src/extractor.rs:211:0-219:1) API by adding a closure-based customization hook that lets callers configure the underlying `AgentBuilderSimple` (e.g. `dynamic_context`, [temperature](cci:1://file:///Users/andrestello/Repos/external/rig/rig/rig-core/src/completion/request.rs:665:4-669:5), [max_tokens](cci:1://file:///Users/andrestello/Repos/external/rig/rig/rig-core/src/completion/request.rs:677:4-682:5), etc.) while keeping the extractor’s opinionated defaults (preamble + required `submit` tool).

### Problem
[ExtractorBuilder](cci:2://file:///Users/andrestello/Repos/external/rig/rig/rig-core/src/extractor.rs:211:0-219:1) is implemented on top of `AgentBuilder`, but it only exposed a limited set of configuration methods. This made it awkward/impossible to reuse common agent configuration patterns (especially RAG via `dynamic_context`) when using extractors, and forced users toward workarounds.

### Solution
- Added [ExtractorBuilder::with_agent_builder](cci:1://file:///Users/andrestello/Repos/external/rig/rig/rig-core/src/extractor.rs:239:4-245:5):
  - Signature: `FnOnce(AgentBuilderSimple<M>) -> AgentBuilderSimple<M>`
  - Enables configuring agent builder options inside the extractor builder chain.
- Updated the `gemini_rag` example to use [.with_agent_builder(|agent| agent.dynamic_context(...))](cci:1://file:///Users/andrestello/Repos/external/rig/rig/rig-core/src/extractor.rs:239:4-245:5) to attach vector search context to the extractor.

### Tests
- Added unit test [extractor_builder_with_agent_builder_applies_config](cci:1://file:///Users/andrestello/Repos/external/rig/rig/rig-core/src/extractor.rs:383:4-392:5) validating that [with_agent_builder](cci:1://file:///Users/andrestello/Repos/external/rig/rig/rig-core/src/extractor.rs:239:4-245:5) successfully applies configuration (e.g. [temperature](cci:1://file:///Users/andrestello/Repos/external/rig/rig/rig-core/src/completion/request.rs:665:4-669:5), [max_tokens](cci:1://file:///Users/andrestello/Repos/external/rig/rig/rig-core/src/completion/request.rs:677:4-682:5)) to the built extractor’s inner [Agent](cci:2://file:///Users/andrestello/Repos/external/rig/rig/rig-core/src/agent/completion.rs:50:0-77:1).

### Examples
- Added / updated example: [rig-core/examples/gemini_rag.rs](cci:7://file:///Users/andrestello/Repos/external/rig/rig/rig-core/examples/gemini_rag.rs:0:0-0:0)
  - Demonstrates building embeddings for questionnaire docs, adding dynamic RAG context, and extracting structured questionnaire responses using the Gemini provider.

### Notes
- `cargo test -p rig-core extractor_builder_with_agent_builder_applies_config --features derive` passes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3806,7 +3806,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users 0.5.0",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -10048,6 +10048,27 @@ dependencies = [
 ]
 
 [[package]]
+name = "reqwest-retry"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "105747e3a037fe5bf17458d794de91149e575b6183fc72c85623a44abb9683f5"
+dependencies = [
+ "anyhow",
+ "async-trait",
+ "futures",
+ "getrandom 0.2.16",
+ "http 1.3.1",
+ "hyper 1.6.0",
+ "reqwest 0.12.24",
+ "reqwest-middleware",
+ "retry-policies",
+ "thiserror 2.0.16",
+ "tokio",
+ "tracing",
+ "wasmtimer 0.4.3",
+]
+
+[[package]]
 name = "resolv-conf"
 version = "0.7.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -10058,6 +10079,15 @@ name = "retain_mut"
 version = "0.1.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4389f1d5789befaf6029ebd9f7dac4af7f7e3d61b69d4f30e2ac02b57e7712b0"
+
+[[package]]
+name = "retry-policies"
+version = "0.5.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "46a4bd6027df676bcb752d3724db0ea3c0c5fc1dd0376fec51ac7dcaf9cc69be"
+dependencies = [
+ "rand 0.9.2",
+]
 
 [[package]]
 name = "revision"
@@ -10202,6 +10232,7 @@ dependencies = [
  "redis",
  "reqwest 0.12.24",
  "reqwest-middleware",
+ "reqwest-retry",
  "rig-derive",
  "rmcp",
  "schemars 1.0.4",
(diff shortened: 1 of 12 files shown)


joshua-mo-143:
To be honest we should *probably* really just re-combine the agent builders back together and have the tools thing as an enum so we don't have to deal with this. I'll open a sep issue to solve this

joshua-mo-143:
If you merge `main` back into your branch, it should now be much easier to `AgentBuilder<M, WithBuilderTools>` in the config rather than using one single config.

I think at the moment it will be difficult to use tool server handles with this type currently (which means that using the `WithToolServer` type state variant will be impossible with extractors) because the submit tool basically needs to be guaranteed to exist, but I think this is a reasonable tradeoff for now.

atellou:
Mmmmm got it, thanks for the fast response. In that case, shouldn't the toolchoice function be removed from the ExtractorBuilder?  Also, for my use case I only need to enable RAG. If we enable the dynamic context function, would that be ok?

```
    /// Add dynamic context (RAG) to the extractor.
    ///
    /// On each prompt, `sample` documents will be retrieved from the index based on the RAG text
    /// and inserted in the request.
    pub fn dynamic_context(
        mut self,
        sample: usize,
        dynamic_context: impl VectorStoreIndexDyn + Send + Sync + 'static,
    ) -> Self {
        self.agent_builder = self.agent_builder.dynamic_context(sample, dynamic_context);
        self
    }

joshua-mo-143:
> Mmmmm got it, thanks for the fast response. In that case, shouldn't the toolchoice function be removed from the ExtractorBuilder? Also, for my use case I only need to enable RAG. If we enable the dynamic context function, would that be ok?
> 
> ```
>     /// Add dynamic context (RAG) to the extractor.
>     ///
>     /// On each prompt, `sample` documents will be retrieved from the index based on the RAG text
>     /// and inserted in the request.
>     pub fn dynamic_context(
>         mut self,
>         sample: usize,
>         dynamic_context: impl VectorStoreIndexDyn + Send + Sync + 'static,
>     ) -> Self {
>         self.agent_builder = self.agent_builder.dynamic_context(sample, dynamic_context);
>         self
>     }
> ```

Yeah that's fine.

atellou:
moved to this https://github.com/0xPlaygrounds/rig/pull/1456

Issue #1360, opened 2026-02-05
chore: Propagate current span to tool call


redroy44:
### Problem

ToolServer uses an mpsc channel + tokio::spawn to decouple tool execution from the caller. This breaks tracing span context propagation — any spans created inside a `Tool::call()` implementation become orphaned root spans instead of children of the caller's span tree.

Reproduction: Any application using OpenTelemetry with rig-core that instruments its `Tool::call()` implementations (via `#[instrument]` or manual spans) will see tool execution traces disconnected from the agent's request trace. The tool spans appear as
  separate traces in backends like Jaeger, Grafana Tempo, etc.

  Expected:
```
  http.request
    └── agent.completion
         └── execute_tool        ← rig's own span
              └── my_tool_span   ← user's Tool::call() span
```
  Actual:
```
  http.request
    └── agent.completion
         └── execute_tool

  my_tool_span                   ← orphaned, separate trace
```

PR #1361, opened 2026-02-05
chore: Propagate current span to tool call

redroy44:
### Problem

ToolServer uses an mpsc channel + tokio::spawn to decouple tool execution from the caller. This breaks tracing span context propagation — any spans created inside a `Tool::call()` implementation become orphaned root spans instead of children of the caller's span tree.

Reproduction: Any application using OpenTelemetry with rig-core that instruments its `Tool::call()` implementations (via `#[instrument]` or manual spans) will see tool execution traces disconnected from the agent's request trace. The tool spans appear as
  separate traces in backends like Jaeger, Grafana Tempo, etc.

  Expected:
```
  http.request
    └── agent.completion
         └── execute_tool        ← rig's own span
              └── my_tool_span   ← user's Tool::call() span
```
  Actual:
```
  http.request
    └── agent.completion
         └── execute_tool

  my_tool_span                   ← orphaned, separate trace
```

### Solution

  Capture the caller's `tracing::Span` in `ToolServerHandle::call_tool()`, send it through the channel as part of `ToolServerRequestMessageKind::CallTool`, and `.instrument(span)` the tool execution future on the receiving side.

`tracing::Span` is `Send + Sync`, so it can be sent through the mpsc channel. `.instrument(span)` causes the spawned future to enter the captured span when polled, restoring the parent-child relationship for any child spans created during tool execution.

### Changes

  1. `ToolServerRequestMessageKind::CallTool` — add span: `tracing::Span` field
  2. `ToolServerHandle::call_tool()` — capture `tracing::Span::current()` into the message
  3. `ToolServer::handle_message()` — instrument the tool call future with the captured span

### Backwards compatibility

  ToolServerRequest has private fields, so external code cannot construct it. The enum variant change is invisible to consumers who only interact through ToolServerHandle.

Fixes https://github.com/0xPlaygrounds/rig/issues/1360

### DIFF
--- a/rig/rig-core/src/tool/server.rs
+++ b/rig/rig-core/src/tool/server.rs
@@ -5,6 +5,7 @@ use tokio::sync::{
     RwLock,
     mpsc::{Sender, error::SendError},
 };
+use tracing::Instrument;
 
 use crate::{
     completion::{CompletionError, ToolDefinition},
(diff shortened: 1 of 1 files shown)


joshua-mo-143:
lgtm - thanks for the PR!

PR #1363, opened 2026-02-05
Disable default features on aws-bedrock-runtime

Sagebati:
Hi, 

The default features on aws-bedrock-runtime pulls hyper 0.14. Removing it can improve compile times.


joshua-mo-143:
Hi, can you add an opt-in feature for the rustls feature? Users might still need `rustls`.

Sagebati:
Done, thanks for the feedback!

joshua-mo-143:
Looks like another round of making sure CI passes is required

PR #1384, opened 2026-02-12
fix(agents): correct prompt hook docs, split modules, and fix install script

### LINKED ISSUE #1343: docs: Rig skill
We should make a Rig skill so it's easy for people to write Rig agents using their coding agent of choice.

Mostly putting this here as a reminder to self.

illegalcall:
## Summary

- Split `rig-quality-gates` module into `rig-prompt-hooks` (hook API docs) + `rig-quality-gates` (coding style & quality standards), fixing prompt hook content leaking into unrelated adapter files (e.g. vector-store)
- Fixed incorrect API references in AGENTS.md: `CancelSignal::cancel()` → `ToolCallHookAction::Terminate`/`HookAction::Terminate`, removed fictional dual-trait (`PromptHook` + `StreamingPromptHook`) section, added `#[cfg]` attributes to WasmCompat traits, expanded `Capabilities` example with `Capable<T>`/`Nothing`
- Fixed `install-rig-skill.sh`: `install_gemini` now appends instead of overwriting, `SKILL_SOURCE` detection no longer changes cwd
- Fixed grep pattern in `rig-migrate` skill and `Self::NAME` usage in `tools.md`
- Set executable bit (`100755`) on all three shell scripts — fixes CI "Permission denied" on `sync_agent_instruction_files.sh`
- Regenerated all 27 adapter files via sync script; validation passes

## Test plan

- [x] `scripts/sync_agent_instruction_files.sh` runs without errors
- [x] `scripts/validate_agent_instruction_files.sh` confirms all adapters in sync
- [x] Spot-check Claude Code skills (`/rig-agent-scaffold`, `/rig-provider-checklist`, `/rig-review-gates`)
- [x] Verify `install-rig-skill.sh --agent gemini` appends correctly on a clean target

Fixes #1343

### DIFF
--- a/.claude/commands/rig-agent-scaffold.md
+++ b/.claude/commands/rig-agent-scaffold.md
@@ -43,7 +43,7 @@ Nearly every configurable type uses the builder pattern:
 
 ```rust
 let agent = client
-    .agent(openai::GPT_4O)
+    .agent(openai::GPT_5_2)
     .preamble("System prompt")
     .tool(my_tool)
     .temperature(0.8)
--- a/.claude/skills/rig-agent-scaffold/SKILL.md
+++ b/.claude/skills/rig-agent-scaffold/SKILL.md
@@ -47,7 +47,7 @@ Nearly every configurable type uses the builder pattern:
 
 ```rust
 let agent = client
-    .agent(openai::GPT_4O)
+    .agent(openai::GPT_5_2)
     .preamble("System prompt")
     .tool(my_tool)
     .temperature(0.8)
--- a/.clinerules/01-rig-core.md
+++ b/.clinerules/01-rig-core.md
@@ -30,7 +30,7 @@ Nearly every configurable type uses the builder pattern:
 
 ```rust
 let agent = client
-    .agent(openai::GPT_4O)
+    .agent(openai::GPT_5_2)
     .preamble("System prompt")
     .tool(my_tool)
     .temperature(0.8)
--- a/.continue/rules/rig-core.md
+++ b/.continue/rules/rig-core.md
@@ -29,7 +29,7 @@ Nearly every configurable type uses the builder pattern:
 
 ```rust
 let agent = client
-    .agent(openai::GPT_4O)
+    .agent(openai::GPT_5_2)
     .preamble("System prompt")
     .tool(my_tool)
     .temperature(0.8)
--- a/.cursor/rules/rig-core.mdc
+++ b/.cursor/rules/rig-core.mdc
@@ -33,7 +33,7 @@ Nearly every configurable type uses the builder pattern:
 
 ```rust
 let agent = client
-    .agent(openai::GPT_4O)
+    .agent(openai::GPT_5_2)
     .preamble("System prompt")
     .tool(my_tool)
     .temperature(0.8)
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -33,7 +33,7 @@ Nearly every configurable type uses the builder pattern:
 
 ```rust
 let agent = client
-    .agent(openai::GPT_4O)
+    .agent(openai::GPT_5_2)
     .preamble("System prompt")
     .tool(my_tool)
     .temperature(0.8)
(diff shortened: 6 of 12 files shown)


scripts/install-rig-skill.sh
```diff
@@ -0,0 +1,421 @@
+#!/usr/bin/env bash
```
joshua-mo-143, line 1: How exactly will this file be used? This doesn't seem like something that will see external usage

.claude/skills/rig-agent-scaffold/SKILL.md
```diff
@@ -0,0 +1,344 @@
...
+
+New implementations should implement these traits rather than creating parallel abstractions.
+
+### Builder Pattern
+
+Nearly every configurable type uses the builder pattern:
+
+```rust
+let agent = client
+    .agent(openai::GPT_4O)
```
joshua-mo-143: This should be `openai::GPT_5_2`. We probably need to update this elsewhere

scripts/sync_agent_instruction_files.sh
```diff
@@ -0,0 +1,1163 @@
+#!/usr/bin/env bash
```
joshua-mo-143, line 1: Not sure this is necessary. Users can basically just symlink and it'll be fine.

If we do add this, we should gitignore whichever files get generated from this so that it doesn't clog the repo.

.claude/commands/rig-agent-scaffold.md
```diff
@@ -0,0 +1,340 @@
+<!-- This file is generated by scripts/sync_agent_instruction_files.sh. -->
```
joshua-mo-143, line 1: nit: Haven't skills replaced slash commands in Claude (Code)? I was sure this was the case unless I'm mistaken on this

joshua-mo-143 (changes requested):
Looks mostly OK, see comments

scripts/install-rig-skill.sh
illegalcall: This is for people building with Rig. If someone is using Rig as a dependency in their project and they're using an AI coding agent (Claude, Cursor, Copilot, Windsurf, etc.), they can run 
`install-rig-skill.sh --agent claude `
to install Rig's coding guidelines directly into their agent's context. That way their agent understands Rig's patterns, trait-based abstractions, builder pattern, WASM compatibility, etc and generates code that actually follows the  framework's conventions instead of guessing. It supports multiple agent platforms so it's not locked to one tool.

.claude/commands/rig-agent-scaffold.md
illegalcall: Yes, but I have added them for backward compatibility. In case you decide its not a priority, I can remove it.

scripts/sync_agent_instruction_files.sh
illegalcall: Symlinks would give every skill the entire AGENTS.md, but the sync script extracts specific module blocks per skill e.g. the vector-store skill only gets rig-vector-store-implementation + rig-wasm-compat + rig-quality-gates, not the full document. That scoped context is the point.

On gitignoring the generated files, I'd rather keep them checked in so contributors get working skills on clone without an extra setup step. The validation script ensures they never go stale.

.claude/commands/rig-agent-scaffold.md
joshua-mo-143: hmmmmmm, okay I think we can probably keep this in. I guess if Anthropic removes this, we can update it.

scripts/sync_agent_instruction_files.sh
joshua-mo-143: Oh I see, that makes sense.

Hmmmm, yeah I guess. I am not entirely sure on this since it will clog the toplevel dir (since it seems that we're basically trying to support every single agentic DE out there) but it will definitely make things easier

scripts/install-rig-skill.sh
joshua-mo-143: Hmmmm, okay I see. That means we'll need to add documentation for this elsewhere then

.claude/skills/rig-agent-scaffold/SKILL.md
illegalcall: @joshua-mo-143 Updated.

PR #1385, opened 2026-02-12
add client builder test to all providers

hghalebi:
Added client builder test to all providers.
**Notice:** My RustRover IDE formatter changed some formatting that is just a reordering of imports... . I dont know if it is an issue.


hghalebi:
@joshua-mo-143

PR #1395, opened 2026-02-13
refactor: typed reasoning content model

### LINKED ISSUE #684: refactor: reasoning should not be Vec<String>

## Feature Request

Reasoning was recently made `Vec<String>` to accommodate for the fact that OpenAI seems to have made a very strange choice in making reasoning an array of strings.

We should probably figure out how to deal with this issue as while it does now work, it is also a bit of an unnecessary DX papercut.

### Motivation

Improve DX

darinkishore:
## Summary

- Introduces `ReasoningContent` enum (`Text`, `Encrypted`, `Redacted`, `Summary`) replacing unstructured string-based reasoning representation
- Adds `Reasoning` struct with `id`, `content: Vec<ReasoningContent>` and builder methods (`new`, `with_id`, `new_with_signature`, etc.)
- Adds `ReasoningDelta` variant and `MessageId` variant to `RawStreamingChoice` for streaming infrastructure
- Adds `message_id: Option<String>` to `CompletionResponse` and `StreamingCompletionResponse` for provider-assigned IDs (needed by OpenAI Responses API)
- Removes panic paths in unsupported provider conversion paths, replacing with proper error propagation
- Updates streaming agent loop to accumulate reasoning blocks and yield `Reasoning`/`ReasoningDelta` events

Closes #684

## Breaking API Changes

| Item | Migration |
|------|-----------|
| `AssistantContent::Reasoning` now wraps `Reasoning` struct instead of raw fields | Use `Reasoning::new("text")`, `Reasoning::new_with_signature("text", "sig")`, or match on `reasoning.content` |
| `StreamedAssistantContent::Reasoning` carries `Reasoning` instead of old shape | Pattern match on `Reasoning` struct |
| `StreamedAssistantContent::ReasoningDelta` fields changed | Now `{ reasoning: String, id: Option<String> }` |
| `CompletionResponse<T>` has new `message_id: Option<String>` field | Add `message_id: None` when constructing |
| `RawStreamingChoice` has new `Reasoning`, `ReasoningDelta`, `MessageId` variants | Add arms to exhaustive matches |

## Test plan

- [x] `cargo fmt` clean
- [x] `cargo clippy --all-targets --all-features` clean
- [x] `cargo test -p rig-core --lib` — 271 tests pass
- [x] Existing `openai_responses_input_item` tests updated and passing

### DIFF
--- a/rig/rig-core/src/agent/prompt_request/streaming.rs
+++ b/rig/rig-core/src/agent/prompt_request/streaming.rs
@@ -122,13 +122,8 @@ fn merge_reasoning_blocks(
     }
 }
 
-async fn cancelled_prompt_error(
-    chat_history: &Arc<RwLock<Vec<Message>>>,
-    reason: String,
-) -> StreamingError {
-    StreamingError::Prompt(
-        PromptError::prompt_cancelled(chat_history.read().await.to_vec(), reason).into(),
-    )
+async fn cancelled_prompt_error(chat_history: &Vec<Message>, reason: String) -> StreamingError {
+    StreamingError::Prompt(PromptError::prompt_cancelled(chat_history.to_owned(), reason).into())
 }
 
 fn tool_result_to_user_message(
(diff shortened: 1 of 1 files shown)


joshua-mo-143:
Ah crap I think I broke something while merging main back in. I'll fix this up then we can get it merged

joshua-mo-143:
Looks like the LLM hallucinated, `message_id` actually exists in #1396 so it's not technically in this PR.

Otherwise, lgtm so I'll be merging this

PR #1401, opened 2026-02-16
fix(rig-1193): terminate agent stream on error

### LINKED ISSUE #1399: bug: `on_text_delta termination` action does not stop the streaming from AI

## Bug Report
After update 0.30, when returning `HookAction::terminate(reason)`, the streaming does not stop. Looking at the code, in `streaming.rs` line 510, an error is yielded instead of returned.

## Reproduction
Install version 0.30 and return termination on `on_text_delta`

## Expected behavior
The streaming should stop.

joshua-mo-143:
~~Fixes an issue where instead of terminating the entire stream on a termination action for on_text_delta, the error is incorrectly yielded instead.~~ Fixes an error where the stream continues to yield items even after yielding a termination error.

Closes #1399


joshua-mo-143:
Superceded by #1395

PR #1407, opened 2026-02-18
feat(azure-openai): Add structured outputs support

austinsimpsond41:
Hi @joshua-mo-143,

Thank you for getting structured outputs into Rig! I've copied the code from the regular OpenAI provider into the Azure OpenAI provider.

I also noticed that the ToolChoice struct was being imported from openrouter instead of openai. I'm assuming this was done by mistake, and fixed it. I wanted to call this out on the off chance it wasn't a mistake. 

I added a test for structured outputs as well, and it passed without issues.

Let me know if you need anything else.

Thank you!

### DIFF
--- a/rig/rig-core/src/providers/azure.rs
+++ b/rig/rig-core/src/providers/azure.rs
@@ -1058,7 +1058,7 @@ mod audio_generation {
 
 #[cfg(test)]
 mod azure_tests {
-    use schemars::{JsonSchema, schema_for};
+    use schemars::JsonSchema;
 
     use super::*;
 


joshua-mo-143:
Hey Austin thanks for the PR! I'm currently away from the computer for a couple of days due to Rust Nation but will get back to you asap

austinsimpsond41:
Enjoy! Talk to you when you're back.

rig/rig-core/src/providers/azure.rs
```diff
@@ -610,9 +610,36 @@ impl TryFrom<(&str, CompletionRequest)> for AzureOpenAICompletionRequest {
...
+                    }
+                }
+            });
+            Some(match req.additional_params {
+                Some(existing) => json_utils::merge(existing, response_format),
+                None => response_format,
+            })
+        } else {
+            req.additional_params
+        };
```
joshua-mo-143, line 641: I think here instead of trying to wrangle raw JSON let's add this as a proper field to the outputted `AzureOpenAICompletionRequest`. This particular snippet seems quite fragile

joshua-mo-143 (changes requested):
Just one comment!

Rest looks fine.

rig/rig-core/src/providers/azure.rs
austinsimpsond41: Okay. Should the regular openai provider be updated as well to match?

joshua-mo-143: Oh, huh. I didn't even realise it was the same in the Completions integration - my bad.

OK let's fragment that as a separate ticket and just push this through since barring that, everything else is fine

Issue #1408, opened 2026-02-19
bug/feat: tool call from llama.cpp fails


inqode-lars:
- [x] I have looked for existing issues (including closed) about this

## Bug Report

Tool call from `llama.cpp` can't be deserialized.

```
CompletionError: JsonError: data did not match any variant of untagged enum ApiResponse
```

## Reproduction

```rust
let client: openai::Client = openai::Client::builder()
    .base_url("llama.cpp url")
    .build()
    .expect("Failed to build client");
let agent = client
    .completions_api()
    .agent(&config.model.name)
    .preamble("You are a helpful assistant with access to tools.")
    .tool(HelloWorld)
    .build();

match agent.prompt("Call the hello_world tool and give me the output.").await {
    Ok(response) => {
        println!("\nResponse from model:");
        println!("  {}", response);
    }
    Err(e) => {
        eprintln!("Error: {}", e);
    }
}
```

the model gives:

```json
{
  "choices": [{
      "finish_reason": "tool_calls",
      "index": 0,
      "message": {
          "role": "assistant",
          "content": "",
          "tool_calls": [{ "type": "function", "function": { "name": "hello_world", "arguments": {} }, "id": "xxx" }]
      }
  }],
  "created": 0,
  "model": "unsloth/Qwen3-Coder-Next-GGUF:Q8_0",
  "system_fingerprint": "b8113-xxxx",
  "object": "chat.completion",
  "usage": { "completion_tokens": 13, "prompt_tokens": 255, "total_tokens": 268 },
  "id": "xxx",
  "timings": {
      "cache_n": 0,
      "prompt_n": 255,
      "prompt_ms": 670,
      "prompt_per_token_ms": 2.63,
      "prompt_per_second": 380,
      "predicted_n": 13,
      "predicted_ms": 367,
      "predicted_per_token_ms": 28,
      "predicted_per_second": 35
  }
}
```

I get the error:

```
CompletionError: JsonError: data did not match any variant of untagged enum ApiResponse
```

## Expected behavior

I would be nice if the `llama.cpp` tool calling would be supported. And i could get the output of the `hello_world` tool.

## Additional context

The problem is probably that `llama.cpp` is not exactly compatible to the openai API, because it gives the tool call `arguments` as an map. The openai API expects them to be a JSON string.

[...]

PR #1409, opened 2026-02-19
feat(rig-1197): handle llama.cpp tool call (#1408)

inqode-lars:
handle llama.cpp tool call

closes https://github.com/0xPlaygrounds/rig/issues/1408

### DIFF
--- a/rig/rig-core/src/providers/openai/completion/mod.rs
+++ b/rig/rig-core/src/providers/openai/completion/mod.rs
@@ -16,8 +16,7 @@ use crate::one_or_many::string_or_one_or_many;
 use crate::telemetry::{ProviderResponseExt, SpanCombinator};
 use crate::wasm_compat::{WasmCompatSend, WasmCompatSync};
 use crate::{OneOrMany, completion, json_utils, message};
-use serde::{Deserialize, Deserializer, Serialize, Serializer};
-use serde_json::Value;
+use serde::{Deserialize, Serialize, Serializer};
 use std::convert::Infallible;
 use std::fmt;
 use tracing::{Instrument, Level, enabled, info_span};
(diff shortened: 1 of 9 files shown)


rig/rig-core/src/providers/openai/completion/mod.rs
```diff
@@ -402,10 +403,25 @@ impl TryFrom<crate::message::ToolChoice> for ToolChoice {
...
+    D: Deserializer<'de>,
+{
+    let value = Value::deserialize(deserializer)?;
+
+    match value {
+        Value::String(s) => serde_json::from_str(&s).map_err(serde::de::Error::custom),
+        other => Ok(other),
+    }
+}
+
```
joshua-mo-143: please add docstring to explain why this is required (it's not immediately clear to anyone who reads this in the future if they don't refer back to this PR)

rig/rig-core/src/providers/openai/completion/mod.rs
```diff
@@ -402,10 +403,25 @@ impl TryFrom<crate::message::ToolChoice> for ToolChoice {
 #[derive(Debug, Serialize, Deserialize, PartialEq, Clone)]
 pub struct Function {
     pub name: String,
-    #[serde(with = "json_utils::stringified_json")]
+    #[serde(
+        serialize_with = "json_utils::stringified_json::serialize",
```
joshua-mo-143, line 406: Is this correct? It seems to have been perfectly fine before to my knowledge

rig/rig-core/src/providers/openai/completion/mod.rs
inqode-lars: good idea. Added a comment for that.

rig/rig-core/src/providers/openai/completion/mod.rs
inqode-lars: For openai its fine. But with llama.cpp there is a problem. See here https://github.com/0xPlaygrounds/rig/issues/1408

gold-silver-copper:
This was a good PR overall, I only made some minor changes and tagged on some extra bug fixes.
I created a shared deserialization helper between hugging face and openai (for llama.cpp) since the code was the same.
I standardized all providers to convert `""` args to `{}`, this follows the behavior of other providers, and is in line with other inference abstractions such as here: https://github.com/vllm-project/vllm/issues/19419
Added a few extra tests.

Finally I tested this PR on a local llama.cpp instance and it was successful.

PR #1438, opened 2026-02-25
fix: allow empty arguments openrouter

### LINKED ISSUE #1437: bug(openrouter): deserialisation fails if tool arguments are empty

## Bug Report
When the MCP tool requires no arguments, OpenRouter will send an `"arguments": ""`, which fails to deserialise.

The response is also extremely unclear and vague:
```
JsonError(Error("data did not match any variant of untagged enum ApiResponse", line: 0, column: 0))
```

## Expected behavior
MCP tools with no arguments should work fine.

CremboC:
Fixes #1437 

Also, I've made the error now include the response body because in order to debug this, I had to do `cargo patch` which wasn't a very nice experience 😅

### DIFF
--- a/rig/rig-core/src/providers/openrouter/completion.rs
+++ b/rig/rig-core/src/providers/openrouter/completion.rs
@@ -1743,7 +1743,15 @@ where
             let response_body = response.into_body().into_future().await?.to_vec();
 
             if status.is_success() {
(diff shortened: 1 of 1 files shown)


joshua-mo-143:
lgtm!

PR #1396, opened 2026-02-13
feat: cross-provider reasoning trace roundtrip

darinkishore:
> **Stacked on #1395** — review/merge that first. The diff here includes both PRs; the incremental diff is only the provider-specific changes and integration tests.

## Summary

Builds on #1395 to implement full reasoning trace preservation across all providers, ensuring reasoning content survives multi-turn conversations including tool call loops.

- **OpenAI Responses API**: Threads `msg_` IDs through streaming/non-streaming paths; reverts strip-reasoning workaround; emits `MessageId` from `OutputItemDone`; includes `reasoning_encrypted_content` automatically when reasoning is configured
- **Anthropic**: Multi-block `Reasoning` → `Thinking`/`RedactedThinking` conversion with signature preservation; streaming `RedactedThinking` support via `get_or_insert_with`
- **Gemini**: `thought_signature` roundtrip preservation; emits full `Reasoning` block (not just delta) when signature is present in streaming
- **xAI**: Structured `Reasoning` ↔ Chat Completions `reasoning_content` conversion with encrypted/redacted mapping; fixes `#[serde(rename)]` → `#[serde(rename_all)]`
- **OpenRouter**: Groups `reasoning_details` by ID preserving ordering; maps `Encrypted`/`Redacted` content types through Chat Completions format
- **Streaming agent loop bug fix**: `ReasoningDelta` events now accumulate in a separate text buffer instead of being converted to unsigned `Reasoning` blocks — fixes Anthropic 400 "signature required" errors in multi-turn tool call loops

Closes #1147, closes #1146

## Integration Tests

Four test suites covering both streaming and non-streaming code paths with per-provider coverage:

### Streaming Tests

**`reasoning_roundtrip.rs` (streaming)** — Text-only 2-turn streaming conversation:
- Turn 1: `model.stream()` → collect reasoning blocks/deltas + text
- Turn 2: follow-up with reasoning in chat history → provider accepts

**`reasoning_tool_roundtrip.rs` (streaming)** — Reasoning + tool call via `agent.stream_chat().multi_turn()`:
- Agent receives weather prompt → model thinks → calls `get_weather` tool
- Agent loop preserves `[Reasoning, ToolCall]` in chat history
- Agent sends tool result → model responds with text referencing tool output
- 11 universal assertions per provider (no errors, tool invoked, reasoning before tool call, etc.)
- Provider-specific assertions (OpenAI: Encrypted content, Anthropic: signatures)

### Non-Streaming Tests

[...]

PR #1441, opened 2026-02-25
fix(openai): add reasoning_content to StreamingDelta for OpenAI-compatible providers

### LINKED ISSUE #1440: bug(openai): StreamingDelta missing reasoning_content — reasoning silently dropped for OpenAI-compatible providers
## Bug Description

The OpenAI Chat Completions streaming parser (`StreamingDelta` in `rig-core/src/providers/openai/completion/streaming.rs`) does not include a `reasoning_content` field. This causes reasoning/thinking content from OpenAI-compatible providers to be **silently dropped** during streaming.

`RawStreamingChoice::ReasoningDelta` was added in PR #1395, and PR #1396 implemented reasoning support for 5 specialized providers (OpenAI Responses API, Anthropic, Gemini, xAI, OpenRouter). However, the **generic OpenAI Chat Completions streaming parser** — used by `send_compatible_streaming_request` — was not updated.

The DeepSeek provider has its own `StreamingDelta` with `reasoning_content` (line 722 of `deepseek.rs`), so it works correctly. But any provider using the generic OpenAI-compatible path loses reasoning content.

## Affected Providers

All models that send `reasoning_content` via the standard Chat Completions streaming format:

- **GLM-4.7** (Zhipu AI) — verified, 193 reasoning deltas silently dropped
- **DeepSeek models via OpenAI-compatible endpoints** (not the dedicated provider)
- **Qwen with thinking mode** (Alibaba DashScope)
- **vLLM / Ollama** serving reasoning models via OpenAI-compatible API
- Any other OpenAI-compatible endpoint returning `delta.reasoning_content`

## Root Cause

`StreamingDelta` (line 35) only has `content` and `tool_calls`:

`
ust
struct StreamingDelta {
    #[serde(default)]
    content: Option<String>,
    #[serde(default, deserialize_with = "json_utils::null_or_vec")]
    tool_calls: Vec<StreamingToolCall>,
}
`

The `reasoning_content` field from SSE chunks like:
`json
{"choices":[{"delta":{"reasoning_content":"Let me think...","content":null}}]}
`
is silently ignored by serde (`#[serde(default)]` not present → field not parsed).

## Suggested Fix

Two changes in `rig-core/src/providers/openai/completion/streaming.rs`:

1. Add `reasoning_content` to `StreamingDelta`:
`
ust
struct StreamingDelta {
    #[serde(default)]
    content: Option<String>,
    #[serde(default)]
    reasoning_content: Option<String>,  // ← add this
    #[serde(default, deserialize_with = "json_utils::null_or_vec")]
    tool_calls: Vec<StreamingToolCall>,
}
`

2. Yield `ReasoningDelta` in the stream (before the text content block):
`
ust
if let Some(reasoning) = &delta.reasoning_content && !reasoning.is_empty() {
    yield Ok(streaming::RawStreamingChoice::ReasoningDelta {
        id: None,
        reasoning: reasoning.clone(),
    });
}
`

This matches the existing pattern in the DeepSeek provider (`deepseek.rs` line 840).

## Reproduction

1. Configure an OpenAI-compatible provider with a reasoning model (e.g. GLM-4.7 via `https://open.bigmodel.cn/api/coding/paas/v4`)
2. Use `model.stream()` or `agent.stream_chat()`
3. Observe: no `ReasoningDelta` / `Reasoning` events in the stream
4. The model's thinking content is silently discarded

## Environment

- rig-core: 0.31.0 (latest main)
- Tested with: GLM-4.7 (Zhipu AI OpenAI-compatible API)

Fromsko:
Fixes #1440

## Summary

The generic OpenAI Chat Completions streaming parser (`StreamingDelta`) does not parse the `reasoning_content` field from SSE chunks, causing reasoning/thinking content to be **silently dropped** for all OpenAI-compatible providers.

This is a 2-line fix that adds `reasoning_content` support to `send_compatible_streaming_request`, matching the existing pattern in the DeepSeek-specific provider.

## Changes

**File:** `rig-core/src/providers/openai/completion/streaming.rs`

1. **Add `reasoning_content` field** to `StreamingDelta` struct (line 38-39)
2. **Yield `RawStreamingChoice::ReasoningDelta`** when reasoning content is present in the stream (lines 250-256)

## Context

- `RawStreamingChoice::ReasoningDelta` was added in PR #1395 and is used by specialized providers (Anthropic, Gemini, xAI, OpenRouter) via PR #1396
- The DeepSeek provider has its own `StreamingDelta` with `reasoning_content` (`deepseek.rs:722`), so it works independently
- The **generic OpenAI streaming parser** was the only one missing this field

## Affected Providers

All models using `reasoning_content` via the standard Chat Completions streaming format:
- GLM-4.7 (Zhipu AI) — **verified**: 193 reasoning deltas now captured correctly
- DeepSeek via OpenAI-compatible endpoints
- Qwen with thinking mode (DashScope)
- vLLM / Ollama serving reasoning models

## Testing

Verified end-to-end with GLM-4.7:
- Before fix: 0 `ReasoningDelta` events (reasoning silently dropped)
- After fix: **193 `ReasoningDelta` events** captured, full thinking chain preserved

Existing tests pass — the new `reasoning_content` field uses `#[serde(default)]` so it's backward-compatible (absent field → `None`).

### DIFF
--- a/rig/rig-core/src/providers/openai/completion/streaming.rs
+++ b/rig/rig-core/src/providers/openai/completion/streaming.rs
@@ -12,7 +12,7 @@ use crate::completion::{CompletionError, CompletionRequest, GetTokenUsage};
 use crate::http_client::HttpClientExt;
 use crate::http_client::sse::{Event, GenericEventSource};
 use crate::json_utils::{self, merge};
-use crate::providers::openai::completion::{self, CompletionModel, OpenAIRequestParams, Usage};
+use crate::providers::openai::completion::{CompletionModel, OpenAIRequestParams, Usage};
 use crate::streaming::{self, RawStreamingChoice};
 
 // ================================================================
@@ -160,7 +160,6 @@ where
         // Accumulate tool calls by index while streaming
         let mut tool_calls: HashMap<usize, streaming::RawStreamingToolCall> = HashMap::new();
         let mut text_content = String::new();
-        let mut final_tool_calls: Vec<completion::ToolCall> = Vec::new();
         let mut final_usage = None;
 
         while let Some(event_result) = event_source.next().await {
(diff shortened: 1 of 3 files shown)


rig/rig-core/src/providers/openai/completion/streaming.rs
```diff
@@ -35,6 +35,8 @@ pub(crate) struct StreamingToolCall {
 struct StreamingDelta {
     #[serde(default)]
     content: Option<String>,
+    #[serde(default)]
+    reasoning_content: Option<String>,
```
joshua-mo-143: please add docstring to state that this is not part of the official OpenAI API, it's for compliant providers

Fromsko: He should look like this; I've already changed it. 
Because the change has already been marked in another location.
Thanks for the review.
```rust
#[serde(default)]
reasoning_content: Option<String>, // This is not part of the official OpenAI API
```

PR #1442, opened 2026-02-26
fix(deps): enable reqwest system-proxy for proxy env var support

Phoenix500526:
reqwest 0.13 moved system proxy support behind a feature flag (system-proxy), unlike 0.12 where it was always enabled. Without this feature, reqwest ignores $http_proxy/$https_proxy environment variables and attempts direct TCP connections, which fails in environments that require a proxy for external access.


joshua-mo-143:
lgtm

PR #1446, opened 2026-02-26
feat: add `.extended_details` to `TypedPromptRequest` via typestate

0xMochan:
Updated the typestate to match normal `PromptRequest`.
I used the linear branch name, hopefully thats ok!

We could also make the normal `PromptRequest` generic on type T making the output `String` for normal sitatuations and save a type.


joshua-mo-143 (approved):
lgtm

PR #1450, opened 2026-02-28
feat(rig-core): return conversation messages from non-streaming agent…

illegalcall:
## Summary
- Adds `messages: Option<Vec<Message>>` field to `PromptResponse` (with `#[non_exhaustive]`)
- Populated only via `extended_details()` path — Standard path pays zero cost
- Internal `collect_messages` bool + private `into_extended()` avoids unnecessary `chat_history.to_vec()` on Standard delegation
- Mirrors existing streaming pattern (`FinalResponse::history`)

## Changes
- `rig-core/src/agent/prompt_request/mod.rs`: `collect_messages` field, conditional message collection, `PromptResponse::with_messages()` builder
- `rig-core/tests/prompt_response_messages.rs`: 10 integration tests (mock models, multi-turn, backward compat, error paths)

## Test plan
- [x] `cargo test -p rig-core --lib --tests` (396 passed)
- [x] `cargo check --workspace` (clean)
- [x] Standard path backward compat (returns `String`, no perf regression)
- [x] `extended_details()` populates messages (single-turn + multi-turn with tool calls)
- [x] `with_history()` + `extended_details()` interop
- [x] `MaxTurnsError` path unaffected
- [x] `PromptResponse::new()` 2-arg signature preserved

### DIFF
--- a/rig/rig-core/src/agent/prompt_request/mod.rs
+++ b/rig/rig-core/src/agent/prompt_request/mod.rs
@@ -125,11 +125,12 @@ where
     M: CompletionModel,
     P: PromptHook<M>,
 {
-    /// Enable returning extended details for responses (includes aggregated token usage)
+    /// Enable returning extended details for responses (includes aggregated token usage
+    /// and the full message history accumulated during the agent loop).
     ///
     /// Note: This changes the type of the response from `.send` to return a `PromptResponse` struct
     /// instead of a simple `String`. This is useful for tracking token usage across multiple turns
-    /// of conversation.
+    /// of conversation and inspecting the full message exchange.
     pub fn extended_details(self) -> PromptRequest<'a, Extended, M, P> {
         PromptRequest {
             prompt: self.prompt,
@@ -240,9 +241,19 @@ where
 }
 
 #[derive(Debug, Clone)]
+#[non_exhaustive]
 pub struct PromptResponse {
     pub output: String,
     pub usage: Usage,
+    /// The message history accumulated during the agent loop.
+    /// Always populated when using `extended_details()`.
+    pub messages: Option<Vec<Message>>,
+}
+
+impl std::fmt::Display for PromptResponse {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.output.fmt(f)
+    }
 }
 
 impl PromptResponse {
@@ -250,7 +261,13 @@ impl PromptResponse {
         Self {
             output: output.into(),
             usage,
+            messages: None,
         }
+    }
+
+    pub fn with_messages(mut self, messages: Vec<Message>) -> Self {
+        self.messages = Some(messages);
+        self
     }
 }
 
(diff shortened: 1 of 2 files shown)


rig/rig-core/src/agent/prompt_request/mod.rs
```diff
@@ -229,43 +234,60 @@ where
...
+            max_tokens: self.max_tokens,
+            additional_params: self.additional_params,
+            tool_server_handle: self.tool_server_handle,
+            dynamic_context: self.dynamic_context,
+            tool_choice: self.tool_choice,
+            state: PhantomData,
+            hook: self.hook,
+            concurrency: self.concurrency,
+            output_schema: self.output_schema,
+            collect_messages: false,
```
joshua-mo-143: This doesn't appear correct, `collect_messages` shouldn't automatically become false here

rig/rig-core/src/agent/prompt_request/mod.rs
```diff
@@ -149,6 +152,7 @@ where
             hook: self.hook,
             concurrency: self.concurrency,
             output_schema: self.output_schema,
+            collect_messages: true,
```
joshua-mo-143: This should not be arbitrarily set to true

rig/rig-core/src/agent/prompt_request/mod.rs
```diff
@@ -88,6 +88,8 @@ where
     concurrency: usize,
     /// Optional JSON Schema for structured output
     output_schema: Option<schemars::Schema>,
+    /// Whether to collect message history in the response
+    collect_messages: bool,
```
joshua-mo-143: Needs a method to set to true/false

rig/rig-core/src/agent/prompt_request/mod.rs
```diff
@@ -229,43 +234,60 @@ where
...
 }
 
 #[derive(Debug, Clone)]
+#[non_exhaustive]
 pub struct PromptResponse {
     pub output: String,
     pub total_usage: Usage,
+    /// The message history accumulated during the agent loop, if collected
+    pub messages: Option<Vec<Message>>,
 }
```
joshua-mo-143, line 251: Can you add a Display impl here that displays the output? Might be more ergonomic than forcing the user to use `foo.output`

joshua-mo-143 (changes requested):
please see comments

illegalcall:
Thanks for the review! I believe with_hook() was actually preserving via self.collect_messages here, but agreed the overall collect_messages design was confusing. 

I've simplified it, removed the bool entirely, Extended path now always populates messages. 
The to_vec() cost is negligible vs LLM latency and this should resolve comments 2 and 3 as well. 
Also added the Display impl. 

Please lmk if this looks good now. Thank you :)

rig/rig-core/src/agent/prompt_request/mod.rs
illegalcall: https://github.com/0xPlaygrounds/rig/pull/1450#issuecomment-3979963796

illegalcall:
Fixed merge conflicts.

Issue #565, opened 2025-07-06
feat: impl `InsertDocuments` for all vector store providers


joshua-mo-143:
- [x] I have looked for existing issues (including closed) about this

## Feature Request

We should try to implement `InsertDocuments` for all vector store providers if it is possible to do so.

### Motivation

Consistency and better DX

### Proposal

Self explanatory.

joshua-mo-143:
Assigning to @ksaritek at user's request to work on the issue.

ksaritek:
if that is ok, i would like to work on that issue

ksaritek:
thank you mate 🙏

ksaritek:
implementations of insertdocuments are missing for below: 

- [x] MongoDB implementation (`rig-mongodb`)
- [ ] SQLite implementation (`rig-sqlite`)
- [ ] Neo4j implementation (`rig-neo4j`) 
- [ ] LanceDB implementation (`rig-lancedb`)

harshsH-28:
Could you please assign this task to me, I want to work on it.

harshsH-28:
Thanks @joshua-mo-143

Sarah-2003:
Hi! I'd like to work on implementing InsertDocuments for rig-lancedb. Is this still available?

gold-silver-copper:
This is done now

PR #1478, opened 2026-03-06
feat(rig-sqlite): impl InsertDocuments trait for sqlite

howenyap:
# Changes
- implements `InsertDocuments` trait for `rig-sqlite`, part of #565 
- updated documentation comments to use `insert_documents`
- updated outdated `top_n` call [here](https://github.com/0xPlaygrounds/rig/compare/main...howenyap:feat/insert-documents-sqlite?expand=1#diff-6934eba90c09ab3a89551a8895c7a373049168458626f725becd31260d2f4bd2L476-L478)
- added integration test to cover `insert_documents`
- updated sqlite connection to be [in memory only](https://www.sqlite.org/uri.html), there are only two tests now but this would make tests less flaky in the future, in case we have leftover rows from previous runs (lmk if i should remove this)

### DIFF
--- a/rig-integrations/rig-sqlite/examples/vector_search_sqlite.rs
+++ b/rig-integrations/rig-sqlite/examples/vector_search_sqlite.rs
@@ -2,15 +2,18 @@ use rig::client::{EmbeddingsClient, ProviderClient};
 use rig::providers::openai;
 use rig::vector_store::request::VectorSearchRequest;
 use rig::{
-    Embed, embeddings::EmbeddingsBuilder, providers::openai::Client, vector_store::VectorStoreIndex,
+    Embed,
+    embeddings::EmbeddingsBuilder,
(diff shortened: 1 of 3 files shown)


howenyap:
hm errors are unrelated to my changes, all pointing towards the [gemini client](https://github.com/0xPlaygrounds/rig/blob/main/rig/rig-core/src/providers/gemini/client.rs#L163-L182)

saw some changes to client in the main branch so i rebased, not sure if it'll fix the errors

howenyap:
verified that previously failing actions now pass locally

howenyap:
bump @joshua-mo-143 ICYMI 🙏

joshua-mo-143:
Hey @howenyap!! Sorry I missed this, I'll be going OOO in a bit but I will review when I am back home

PR #1482, opened 2026-03-06
fix(gemini): correct ProviderBuilder impl for GeminiInteractionsBuilder

adrianncovaci:


joshua-mo-143:
Hey @adrianncovaci huge thanks for this! Didn't realise we accidentally broke main branch

PR #1518, opened 2026-03-15
fix: embedding deserialization breaks with serde_json/arbitrary_precision

EinarasGar:
## Problem

Embedding deserialization fails when `serde_json`'s `arbitrary_precision` feature is enabled anywhere in the dependency tree. The error is:

```
JsonError: data did not match any variant of untagged enum ApiResponse
```

This is commonly triggered by `rust_decimal` with the `serde-with-arbitrary-precision` feature, which is widely used in financial applications.

### Root cause

When `serde_json/arbitrary_precision` is active, **all** JSON numbers are internally represented as strings. The `Vec<f64>` fields in embedding response structs (`EmbeddingValues`, `EmbeddingData`, etc.) fail to deserialize because serde expects native floats but receives string-backed numbers.

The `arbitrary_precision` feature is a compile-time, crate-wide flag — it cannot be opted out of per-deserialization call. Any crate in the dependency tree that enables it (directly or transitively) affects all `serde_json` usage in the binary.

### Reproduction

Any project that combines rig embeddings with `rust_decimal`:

```toml
[dependencies]
rig-core = "0.32"
rust_decimal = { version = "1", features = ["serde-with-arbitrary-precision"] }
```

```rust
let client = gemini::Client::new("key").unwrap();
let model = client.embedding_model_with_ndims("gemini-embedding-001", 768);
let result = model.embed_text("hello").await; // fails with JsonError
```

The same code works in a crate without `rust_decimal` in the dependency tree.

## Fix

Change embedding response structs from `Vec<f64>` to `Vec<serde_json::Number>`, then convert to `f64` via `.as_f64()` when building the `Embedding` struct.

`serde_json::Number` is the canonical number type that works correctly regardless of whether `arbitrary_precision` is enabled — it's the internal representation serde_json uses in both modes.

### Affected providers

All 6 providers with embedding support:
- Gemini (`EmbeddingValues.values`)
- OpenAI (`EmbeddingData.embedding`)
- Mistral (`EmbeddingData.embedding`)
- OpenRouter (`EmbeddingData.embedding`)
- Together (`EmbeddingData.embedding`)
- Cohere (`EmbeddingResponse.embeddings`)

### Change summary

Each provider's embedding response struct: `Vec<f64>` → `Vec<serde_json::Number>`, with `.filter_map(|n| n.as_f64())` at the point where the `Embedding` struct is built. No new dependencies, no API changes.

### DIFF
--- a/rig/rig-core/src/providers/mistral/embedding.rs
+++ b/rig/rig-core/src/providers/mistral/embedding.rs
@@ -99,7 +99,11 @@ where
                         .zip(documents.into_iter())
                         .map(|(embedding, document)| embeddings::Embedding {
                             document,
-                            vec: embedding.embedding.into_iter().filter_map(|n| n.as_f64()).collect(),
+                            vec: embedding
+                                .embedding
+                                .into_iter()
+                                .filter_map(|n| n.as_f64())
+                                .collect(),
                         })
                         .collect())
                 }
--- a/rig/rig-core/src/providers/openai/embedding.rs
+++ b/rig/rig-core/src/providers/openai/embedding.rs
@@ -150,7 +150,11 @@ where
                         .zip(documents.into_iter())
                         .map(|(embedding, document)| embeddings::Embedding {
                             document,
-                            vec: embedding.embedding.into_iter().filter_map(|n| n.as_f64()).collect(),
+                            vec: embedding
+                                .embedding
+                                .into_iter()
+                                .filter_map(|n| n.as_f64())
+                                .collect(),
                         })
                         .collect())
                 }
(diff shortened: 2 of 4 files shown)


joshua-mo-143:
Seems fine to me. Thanks for contributing this PR!

joshua-mo-143:
@EinarasGar Looks like the cargo fmt check failed, if you could run it locally and push I'll get this merged in asap

EinarasGar:
@joshua-mo-143 Thanks for taking a look! Apologies, committed the fmt changes there! Will look out for it the next time

EinarasGar:
@joshua-mo-143 notice build failing after merge to main due to #1519 and this PR changes. Will create a fix PR for this in a bit here

PR #1520, opened 2026-03-15
feat(gemini): add Gemini 3 model constants and thinking_level support

EinarasGar:
## Summary

- Add model constants for Gemini 3 series (`GEMINI_3_FLASH_PREVIEW`, `GEMINI_3_1_FLASH_LITE_PREVIEW`)
- Add `thinking_level: Option<String>` field to `ThinkingConfig` for Gemini 3 models, which use `thinkingLevel` ("minimal", "low", "medium", "high") instead of Gemini 2.5's `thinkingBudget`
- Make `thinking_budget` optional (`Option<u32>`) so users can set either field depending on their model family
- Updated the example file to use new gemini model 

## Context

Gemini 3 models (released March 2026) use a different parameter for controlling reasoning depth. Per the [Gemini 3 Developer Guide](https://ai.google.dev/gemini-api/docs/gemini-3), `thinking_level` and `thinking_budget` are mutually exclusive and cannot be set in the same request.

Both fields use `skip_serializing_if = "Option::is_none"` so existing Gemini 2.5 configurations that set `thinking_budget` via `additional_params` JSON continue to work unchanged.

## Example usage

```rust
// Gemini 3
GenerationConfig {
    thinking_config: Some(ThinkingConfig {
        thinking_budget: None,
        thinking_level: Some("medium".to_string()),
        include_thoughts: Some(true),
    }),
    ..Default::default()
}

// Gemini 2.5 (unchanged)
GenerationConfig {
    thinking_config: Some(ThinkingConfig {
        thinking_budget: Some(2048),
        thinking_level: None,
        include_thoughts: Some(true),
    }),
    ..Default::default()
}
```

## Test plan

Tested with 

```rust
 .additional_params(serde_json::json!({
            "generationConfig": {
                "thinkingConfig": {
                    "thinking_level": "High",
                    "includeThoughts": true
                }
            }
        }))
```

and

```rust
let gen_cfg = GenerationConfig {
    thinking_config: Some(ThinkingConfig {
        thinking_budget: None,
        thinking_level: Some(ThinkingLevel::Medium),
        include_thoughts: Some(true),
    }),
    ..Default::default()
};
let cfg = AdditionalParameters::default().with_config(gen_cfg);
```

### DIFF
--- a/rig/rig-core/src/providers/gemini/completion.rs
+++ b/rig/rig-core/src/providers/gemini/completion.rs
@@ -1518,6 +1518,16 @@ pub mod gemini_api_types {
         }
     }
 
+    /// Thinking depth level for Gemini 3 models.
+    #[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
+    #[serde(rename_all = "snake_case")]
+    pub enum ThinkingLevel {
+        Minimal,
+        Low,
+        Medium,
+        High,
+    }
+
     /// Configuration for the model's thinking/reasoning process.
     /// Note: `thinking_budget` (Gemini 2.5) and `thinking_level` (Gemini 3) are mutually exclusive
     /// and cannot be set in the same request.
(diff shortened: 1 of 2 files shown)


rig/rig-core/src/providers/gemini/completion.rs
```diff
@@ -1513,10 +1517,21 @@ pub mod gemini_api_types {
...
     #[serde(rename_all = "camelCase")]
     pub struct ThinkingConfig {
-        pub thinking_budget: u32,
+        /// Token budget for thinking. Used by Gemini 2.5 models. Range: 0 to 32768.
+        #[serde(skip_serializing_if = "Option::is_none")]
+        pub thinking_budget: Option<u32>,
+        /// Thinking depth level. Used by Gemini 3 models.
+        /// Values: "minimal", "low", "medium", "high".
+        #[serde(skip_serializing_if = "Option::is_none")]
+        pub thinking_level: Option<String>,
```
joshua-mo-143: We probably need to encode this as an enum

EinarasGar: Updated. Added enum. Also updated gemini_streaming example to point to new model and use the thinking config. Updated test plan in the PR description

PR #1528, opened 2026-03-17
fix(openai): make strict field optional in StructuredOutputsInput

### LINKED ISSUE #1512: bug: JsonError: missing field `strict` at line 1 column 1200
The Qwen3.5 model prompts this error, but other models are normal. My version is rig core="0.32"

BillionClaw:
The strict field in StructuredOutputsInput was required during deserialization, causing JsonError when using models like Qwen3.5 via OpenRouter that don't include this field in responses.

Adding #[serde(default)] allows the field to default to false when not present, matching the intended behavior.

Fixes #1512


joshua-mo-143:
Lgtm

PR #1529, opened 2026-03-17
fix(openai): handle encrypted_content field in assistant messages

### LINKED ISSUE #1515: bug: json: unknown field \"encrypted_content\"
version: rig-core = "0.32"

use rig::client::CompletionClient;
    use rig::completion::Prompt;

    // 使用 Client::new 创建客户端
    let client = rig::providers::openai::Client::builder()
        .api_key(&config.api_key)
        .base_url(&config.base_url)
        .build()?;
    // 使用预定义模型常量
    let agent = client
        .agent(&config.model_name)
        .tool(Calculator)
        .preamble("你是一个智能助手。")
        .build();
    let res = agent
        .prompt("介绍下你自己，并使用计算器技能计算 1 + 1")
        .await?;

During the second loop (after the tool is executed and requested again), it may be due to many models not supporting this attribute
my model = doubao-seed-2-0-pro
Is it true that only the client of OpenRoute has special treatment for this relatively universal protocol that requires setting base_url and follows the protocol of OpenAI? (Their version is too new, and other open source vendors still support it.)

Error: CompletionError: HttpError: Invalid status code 400 Bad Request with message: {"error":{"code":"InvalidParameter","message":"The parameter `input` specified in the request are not valid: `json: unknown field \"encrypted_content\"`. Request id: 02177345002536381a9193ce9f89e3456d252c7b9d01d03d4e1e9","param":"input","type":"BadRequest"}}

BillionClaw:
Fixes #1515

## Problem
Some providers (e.g., Doubao/ByteDance) return an "encrypted_content" field in assistant messages that wasn't handled by the OpenAI completion API client, causing a JSON deserialization error.

## Solution
Added an `additional_params` field with `#[serde(flatten)]` to the `Message::Assistant` variant to capture any provider-specific fields, making deserialization flexible and forward-compatible.

## Changes
- Added `additional_params: Option<serde_json::Value>` to `Message::Assistant`
- Updated all code constructing/pattern-matching on `Message::Assistant`
- Added regression tests for `encrypted_content` and unknown field handling

## Testing
- All 441 tests pass
- Clippy and fmt checks pass

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4313,6 +4313,7 @@ version = "0.32.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3f3901269ec6d4f6068d3f09e5f02f995bd076398dcd1dfec407cd230b02d11b"
 dependencies = [
+ "earcutr",
  "float_next_after",
  "geo-types",
  "geographiclib-rs",
@@ -4324,6 +4325,7 @@ dependencies = [
  "rstar 0.12.2",
  "serde",
  "sif-itree",
+ "spade",
 ]
 
 [[package]]
@@ -5454,7 +5456,7 @@ dependencies = [
  "js-sys",
  "log",
  "wasm-bindgen",
- "windows-core 0.57.0",
+ "windows-core",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


onthebed:
Friendly bump on this PR — happy to make changes or close it out if it is no longer useful.

PR #1537, opened 2026-03-19
feat(rig-core): add ToolCallContext for per-call runtime context in tool dispatch

### LINKED ISSUE #1536: feat(rig-core): Per-call runtime context for tool dispatch (ToolCallContext)
### Is your feature request related to a problem?

  Yes. `Tool::call(args)` only receives LLM-provided arguments. There is no mechanism to pass runtime context (auth credentials, session IDs, conversation state) from the caller through the dispatch chain to the tool at call time.

  Without this, callers who need to inject per-request context into tool calls must resort to workarounds: manually mutating the tool arguments JSON before dispatch, using thread-local/task-local storage, or rebuilding agents per request. All of these are
  fragile and couple the orchestration layer to tool-specific concerns.

  ### Describe the feature

  A `ToolCallContext` type-map (modeled after `http::Extensions`) threaded through the entire dispatch chain: `ToolServerHandle` → channel → `ToolServer` → `ToolSet` → `ToolType` → `ToolDyn` → `Tool`.

  Tools opt in by overriding `call_with_context(args, &ctx)` instead of `call(args)`. Tools that don't need context are unchanged — the default delegates to `call`.

  ### Describe the solution you'd like

  - A `ToolCallContext` struct: type-erased, cloneable map keyed by `TypeId`, with `insert<T>()`, `get<T>()`, `get_mut<T>()`, `remove<T>()`, `contains<T>()`
  - `Tool::call_with_context(args, &ctx)` — default method that delegates to `call()`
  - `ToolDyn::call_with_context(args, ctx)` — default method; blanket impl threads to `Tool::call_with_context`
  - `ToolSet::call_with_context()` and `ToolServerHandle::call_tool_with_context()` to thread context through dispatch
  - Fully backward-compatible: no existing `Tool` or `ToolDyn` implementations break
  - WASM-compatible via `cfg(target_family = "wasm")` gating on `Send + Sync` bounds

  ### Describe alternatives you've considered

  1. **Task-local storage** — fragile, invisible coupling, not portable to WASM
  2. **Rebuilding agents per request** — expensive, prevents reuse of pre-built agent configurations
  3. **Mutating tool args JSON** — breaks type safety, couples orchestration to tool internals
  4. **Generic context parameter on `Tool` trait** — would be a breaking change to the trait signature

  ### Additional context

  I have a working implementation ready to PR.

redroy44:
## Summary

  - Adds `ToolCallContext`, a type-erased, cloneable type-map (modeled after `http::Extensions`' `AnyClone` pattern) that lets callers inject per-call runtime values (auth tokens, session IDs, conversation state) into tool execution
  - Threads context through the full dispatch chain: `ToolServerHandle` → channel → `ToolServer` → `ToolSet` → `ToolType` → `ToolDyn` → `Tool`
  - Fully backward-compatible: all existing `Tool` and `ToolDyn` implementations continue to work without changes

  ## New public API

  | Item | Description |
  |---|---|
  | `ToolCallContext` (struct) | Type-map with `new()`, `insert<T>()`, `get<T>()`, `get_mut<T>()`, `remove<T>()`, `contains<T>()`. Re-exported from crate root. |
  | `Tool::call_with_context(args, &ctx)` | Default method — delegates to `call()`. Override to access context. |
  | `ToolDyn::call_with_context(args, ctx)` | Default method — delegates to `call()`. Blanket impl threads to `Tool::call_with_context`. |
  | `ToolSet::call_with_context(name, args, ctx)` | Dispatches to tool with context. |
  | `ToolServerHandle::call_tool_with_context(name, args, ctx)` | Sends context across the channel to the tool server. |

  No existing public API signatures were changed. No breaking changes.

  ## Design

  `ToolCallContext` uses `Option<Box<HashMap<TypeId, Box<dyn AnyClone>>>>` so empty contexts (the common path) have zero allocation. WASM-compatible via `cfg(target_family = "wasm")` gating on `Send + Sync` bounds.

  Delegation direction:
  - Dispatch layers (handle → set → dyn): old method delegates to new `_with_context` method + empty context
  - `Tool` trait: new `call_with_context` default delegates to `call` (ignoring context)

  ## Test plan

  - [x] 12 unit tests for `ToolCallContext` (insert/get/get_mut/remove/clone/contains/debug/empty allocation)
  - [x] 3 integration tests proving context flows end-to-end through `ToolServer` dispatch
  - [x] All 454 pre-existing + new lib tests pass
  - [x] `cargo clippy --lib --tests -- -D warnings` clean
  - [x] `cargo fmt` clean
  
This PR was generated with AI assistance (Claude).

Fixes #1536


rig/rig-core/src/tool/mod.rs
```diff
@@ -340,12 +386,26 @@ impl ToolSet {
...
+
+    /// Call a tool with the given name, arguments, and per-call runtime context.
+    ///
+    /// The context is threaded through to [`Tool::call_with_context`], allowing
+    /// tools to access caller-provided values (auth tokens, session IDs, etc.).
+    pub async fn call_with_context(
+        &self,
+        toolname: &str,
+        args: String,
+        ctx: ToolCallContext,
```
mccormickt: With the changes from #1607, tool calls no longer cross a spawn boundary. We can borrow `ctx` here to avoid per-call cloning.

mccormickt (commented):
Thanks for working on this! This is a much needed change to enable passing conversations to sub-agents and a2a integration.

gold-silver-copper:
So, I believe that while this is a good PR, it's not the proper path forward for rig as it's presented. This is because Rig is planning on abandoning it's current handwritten toolserver, and will replace it completely with the rmcp implementation which already has a toolcallcontext. https://docs.rs/rmcp/latest/rmcp/handler/server/tool/struct.ToolCallContext.html

It does not make sense to merge a port of toolcall context, when it will probably be completely removed in a release or two, and would lock down potentially the wrong minute behavior. 

Will keep it open until the actual rmcp migration lands, please share your thoughts if any.

redroy44:
Hey @gold-silver-copper , thanks for a review!

The primary reason for this PR was integrating a A2A agent as a Rig Tool. `ToolCallContext` would carry A2A `context_id` and `task_id` needed to build A2A message payload + auth headers that vary per-call.

For now, the only way to do it is to inject these extra params into `tool_call.function.arguments` before calling `tool_server_handle.call_tool()` not to expose these to the llm.

If `rmcp` `ToolCallContext` can handle such case then that would be great. Cheers!

gold-silver-copper:
@redroy44 

Rmcp has `rmcp::model::Extensions` and `rmcp::model::Meta` which are very close to what you need. They might not be 100% what you need, but would be best to add the things we need on top of rmcp once the migration is done. Do appreciate the issue and PR, it bumps up the tool rewrite in priority.

A2A is really cool, we definitely want to support it as best as possible

gold-silver-copper:
After some further discussion in #1906 , I am reverting the decision to full migrate Rig tool support to rmcp. Thus this PR is still valid and will be getting a further review soon™

mccormickt:
>The primary reason for this PR was integrating a A2A agent as a Rig Tool. ToolCallContext would carry A2A context_id and task_id needed to build A2A message payload + auth headers that vary per-call.

FWIW, this is the exact reason I started looking into this issue and advocated for this PR.

gold-silver-copper:
Opened #1954, which rebases this work onto current `main` and completes it: it threads `ToolCallContext` through the agent run loop (the piece missing here — `agent.prompt(..).tool_context(ctx)`), propagates it into sub-agents and MCP tools (`_meta`), and folds in the review feedback (hardening, dedup, dead-code removal, tests). The first commit replays this PR verbatim with @redroy44 co-authored, so credit is preserved. Suggest we continue review on #1954 and close this in its favor once that lands.

gold-silver-copper:
This has been implemented in #1954 (as ToolCallExtensions since that is the more accurate rmcp compatible name)

Thanks everyone ❤️

PR #1543, opened 2026-03-22
Add GPT Image 1.5

kevinastock:
I asked claude to add gpt image 1.5 as an option so I could use it in my project, this was all it took.


axiangcoding:
When will the version that merged this PR be released? @gold-silver-copper

gold-silver-copper:
@axiangcoding Later today

PR #1547, opened 2026-03-23
fix(gemini): infer string type for enum schemas in anyOf/oneOf

### LINKED ISSUE #1540: bug: Gemini tools: Schema conversion fails for Option<enum> (anyOf + nested oneOf), emits empty type and triggers 400 Bad Request

## Bug Report
Getting a "400 Bad Request" error while running the code below with Gemini provider. No error is observed when the OpenAI provider is used.
It seems that Gemini provider's type inference code can produce an empty string type, rejected by Gemini. 
The presented code outputs the generated schema for easier inspection.

## Reproduction
Run the below snippet twice: with Gemini and OpenAI providers and observe failure and success respectively:
```sh
 cargo run --release -- -p gemini
 cargo run --release -- -p openai
```

```rust
use clap::{Parser, ValueEnum};
use rig::client::CompletionClient;
use rig::completion::Prompt;
use rig::completion::ToolDefinition;
use rig::providers::{gemini, openai};
use rig::tool::Tool;
use serde::{Deserialize, Serialize};
use std::fmt::{Display, Formatter};

#[derive(Debug, Clone, Copy, ValueEnum)]
enum Provider {
    Gemini,
    Openai,
}

#[derive(Debug, Parser)]
#[command(name = "gemini_rag_bad_request")]
struct Cli {
    #[arg(short = 'p', long = "provider", value_enum)]
    provider: Provider,
}

#[derive(Debug, Serialize, Deserialize, schemars::JsonSchema)]
pub enum FooEnum {
    Foo(String),
}

#[derive(Debug, Serialize, Deserialize, schemars::JsonSchema)]
pub struct BadRequestArgs {
    pub foo_opt: Option<FooEnum>,
}

struct BadRequestTool;

#[derive(Debug)]
struct BadToolError(String);

impl Display for BadToolError {
    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
        write!(f, "BadToolError: {}", self.0)
    }
}

impl std::error::Error for BadToolError {}

impl Tool for BadRequestTool {
    const NAME: &'static str = "bad_request_tool";
    type Error = BadToolError;
    type Args = BadRequestArgs;
    type Output = String;

    async fn definition(&self, _prompt: String) -> ToolDefinition {
        ToolDefinition {
            name: Self::NAME.to_string(),
            description: "tool".to_string(),
            parameters: schemars::schema_for!(Self::Args).into(),
        }
    }

    async fn call(&self, _args: Self::Args) -> Result<Self::Output, Self::Error> {
        Ok("foo".to_string())
    }
}


#[tokio::main]
async fn main() -> anyhow::Result<()> {
    let args = Cli::parse();

    let tool_schema = schemars::schema_for!(BadRequestArgs);
    println!(
        "Generated JSON schema for tool args:\n{}",
        serde_json::to_string_pretty(&tool_schema)?
    );

    match args.provider {
        Provider::Gemini => {
            let api_key = std::env::var("GEMINI_API_KEY")?;
            let llm = gemini::Client::new(api_key)?
                .agent(rig::providers::gemini::completion::GEMINI_2_5_FLASH)
                .preamble("You are a helpful assistant.")
                .tools(vec![Box::new(BadRequestTool)])
                .build();

            let resp = llm.prompt("Say hello. Do not call any tools.").await?;
            println!("Model response:\n{resp}");
        }
        Provider::Openai => {
            let api_key = std::env::var("OPENAI_API_KEY")?;
            let llm = openai::Client::new(api_key)?
                .agent(rig::providers::openai::completion::GPT_4O_MINI)
                .preamble("You are a helpful assistant.")
                .tools(vec![Box::new(BadRequestTool)])
                .build();

            let resp = llm.prompt("Say hello. Do not call any tools.").await?;
            println!("Model response:\n{resp}");
        }
    }

    Ok(())
}
```

## Expected behavior
Running the binary with --openai flag gives an example of an expected model response: "Hello! ..."

## Additional context
The culprit seems to reside in this code (`rig/rig-core/src/providers/gemini/completion.rs`):
```rust
fn infer_type(obj: &serde_json::Map<String, Value>) -> String {
    if let Some(type_val) = obj.get("type")
        && let Some(type_str) = extract_type(type_val)
    {
        return type_str;
    }

    if let Some(any_of) = obj.get("anyOf")
        && let Some(type_str) = extract_type_from_composition(any_of)
    {
        return type_str;
    }

    if let Some(one_of) = obj.get("oneOf")
        && let Some(type_str) = extract_type_from_composition(one_of)
    {
        return type_str;
    }

    if obj.contains_key("properties") {
        "object".to_string()
    } else {
        String::new()
    }
}
```

CMLKevin:
Fixes #1540

## Description

Schema conversion for `Option<enum>` fields (anyOf/oneOf containing objects with an `enum` key) was producing an empty type string, causing a 400 Bad Request from the Gemini API.

The root cause is in `extract_type_from_composition`: when iterating over anyOf/oneOf schemas, it checks for an explicit `type` field and `properties`, but not for the `enum` key. Enum schemas without an explicit type were falling through to `None` and ultimately returning an empty string.

The fix adds an `else if obj.contains_key("enum")` branch that returns `"string"`.

## Changes

- `rig-core/src/providers/gemini/completion.rs`: Added `enum` key check in `extract_type_from_composition` to return `"string"`

## Testing

`cargo test -p rig-core --lib` passes (439 tests).

### DIFF
--- a/rig/rig-core/src/providers/gemini/completion.rs
+++ b/rig/rig-core/src/providers/gemini/completion.rs
@@ -1744,6 +1744,9 @@ pub mod gemini_api_types {
                     obj.get("type").and_then(extract_type).or_else(|| {
                         if obj.contains_key("properties") {
                             Some("object".to_string())
+                        } else if obj.contains_key("enum") {
+                            // Enum schemas without explicit type are string-backed
+                            Some("string".to_string())
                         } else {
                             None
                         }


joshua-mo-143:
thank you! lgtm

PR #1548, opened 2026-03-23
Fix(gemini): Make `prompt_token_count` optional in gemini response

marieaurore123:
Sometimes, gemini doesn't return the field `UsageMetadata.prompt_token_count` or `PartialUsage.prompt_token_count`. Since those structs are expecting this field to be set, this results in a de-serialization error. The completion request should not fail if this field is not returned - set it to 0 if it's not present.

## Implementation
Uses `#[serde(default)]` instead of setting `prompt_token_count` to `Optional` to prevent breaking changes

### DIFF
--- a/rig/rig-core/src/providers/gemini/completion.rs
+++ b/rig/rig-core/src/providers/gemini/completion.rs
@@ -503,7 +503,7 @@ impl TryFrom<GenerateContentResponse> for completion::CompletionResponse<Generat
             .usage_metadata
             .as_ref()
             .map(|usage| completion::Usage {
-                input_tokens: usage.prompt_token_count.unwrap_or(0) as u64,
+                input_tokens: usage.prompt_token_count as u64,
                 output_tokens: usage.candidates_token_count.unwrap_or(0) as u64,
                 total_tokens: usage.total_token_count as u64,
                 cached_input_tokens: 0,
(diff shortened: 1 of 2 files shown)


rig/rig-core/src/providers/gemini/completion.rs
```diff
@@ -1300,6 +1300,7 @@ pub mod gemini_api_types {
     #[derive(Debug, Deserialize, Clone, Default, Serialize)]
     #[serde(rename_all = "camelCase")]
     pub struct UsageMetadata {
+        #[serde(default)]
         pub prompt_token_count: i32,
```
0xMochan, line 1304: This would default to 0, wouldn't this be incorrect? I understand maintaining the types but this would semantically be wrong.

Why not `Option<...>` here?

0xMochan (changes requested):

rig/rig-core/src/providers/gemini/completion.rs
0xMochan: Ok i see copilot suggested the opposite bc of breaking change.

marieaurore123: I decided against Option since it could cause some breaking changes

0xMochan (approved):

0xMochan:
for now we can keep this not a breaking change but reconsider later.

PR #1549, opened 2026-03-23
fix: add Unknown variant to Output enum for web_search_call support

### LINKED ISSUE #1546: bug: Output enum missing web_search_call variant breaks streaming usage tracking
## Description

When using the hosted `web_search` tool (introduced in #1430 / v0.33.0) with the OpenAI Responses API in streaming mode, **token usage is always reported as 0** and web_search-related streaming events are silently dropped.

## Root Cause

The `Output` enum in `rig-core/src/providers/openai/responses_api/mod.rs` only has three variants:

```rust
pub enum Output {
    Message(OutputMessage),
    FunctionCall(OutputFunctionCall),
    Reasoning { ... },
}
```

When OpenAI returns a `response.completed` SSE event, the `output` array contains `web_search_call` items:

```json
{
  "type": "web_search_call",
  "id": "ws_...",
  "status": "completed"
}
```

Since `Output` is tagged with `#[serde(tag = "type")]` and has no variant for `web_search_call`, the **entire `CompletionResponse` (including the `usage` field) fails to deserialize**. In `streaming.rs` (line ~310-314), this falls into the `Err` branch and is silently skipped via `continue`:

```rust
let Ok(data) = data else {
    let err = data.unwrap_err();
    debug!("Couldn't serialize data as StreamingCompletionResponse: {:?}", err);
    continue;
};
```

This means:
1. `final_usage` stays at its default (all zeros)
2. `FinalResponse.aggregated_usage` reports 0 input/output tokens
3. Web search streaming events (`web_search_call.in_progress`, `web_search_call.completed`, etc.) are also silently dropped since `ItemChunkKind` has no variants for them

## Steps to Reproduce

1. Create an agent with `web_search` as a hosted tool via `additional_params`
2. Send a prompt that triggers web search (e.g., "What is the current price of Bitcoin?")
3. Use `stream_prompt().multi_turn()` 
4. Observe that `FinalResponse.usage()` returns all zeros

## Expected Behavior

- `FinalResponse.usage()` should return the actual token counts from OpenAI
- Web search streaming events should be exposed (or at minimum, not break deserialization of `response.completed`)

## Suggested Fix

Add a catch-all variant to `Output` so unknown output types don't break deserialization:

```rust
pub enum Output {
    Message(OutputMessage),
    #[serde(alias = "function_call")]
    FunctionCall(OutputFunctionCall),
    Reasoning { ... },
    #[serde(other)]
    Unknown,
}
```

This would also future-proof against `file_search_call`, `computer_use_call`, `code_interpreter_call`, and any other hosted tool output types.

Separately, adding explicit variants and exposing web search events via the streaming API would be a nice enhancement but is not required to fix the usage tracking regression.

## Environment

- **rig-core**: 0.33.0
- **OpenAI model**: gpt-5.4 (Responses API)
- **Hosted tool**: `web_search` added via `ResponsesToolDefinition::web_search()`

## Related

- #1430 (PR that introduced hosted tool support)

BillionClaw:
The `Output` enum in the OpenAI Responses API module was missing a catch-all variant for unknown output types like `web_search_call`. When OpenAI returns a `response.completed` event containing a `web_search_call` item, serde failed to deserialize the entire `CompletionResponse` (including the `usage` field) because it had no matching variant.

This caused streaming usage tracking to report 0 tokens since the deserialization error was silently caught and skipped.

Added `#[serde(other)] Unknown` variant to the `Output` enum and handled it in both the `From<Output>` conversion and the streaming `ItemChunkKind::OutputItemDone` match arm so unknown output types are skipped without breaking usage tracking or deserialization.

Fixes #1546

### DIFF
--- a/rig/rig-core/src/providers/openai/responses_api/mod.rs
+++ b/rig/rig-core/src/providers/openai/responses_api/mod.rs
@@ -1179,6 +1179,8 @@ pub enum Output {
         #[serde(default)]
         status: Option<ToolStatus>,
     },
+    #[serde(other)]
+    Unknown,
 }
 
 impl From<Output> for Vec<completion::AssistantContent> {
@@ -1221,6 +1223,7 @@ impl From<Output> for Vec<completion::AssistantContent> {
                     },
                 )]
             }
+            Output::Unknown => Vec::new(),
         };
 
         res
--- a/rig/rig-core/src/providers/openai/responses_api/streaming.rs
+++ b/rig/rig-core/src/providers/openai/responses_api/streaming.rs
@@ -357,6 +357,10 @@ where
                                         StreamingItemDoneOutput { item: Output::Message(msg), .. } => {
                                             yield Ok(streaming::RawStreamingChoice::MessageId(msg.id.clone()));
                                         }
+                                        StreamingItemDoneOutput { item: Output::Unknown, .. } => {
+                                            // Unknown output types (e.g. web_search_call, file_search_call)
+                                            // are silently skipped so they don't break usage tracking
+                                        }
                                     }
                                 }
                                 ItemChunkKind::OutputTextDelta(delta) => {


BillionClaw:
Friendly bump -- this PR has been open for a while. Happy to make any changes if needed, or close it if no longer relevant.

lgx0319:
Friendly bump -- this PR has been open for a while. Happy to make any changes if needed, or close it if no longer relevant.

gold-silver-copper:
duplicate of 1552

PR #1550, opened 2026-03-23
fix: handle null values in metadata field for OpenAI Responses API

### LINKED ISSUE #1538: bug: metadata field does not handle null values
The bug is in rig-core-0.33.0 at src/providers/openai/responses_api/mod.rs. The metadata field should be Option<Map> or use #[serde(deserialize_with)] to handle null:                                                                                                                                            
                  
  // Current (broken):                                                                                                                                                                                                                                                                                              
  pub metadata: serde_json::Map<String, serde_json::Value>,                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                    
  // Should be:                                                                                                                                                                                                                                                                                                     
  pub metadata: Option<serde_json::Map<String, serde_json::Value>>,

Currently i have to use the completion API instead of the responses API due to this bug.

BillionClaw:
OpenAI Responses API can return `null` for the `metadata` field, but `AdditionalParameters.metadata` was defined as `Map<String, Value>` which causes a deserialization error.

Changed `metadata` to `Option<Map<String, Value>>` with appropriate serde attributes so it properly handles null, missing, and object values.

Added tests verifying the fix. All existing tests pass.

Fixes #1538

### DIFF
--- a/rig/rig-core/src/providers/openai/responses_api/mod.rs
+++ b/rig/rig-core/src/providers/openai/responses_api/mod.rs
@@ -999,8 +999,8 @@ pub struct AdditionalParameters {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub user: Option<String>,
     /// Any additional metadata you'd like to add. This will additionally be returned by the response.
-    #[serde(skip_serializing_if = "Map::is_empty", default)]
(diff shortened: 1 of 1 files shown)


lgx0319:
Friendly bump -- this PR has been open for a while. Happy to make any changes if needed, or close it if no longer relevant.

PR #1563, opened 2026-03-26
feat: make history generic and immutable

FayCarsons:
This changes prompt generation methods like `prompt` and `stream_prompt` etc to take iterators over references to `Message` as opposed to owned or mutable vectors so that users may decide how to manage chat history and potentially avoid performance pitfalls like cloning large histories.

Because these changes were mostly mechanical I did use Claude Code, but have gone back in and cleaned up any code I felt was low quality. Still, I may have missed some

### DIFF
--- a/rig/rig-core/src/pipeline/mod.rs
+++ b/rig/rig-core/src/pipeline/mod.rs
@@ -280,10 +280,9 @@ impl<E> PipelineBuilder<E> {
 }
 
 #[derive(Debug, thiserror::Error)]
-#[allow(clippy::large_enum_variant)]
 pub enum ChainError {
     #[error("Failed to prompt agent: {0}")]
-    PromptError(#[from] completion::PromptError),
+    PromptError(#[from] Box<completion::PromptError>),
 
     #[error("Failed to lookup documents: {0}")]
     LookupError(#[from] vector_store::VectorStoreError),
--- a/rig/rig-core/src/pipeline/mod.rs
+++ b/rig/rig-core/src/pipeline/mod.rs
@@ -280,6 +280,7 @@ impl<E> PipelineBuilder<E> {
 }
 
 #[derive(Debug, thiserror::Error)]
+#[allow(clippy::large_enum_variant)]
 pub enum ChainError {
     #[error("Failed to prompt agent: {0}")]
     PromptError(#[from] completion::PromptError),
--- a/rig/rig-core/src/providers/anthropic/completion.rs
+++ b/rig/rig-core/src/providers/anthropic/completion.rs
@@ -966,7 +966,21 @@ fn sanitize_schema(schema: &mut serde_json::Value) {
             sanitize_schema(items);
         }
 
-        for key in ["anyOf", "oneOf", "allOf"] {
+        // Anthropic doesn't support oneOf, convert to anyOf
+        if let Some(one_of) = obj.remove("oneOf") {
+            match obj.get_mut("anyOf") {
+                Some(Value::Array(existing)) => {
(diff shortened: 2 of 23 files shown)


rig/rig-core/src/streaming.rs
```diff
@@ -489,25 +489,28 @@ where
     /// ```
     type Hook: PromptHook<M>;
 
-    /// Stream a chat with history to the model
-    ///
-    /// The updated history (including the new prompt and response) is returned
-    /// in [`FinalResponse::history()`](crate::agent::prompt_request::streaming::FinalResponse::history).
```
0xMochan, line 495: I think if this comment needs to be rewritten, patterns for gathering the final chat history should be made clear still.

rig/rig-core/src/providers/anthropic/completion.rs
```diff
@@ -966,7 +966,21 @@ fn sanitize_schema(schema: &mut serde_json::Value) {
             sanitize_schema(items);
         }
 
-        for key in ["anyOf", "oneOf", "allOf"] {
```
0xMochan, line 969: Is this related to this PR?

rig/rig-core/src/pipeline/mod.rs
```diff
@@ -282,7 +282,7 @@ impl<E> PipelineBuilder<E> {
 #[derive(Debug, thiserror::Error)]
 pub enum ChainError {
     #[error("Failed to prompt agent: {0}")]
-    PromptError(#[from] completion::PromptError),
+    PromptError(#[from] Box<completion::PromptError>),
```
0xMochan, line 285: iirc, pipeline might have been deprecated but i'm also surprised to see this change in this PR.

0xMochan (changes requested):
These changes will be really positive moving forward but i do want to make sure this change is well seen. Here, the only way to leverage the produced message history is via the `.extended_details` modifier and adding the extra messages `response.messages` into however you are managing messages.

While this PR is "backwards-compatible" in the type sense, it will result in a breaking change semantically as mutable message history will no longer be extended in the normal completion use-case (though any streaming use-case will benefit since it never had the mutable version). Essentially, when you do a run of agents with tool calling, the intermediate messages won't auto-exist within the message history.

This enables better utilization of history esp. since mutable vectors are painful in several situation (leading to excess clones or it not being available in streaming). I'd also like to ensure bigger projects building on top of rig see this PR as it might motivate some aspects of message history handling. Should allow for easy streaming of messages from a database saving a ton of extra clones.

CC @jamiepine

rig/rig-core/src/pipeline/mod.rs
FayCarsons: I had to touch this error or another somewhere, and I noticed we were ignoring clippys warning about large enum variants so I figured I'd go in and fix that while I was there

0xMochan:
Also CC @illegalcall as your PR (#1450) a bit ago introduced the concept of returning messages which is now the main method to work with message history in this PR

rig/rig-core/src/providers/anthropic/completion.rs
FayCarsons: This was something I needed to test it out, I figured I'd leave it in bc it fixes structured outputs for anthropic

rig/rig-core/src/pipeline/mod.rs
joshua-mo-143: Yeah I think I initially introduced this. The reason we had to do so is because of the lint that we had to fix.

joshua-mo-143 (approved):
lgtm

PR #1567, opened 2026-03-27
feat(rig-bedrock): add OpenTelemetry tracing to completion model

### LINKED ISSUE #1566: rig-bedrock: missing OpenTelemetry tracing instrumentation on chat spans
## Problem

`rig-bedrock` is the only completion model provider with zero tracing instrumentation. All built-in providers (OpenAI, Anthropic, Cohere, Gemini, HuggingFace, Mistral) implement `ProviderResponseExt` and `GetTokenUsage`, and record `gen_ai.*` span attributes on the chat span. Because rig-bedrock doesn't, observability backends (Langfuse, Jaeger, etc.) show:

- `gen_ai.provider.name`: missing
- `gen_ai.request.model`: missing
- `gen_ai.usage.input_tokens` / `output_tokens`: zeros
- `gen_ai.response.id` / `gen_ai.response.model`: missing

Tool spans work fine (rig-core handles those). Only the `chat` GENERATION spans are broken.

## Proposed Solution

- Implement `ProviderResponseExt` and `GetTokenUsage` for `AwsConverseOutput`
- Record `gen_ai.*` span attributes in `CompletionModel::completion()`, matching the pattern used by Anthropic and OpenAI

Note: `gen_ai.response.id` and `gen_ai.response.model` will remain empty — Bedrock's Converse API doesn't return these fields. This is an AWS API limitation.

sachin-punyani:
Fixes #1566

> **AI Disclosure:** This PR was generated with AI assistance (Kiro CLI).

## Problem

`rig-bedrock` is the only completion model provider that has **zero tracing instrumentation**.
All built-in providers (OpenAI, Anthropic, Cohere, Gemini, HuggingFace, Mistral) implement
`ProviderResponseExt` and `GetTokenUsage`, and record `gen_ai.*` span attributes on the chat
span created by rig-core's agent loop. Because rig-bedrock doesn't, observability backends
(Langfuse, Jaeger, etc.) show:

- `gen_ai.provider.name`: missing
- `gen_ai.request.model`: missing
- `gen_ai.usage.input_tokens` / `output_tokens`: zeros
- `gen_ai.response.id` / `gen_ai.response.model`: missing

Tool spans (`execute_tool`) work fine because rig-core handles those directly.
Only the `chat` GENERATION spans are broken.

## Root Cause

rig-core creates chat spans with `gen_ai.*` fields set to `tracing::field::Empty`
(see `rig-core/src/agent/prompt_request/`). It expects the **completion model provider**
to call `span.record(...)` to fill them in — exactly as Anthropic does in
`rig-core/src/providers/anthropic/completion.rs`:

```rust
let span = tracing::Span::current();
span.record_response_metadata(&completion);
span.record_token_usage(&completion.usage);
```

`rig-bedrock`'s `CompletionModel::completion()` never touches the span.

### 1. `rig-integrations/rig-bedrock/src/types/assistant_content.rs`

Implement `ProviderResponseExt` and `GetTokenUsage` for `AwsConverseOutput`.

### 2. `rig-integrations/rig-bedrock/src/completion.rs`

Record span attributes in `completion()`, matching the Anthropic pattern:
- Create `info_span!` with all `gen_ai.*` fields
- Wrap async work in `.instrument(span)`
- Call `span.record_response_metadata()` and `span.record_token_usage()` after response

No `Cargo.toml` change needed — `tracing` is already a dependency.

## What This Fixes

| Field | Before | After |
|-------|--------|-------|
| `gen_ai.provider.name` | _(missing)_ | `aws_bedrock` |
| `gen_ai.request.model` | _(missing)_ | e.g. `anthropic.claude-3-5-sonnet-20241022-v2:0` |
| `gen_ai.system_instructions` | _(missing)_ | preamble text |
| `gen_ai.usage.input_tokens` | 0 | actual token count |
| `gen_ai.usage.output_tokens` | 0 | actual token count |
| `gen_ai.usage.cached_tokens` | 0 | cache_read + cache_write tokens |

Note: `gen_ai.response.id` and `gen_ai.response.model` remain empty — Bedrock's Converse API doesn't return these. This is an AWS API limitation.

## Tests

7 new unit tests (60 total, all passing, zero clippy warnings):

- `provider_response_ext_get_text_response`
- `provider_response_ext_response_id_is_none`
- `provider_response_ext_get_usage_with_tokens`
- `provider_response_ext_get_usage_none_when_missing`
- `provider_response_ext_output_messages_serializable`
- `get_token_usage_delegates_to_provider_response_ext`
- `get_token_usage_none_when_no_usage`

## Scope

- **Non-streaming completion only.** Streaming tracing is not implemented by any external integration and is a separate concern.
- **First external integration with full tracing.** rig-gemini-grpc implements `ProviderResponseExt` but never records spans. rig-vertexai has no tracing at all.

## Reference

Pattern follows:
- `rig-core/src/providers/anthropic/completion.rs`
- `rig-core/src/providers/openai/completion/mod.rs`
- `rig-core/src/telemetry/mod.rs`


joshua-mo-143:
> Non-streaming completion only. Streaming tracing is not implemented by any external integration and is a separate concern.

We do want to stream tracing for all providers, but haven't been able to get the chance to do so yet.

(btw `cargo fmt` needs to be run)

sachin-punyani:
Executed cargo fmt and pushed the fix. 

Regarding streaming tracing — I can take that as a follow up - in a new PR

gold-silver-copper:
I pulled this branch locally and made a follow-up pass on the Bedrock tracing changes.

What I changed:
- Made `request_model` drive the actual Bedrock request in both non-streaming and streaming paths, so telemetry matches the model that is actually invoked.
- Aligned the fallback completion span with the fields `record_token_usage()` writes by using:
  - `gen_ai.usage.cache_read.input_tokens`
  - `gen_ai.usage.cache_creation.input_tokens`
- Unified Bedrock usage normalization behind one helper so `ProviderResponseExt` / `GetTokenUsage` and `CompletionResponse.usage` no longer interpret cache write tokens differently.
- Updated `get_text_response()` to join all Bedrock text blocks instead of returning only the first one, which is consistent with the multi-block handling used by providers like Anthropic.

Why:
- Before this, Bedrock telemetry could report a request model that Bedrock never actually used.
- Cache usage could disagree between telemetry and the returned `CompletionResponse`.
- The fallback span declared cache fields that were different from the ones actually recorded.
- Multi-block text responses were truncated in the telemetry adapter.

rig-integrations/rig-bedrock/src/types/assistant_content.rs
```diff
@@ -204,9 +258,89 @@ mod tests {
     use aws_sdk_bedrockruntime::types as aws_bedrock;
     use rig::{
         OneOrMany, completion,
+        completion::GetTokenUsage,
         message::{AssistantContent, ReasoningContent},
+        telemetry::ProviderResponseExt,
     };
 
+    /// Helper: build an AwsConverseOutput with text content and optional usage.
+    fn make_output(text: &str, usage: Option<aws_bedrock::TokenUsage>) -> AwsConverseOutput {
```
0xMochan, line 267: Is there a reason why this isn't an `impl` on the `AwsConverseOutput`

gold-silver-copper: The reason is that it's a text fixture, not a public API. make_output creates a dummy message, builds it, and converts it. It also hard codes  the message role and type, the builder stop reason, and does some unwrapping, none of which are useful for a production `impl` 

In theory we could do something like 
```rust
#[cfg(test)]
impl AwsConverseOutput {
    fn from_text_and_usage(...) -> Self { ... }
}
```

But I think that's too noisy here, just imo

PR #1568, opened 2026-03-27
Fix VoyageAI Usage deserialization failure on missing prompt_tokens

### LINKED ISSUE #1564: bug: VoyageAI embeddings fail with JsonError: Usage.prompt_tokens not returned by API

## Bug Report
rig-core VoyageAI embeddings fail with deserialization error because `Usage.prompt_tokens` is a required field, but the VoyageAI API never returns it — it only returns `total_tokens`.

Error at runtime:
`JsonError: data did not match any variant of untagged enum ApiResponse`

This affects any use of the VoyageAI embeddings provider. The official repo example would also fail.

## Reproduction
```rust
use rig::providers::voyageai;

  #[tokio::main]
  async fn main() {
      let client = voyageai::Client::new("<VOYAGE_API_KEY>");
      let model = client.embedding_model(voyageai::VOYAGE_3);

      // This panics / errors with JsonError about untagged enum ApiResponse
      let _ = model.embed_text("hello world").await.unwrap();
  }
```

## Expected behavior
The embedding request should succeed. prompt_tokens should be treated as optional (defaulting to 0) since the VoyageAI API response only includes total_tokens, not prompt_tokens.

## Screenshots
N/A

## Additional context
 Root cause: in rig-core/src/providers/voyageai.rs, the Usage struct mirrors the OpenAI pattern but prompt_tokens is not part of VoyageAI's actual API response:

 ```rust
  // Current (broken)
  struct Usage {
      total_tokens: usize,
      prompt_tokens: usize,  // VoyageAI never returns this field
  }

```
One-line fix — add `#[serde(default)]`:

 ```rust
   // Fixed
  struct Usage {
      total_tokens: usize,
      #[serde(default)]
      prompt_tokens: usize,
  }
```
  This was discovered while building against rig-core with a real VoyageAI API key. The field appears to have been
  copied from the OpenAI provider without verifying against the actual Voyage API response schema.

joaquinhuigomez:
Fixes #1564

VoyageAI's embedding API doesn't return `prompt_tokens` in the usage response, but the `Usage` struct requires it during deserialization. This causes every embedding call to fail with a serde error.

Added `#[serde(default)]` to `prompt_tokens` so it falls back to 0 when the field is absent, matching how `total_tokens` is already handled.

### DIFF
--- a/rig/rig-core/src/providers/voyageai.rs
+++ b/rig/rig-core/src/providers/voyageai.rs
@@ -136,8 +136,6 @@ pub struct EmbeddingResponse {
 
 #[derive(Clone, Debug, Deserialize)]
 pub struct Usage {
-    #[serde(default)]
-    pub prompt_tokens: usize,
     pub total_tokens: usize,
 }
 


joshua-mo-143:
Hey @joaquinhuigomez, feel free to totally delete the field. Looks like we made an oversight while creating the provider integration.

joaquinhuigomez:
Done — removed the field entirely. Pushed to the branch.

PR #1570, opened 2026-03-28
fix(rig-derive): propagate function visibility to generated tool structs

tomasz-feliksik:
Summary                                                                                                                                                                                          
                                                                                                                                                                                                   
  - #[rig_tool] hardcodes pub(crate) on all generated structs and the static, regardless of the annotated function's visibility                                                                      
  - This prevents library crates from exporting macro-generated tools to downstream consumers — a pub function produces a pub(crate) struct that can't cross the crate boundary
  - Fix: capture the function's vis token and apply it to the generated struct, its fields, and the static                                                                                           
                                                                                                                                                                                                     
  Changes                                                                                                                                                                                          
                                                                                                                                                                                                     
  3 lines changed in rig-derive/src/lib.rs:                                                                                                                                                        
  - Capture let vis = &input_fn.vis;                                                                                                                                                               
  - Replace pub(crate) with #vis on the tool struct, params struct (+ fields), and static

  Added rig-derive/tests/visibility.rs verifying that pub and private tools work correctly across module boundaries.                                                                                 
   
  Test plan                                                                                                                                                                                          
                                                                                                                                                                                                   
  - Existing calculator test still passes (backward compatible)                                                                                                                                    
  - New visibility test: pub function produces struct accessible outside its module
  - New visibility test: private function produces struct only accessible within its module                                                                                                          
  - End-to-end: used in a real workspace where common crate exports pub tools to binary crates

### DIFF
--- a/rig/rig-derive/examples/rig_tool/simple.rs
+++ b/rig/rig-derive/examples/rig_tool/simple.rs
@@ -57,6 +57,12 @@ async fn main() {
         .preamble("You are an agent with tools access, always use the tools")
         .max_tokens(1024)
         .tool(Add)
+        .tool(Subtract)
+        .tool(Multiply)
+        .tool(Divide)
+        .tool(AnswerSecretQuestion)
+        .tool(HowManyRs)
+        .tool(SumNumbers)
         .build();
 
     for prompt in [


tomasz-feliksik:
That was quick!
Thank you @joshua-mo-143 :)

PR #1579, opened 2026-03-31
fix: spawn GetToolDefs to avoid blocking actor loop

### LINKED ISSUE #1573: bug: `ToolServer` actor loop blocks under concurrent tool execution and dynamic tool lookups

## Bug Report

The `ToolServer` actor in `src/tool/server.rs` has two concurrency issues in `handle_message` that cause the actor loop to block under concurrent load. Both stem from async work being awaited inside the loop without being offloaded to a separate task.

**Issue 1**: In the `CallTool` branch, a read lock on the toolset is acquired and held across the entire `tool.call(...).await`. Any concurrent `AddTool` or `RemoveTool` that requires a write lock will block — and because `RwLock` write-lock requests block new readers, all concurrent `CallTool` operations stall for the full duration of the slow tool.

**Issue 2**: The `GetToolDefs` branch awaits `get_tool_definitions` inline inside the actor loop. This function performs network I/O against one or more vector stores. While it awaits, the actor cannot process any other message — including `CallTool` requests from other agents.

## Reproduction

```rust
// Issue 1: spawn N slow tool calls, then request `AddTool` concurrently.
// `AddTool` will stall until all tool calls complete.
let handle = ToolServer::new().tool(SlowTool).run();
let calls: Vec<_> = (0..5).map(|_| handle.call_tool("slow", "{}")).collect();
let add = handle.add_tool(NewTool); // blocked until all calls finish
futures::future::join_all(calls).await;
add.await.unwrap();
```

```rust
// Issue 2: call `get_tool_defs` with dynamic tools while also calling a tool.
// The `call_tool` response is delayed by the vector store round-trip.
let defs = handle.get_tool_defs(Some("query".into())); // freezes actor
let call = handle.call_tool("fast_tool", "{}");        // queued behind defs
futures::future::join(defs, call).await;
```

## Expected behavior

The actor loop should remain responsive at all times. Concurrent `AddTool`, `CallTool`, and `GetToolDefs` messages should not block each other.

## Additional context

**Possible fixes for Issue 1**: 

**Minimal**: change `ToolSet` storage from `Box<dyn ToolDyn>` to `Arc<dyn ToolDyn>`. Clone the `Arc` before dropping the read lock, then await outside it. One-line storage change, no API surface impact.

**Structural**: drop the actor pattern for read-heavy paths — pass `Arc<RwLock<ToolSet>>` directly to callers and let them manage locking. Eliminates the bottleneck entirely but is a larger refactor.

**Possible fixes for Issue 2**:  

**Minimal**: wrap `dynamic_tools` in `Arc<Vec<...>>` and `tokio::spawn` the `GetToolDefs` handler, sending the result back via the existing `oneshot` channel.

**Structural**: introduce a dedicated channel and task for `GetToolDefs` messages, keeping the main actor loop free of any I/O.

Happy to submit a PR if this makes sense. A few questions before doing that:

- Is the actor pattern worth preserving or is a simpler shared-state approach acceptable?

- Is changing `ToolSet` storage to `Arc` acceptable or should tool ownership stay with the server?

joaquinhuigomez:
Fixes #1573 (Issue 2)

`GetToolDefs` with a prompt awaits vector store I/O inline in the actor loop, blocking all concurrent messages (`CallTool`, `AddTool`, etc.) for the duration of the round-trip. `CallTool` already avoids this by spawning to a separate task.

**Changes**

- Wrap `dynamic_tools` in `Arc` so it can be shared with spawned tasks
- Spawn `GetToolDefs` handler to a separate task, matching the existing `CallTool` pattern
- Extract `get_tool_definitions` into a free function that takes owned/shared references instead of `&mut self`
- Handle `ToolError` response variant in `ToolServerHandle::get_tool_defs`

**Test plan**

- All 4 existing `tool::server` tests pass (including `test_toolserver_dynamic_tools`)
- No API surface changes — `dynamic_tools` field is private, `Arc` wrapping is internal

### DIFF
--- a/rig/rig-core/src/tool/server.rs
+++ b/rig/rig-core/src/tool/server.rs
@@ -18,7 +18,8 @@ pub struct ToolServer {
     /// These tools will always exist on the tool server for as long as they are not deleted.
     static_tool_names: Vec<String>,
     /// Dynamic tools. These tools will be dynamically fetched from a given vector store.
-    dynamic_tools: Vec<(usize, Box<dyn VectorStoreIndexDyn + Send + Sync>)>,
+    /// Wrapped in Arc to allow sharing with spawned tasks without blocking the actor loop.
+    dynamic_tools: Arc<Vec<(usize, Box<dyn VectorStoreIndexDyn + Send + Sync>)>>,
     /// The toolset where tools are called (to be executed).
     /// Wrapped in Arc<RwLock<...>> to allow concurrent tool execution.
     toolset: Arc<RwLock<ToolSet>>,
@@ -34,7 +35,7 @@ impl ToolServer {
     pub fn new() -> Self {
         Self {
             static_tool_names: Vec::new(),
-            dynamic_tools: Vec::new(),
+            dynamic_tools: Arc::new(Vec::new()),
             toolset: Arc::new(RwLock::new(ToolSet::default())),
         }
     }
@@ -53,7 +54,7 @@ impl ToolServer {
         mut self,
         dyn_tools: Vec<(usize, Box<dyn VectorStoreIndexDyn + Send + Sync>)>,
     ) -> Self {
-        self.dynamic_tools = dyn_tools;
+        self.dynamic_tools = Arc::new(dyn_tools);
         self
     }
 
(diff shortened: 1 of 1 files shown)


gold-silver-copper:
@joaquinhuigomez Hey there, please run `cargo fmt` to make all checks pass, thank you 🙏

gold-silver-copper:
Replaced by https://github.com/0xPlaygrounds/rig/pull/1607

PR #1582, opened 2026-04-01
fix(openai): capture ResponseFailed errors in stream mode

### LINKED ISSUE #1562: Errors are not captured when using OpenAI in stream mode.
```toml
rig-core = { version = "0.33.0", features = ["derive"] }
```

When using OpenAI agent, if the API returns an error (in my case I was pushing a PDF with too much content in it) Rig doesn't display it.
This only happens with OpenAI(GPT..). With Anthropic(Claude) seems to be ok and error is returned.

OpenAI code:
```rs
let client: openai::Client = match openai::Client::new(&openai_api_key) {
    Ok(client) => client,
    Err(err) => {
        tracing::error!("{:?}", err);
        return;
    }
};

let agent = client
    .agent(body.model.to_string())
    .name("default")
    .preamble(&ag_preamble)
    .build();

let mut stream = agent.stream_chat(ag_chat_prompt, ag_chat_history).await;
while let Some(chunk) = stream.next().await {
    dbg!(&chunk);
}
```

OpenAI code output:
```
&chunk = Ok(
    FinalResponse(
        FinalResponse {
            response: "",
            aggregated_usage: Usage {
                input_tokens: 0,
                output_tokens: 0,
                total_tokens: 0,
                cached_input_tokens: 0,
            },
            history: Some(
                [
                    User {
                        content: OneOrMany {
                            first: Document(
                                Document {
                                    data: Base64(
                                        "..."
                                    ),
                                    media_type: Some(
                                        PDF,
                                    ),
                                    additional_params: None,
                                },
                            ),
                            rest: [
                                Text(
                                    Text {
                                        text: "explain file",
                                    },
                                ),
                            ],
                        },
                    },
                ],
            ),
        },
    ),
)
```

Anthropic code:
```rs
let client: anthropic::Client = match anthropic::Client::new(&anthropic_api_key) {
    Ok(client) => client,
    Err(err) => {
        tracing::error!("{:?}", err);
        return;
    }
};

let agent = client
    .agent(body.model.to_string())
    .name("default")
    .preamble(&ag_preamble)
    .build();

let mut stream = agent.stream_chat(ag_chat_prompt, ag_chat_history).await;
while let Some(chunk) = stream.next().await {
    dbg!(&chunk);
}
```

Anthropic code output:
```
&chunk = Err(
    Completion(
        ProviderError(
            "SSE Error: Invalid status code 429 Too Many Requests with message: {\"type\":\"error\",\"error\":{\"type\":\"rate_limit_error\",\"message\":\"This request would exceed your organization's rate limit of 800,000 input tokens per minute (org: f6a4ab2a-6b3c-4123-99de-8d3a12200c36, model: claude-sonnet-4-6). For details, refer to: https://docs.claude.com/en/api/rate-limits. You can see the response headers for current usage. Please reduce the prompt length or the maximum tokens requested, or try again later. You may also contact sales at https://claude.com/contact-sales to discuss your options for a rate limit increase.\"},\"request_id\":\"req_011CZRkNdhbmcLyGWXkztMsF\"}",
        ),
    ),
)
```

If I do the request to OpenAI using postman I get error from them as well:
```
event: error
data: {"type":"error","error":{"type":"invalid_request_error","code":"context_length_exceeded","message":"Your input exceeds the context window of this model. Please adjust your input and try again.","param":"input"},"sequence_number":2}
```

So both providers return an error (which is ok) but Rig only catches the error from Anthropic. For OpenAI no error is returned, which is the bug I'm reportinig.

gabrielrondon:
Fixes #1562

When OpenAI returns an error during streaming (e.g. `context_length_exceeded`), the response chunk arrives as `ResponseFailed` but was silently ignored by the `else` branch, resulting in an empty `FinalResponse` with zero tokens.

This change replaces the `if-let` with an explicit `match` on `ResponseChunkKind`:

- **`ResponseCompleted`**: existing behavior (record span, extract usage)
- **`ResponseFailed | ResponseIncomplete`**: extract `response.error` (code + message) and yield `CompletionError::ProviderError`, consistent with how SSE errors and the WebSocket implementation handle failures
- **Other variants**: continue (no change)

### DIFF
--- a/rig/rig-core/src/providers/openai/responses_api/streaming.rs
+++ b/rig/rig-core/src/providers/openai/responses_api/streaming.rs
@@ -429,15 +429,9 @@ where
                                         final_usage = usage;
                                     }
                                 }
-                                ResponseChunkKind::ResponseFailed => {
+                                ResponseChunkKind::ResponseFailed | ResponseChunkKind::ResponseIncomplete => {
                                     let error_message = response_chunk_error_message(&kind, &response)
-                                        .expect("failed response should have an error message");
-                                    yield Err(CompletionError::ProviderError(error_message));
-                                    break;
-                                }
-                                ResponseChunkKind::ResponseIncomplete => {
-                                    let error_message = response_chunk_error_message(&kind, &response)
-                                        .expect("incomplete response should have an error message");
+                                        .expect("terminal response should have an error message");
                                     yield Err(CompletionError::ProviderError(error_message));
                                     break;
                                 }


gold-silver-copper:
I pushed a follow-up to keep the streaming path closer to the existing websocket implementation and to add regression coverage for the original bug in #1562.

The core behavior change is the same: when the OpenAI Responses API ends a stream with `response.failed` or `response.incomplete`, we now surface that as `CompletionError::ProviderError(...)` instead of falling through and returning an empty `FinalResponse` with zero tokens.

I also refactored the streaming-side error formatting into small local helpers so the architecture matches websocket more closely:
- empty `error.code` returns just the provider message
- non-empty `error.code` returns `{code}: {message}`
- `response.incomplete` uses `incomplete_details.reason`

I also removed the extra `tracing::error!` here, since these are provider-declared terminal outcomes rather than transport failures, and the websocket path already returns `ProviderError` without separately logging them.

Finally, I added streaming regression tests using `MockStreamingClient` against the real `ResponsesCompletionModel::stream()` path. The new tests cover:
- `response.failed` with an empty error code
- `response.failed` with a non-empty code
- `response.incomplete` with `incomplete_details.reason`
- `response.completed` still populating final usage after the refactor

I kept the scope local. Preserving structured provider errors (`code` / `message`) on `CompletionError` would be a broader API change, so this PR still follows the existing `CompletionError::ProviderError(String)` contract while fixing the original stream-mode bug.

Issue #1583, opened 2026-04-01
feat: rig-bedrock needs support for prompt caching


marcbrooker:
- [X ] I have looked for existing issues (including closed) about this

## Feature Request
Support for prompt caching and CachePointBlock in rig-bedrock. 

### Motivation
Prompt caching reduces costs and improves performance, especially for agentic workloads.

### Proposal
Add backward-compatible single boolean option to enable prompt caching of system prompt and each agent interaction.

### Alternatives
A more complex approach could expose fine-grained control over caching, but would be hard to do with backward compatibility with the existing API.

PR #1584, opened 2026-04-01
Add support for prompt caching in rig-bedrock

marcbrooker:
Implement simple support for prompt caching in rig-bedrock.

Fixes: https://github.com/0xPlaygrounds/rig/issues/1583

### DIFF
--- a/rig-integrations/rig-bedrock/src/completion.rs
+++ b/rig-integrations/rig-bedrock/src/completion.rs
@@ -178,14 +178,19 @@ impl CompletionModel {
 
     /// Enable Bedrock prompt caching for this model.
     ///
-    /// When enabled, `CachePoint` blocks are inserted into the system prompt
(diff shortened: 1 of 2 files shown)


gold-silver-copper:
I pushed a follow-up to improve test coverage and clarify the current scope of the prompt caching support.

Specifically, I added direct regression tests for the new `prompt_caching = true` path to verify that `CachePoint` blocks are inserted after the serialized `system` content and after the final message. I also updated the streaming usage mapping so Bedrock cache read/write token counts are reflected consistently in streamed responses, matching the existing non-streaming behavior.

I made a small cleanup pass as well by extracting a shared cache-point helper, reducing repeated test fixture boilerplate, and updating the `with_prompt_caching()` documentation.

This PR remains intentionally scoped to explicit cache checkpoints for `system` and `messages`. Support for tool cache checkpoints on Bedrock models that allow them is left as a follow-up.

PR #1586, opened 2026-04-01
feat(rig-core): add parallel tool execution to streaming multi-turn agents

sonicxml:
## Summary

The non-streaming `PromptRequest` already supports concurrent tool execution via `with_tool_concurrency(n)` and `buffer_unordered`. The streaming path (`StreamingPromptRequest`) was missing this — tool calls were always executed sequentially as they arrived in the stream, blocking on each one before processing the next.

This adds `with_tool_concurrency(n)` to `StreamingPromptRequest`, matching the existing non-streaming API.

### How it works

When `concurrency > 1`, tool calls from a single model response are:
1. **Buffered** during streaming (ToolCall events are still yielded immediately to consumers)
2. **Approved** via pre-execution hooks (`on_tool_call`) sequentially — a `Terminate` prevents all execution, a `Skip` only affects that tool
3. **Dispatched concurrently** via `futures::future::join_all`
4. **Post-processed** via `on_tool_result` hooks sequentially — a `Terminate` stops the agent loop

When `concurrency == 1` (default), the behaviour is **completely unchanged** — the existing inline sequential execution path is preserved.

### Usage

```rust
let stream = agent
    .stream_chat("Call both tools", history)
    .multi_turn(5)
    .with_tool_concurrency(10) // new
    .await;
```

### Design decisions

- **Opt-in only**: Default concurrency is 1 to avoid surprising existing users
- **API consistency**: Method name and semantics match the existing `PromptRequest::with_tool_concurrency`
- **Hook ordering preserved**: Pre-execution hooks run sequentially before dispatch (Terminate prevents everything). Post-execution hooks run sequentially after all tools complete
- **Sequential path untouched**: The `concurrency == 1` code path is the original code, unmodified

## Test plan

- [x] All existing `tool::server` unit tests pass
- [x] `cargo clippy` clean
- [x] New integration tests (behind `#[ignore]`) verify:
  - Parallel execution of 2 slow tools (500ms each) completes in ~500ms, not ~1000ms
  - Sequential execution (default) maintains expected ~1000ms timing
  - Tests for OpenAI and Anthropic providers

### DIFF
--- a/rig/rig-core/src/agent/prompt_request/streaming.rs
+++ b/rig/rig-core/src/agent/prompt_request/streaming.rs
@@ -224,6 +224,8 @@ where
     output_schema: Option<schemars::Schema>,
     /// Optional per-request hook for events
     hook: Option<P>,
+    /// How many tools should be executed at the same time (1 by default).
+    concurrency: usize,
 }
 
 impl<M, P> StreamingPromptRequest<M, P>
@@ -251,6 +253,7 @@ where
             tool_choice: agent.tool_choice.clone(),
             output_schema: agent.output_schema.clone(),
             hook: None,
+            concurrency: 1,
         }
     }
 
@@ -278,6 +281,7 @@ where
             tool_choice: agent.tool_choice.clone(),
             output_schema: agent.output_schema.clone(),
             hook: agent.hook.clone(),
+            concurrency: 1,
         }
     }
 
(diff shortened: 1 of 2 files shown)


gold-silver-copper:
Hey there, thanks for working on this. I did some reviewing of this and it's mostly good but there's a few things that I came across which made me realize I will need to do a much deeper dive into concurrent tool calling before I can give a proper review. So for now I'm gonna say this probably won't make it into the next release, but I will definitely revisit when I have more time. 

Thank you for your patience 🙏

tidely:
Hey, is there a way I could help get this merged? @sonicxml do you have time to resolve the merge conflicts? I'd be happy to take over if you're busy

gold-silver-copper:
Parallel tool calls for streaming  have now been implemented in #1957

PR #1587, opened 2026-04-01
Add ModelLister for Ollama, Anthropic, Mistral, OpenAI, Gemini

LHelge:
# Summary
Continued on the work from #1075 which introduced the primitives to access the API for listing models. I have implemented the `ModelLister` trait for the providers that I currently use in my project.
- Ollama
- Anthropic
- Mistral
- OpenAI

## Usage
For the providers that has this implementation the `list_models()` function should be available on the `Client`
```rust
#[tokio::main]
    let client = anthropic::Client::new("sk-ant-...").unwrap();
    let models = client.list_models().await?;
    println!("Available models: {models:#?}");

    Ok(())
}
```

## Future work
Should be fairly straight forward to copy/paste or vibe the same solution for the remaining providers.


gold-silver-copper:
Added gemini support as well as tests.

Issue #1415, opened 2026-02-21
bug: Type annotation error during Client::builder() is a DX papercut


joshua-mo-143:
- [x] I have looked for existing issues (including closed) about this

## Bug Report

Using the `Client::builder()` method results in the user being required to specify a type for the HTTP client. This is not really something we should need the user to do.

## Reproduction

```rust
 let r = http_client::ReqwestClient::default();
    let client = openai::Client::builder()
        .base_url("127.0.0.1:11435")
        .api_key("none")
        .http_headers(Default::default())
        .http_client(r)
        .build()
        .unwrap();
```

Output:

```rust
error[E0283]: type annotations needed
   --> src\main.rs:37:18
    |
 37 |     let client = openai::Client::builder()
    |                  ^^^^^^^^^^^^^^^^^^^^^^^^^ cannot infer type for type parameter `H`
    |
    = note: cannot satisfy `_: std::default::Default`
note: required by a bound in `rig::client::Client::<Ext, H>::builder`
   --> C:\Users\~~~\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\rig-core-0.31.0\src\client\mod.rs:341:8
    |
341 |     H: Default + HttpClientExt,
    |        ^^^^^^^ required by this bound in `Client::<Ext, H>::builder`
...
345 |     pub fn builder() -> ClientBuilder<Builder, NeedsApiKey, H> {
    |            ------- required by a bound in this associated function
```

Workaround:

```rust
    use rig::http_client::ReqwestClient;
    // Create OpenAI client
    let r = ReqwestClient::default();
    let client = openai::Client::<ReqwestClient>::builder()
        .base_url("127.0.0.1:11435")
        .api_key("none")
        .http_headers(Default::default())
        .http_client(r)
        .build()
        .unwrap();
```

## Expected behavior

It should compile without requiring the type annotation.

PR #1615, opened 2026-04-10
feat(rig-core): add ChatGPT Subscription, GitHub Copilot, and compatibility providers

wey-gu:
This PR adds two subscription-backed providers and three compatibility providers, while keeping provider-specific behavior localized in `rig-core` instead of pushing quirks into callers.

What is included

- `ChatGPT Subscription`
  - OAuth device flow
  - token refresh + local cache
  - `/responses` request normalization for the subscription backend
  - compatibility handling for current backend quirks observed in live testing:
    - `instructions` required
    - `system` messages must be lifted out of `input`
    - SSE responses may omit `content-type`
    - device-code `interval` may be returned as a string

- `GitHub Copilot`
  - OAuth device flow
  - Copilot API-key bootstrap / refresh
  - per-account API base routing from the token bootstrap response
  - `/responses` routing for codex-class models, while keeping chat-completions for normal chat models

- compatibility providers for documented OpenAI-compatible / Anthropic-compatible APIs
  - `MiniMax`
  - `Moonshot`
  - `Z.AI`

Why the shared plumbing changed

- OpenAI-compatible model/client types were generalized so new providers can reuse existing request / response / streaming machinery instead of copying entire providers.
- Anthropic-compatible builder/client plumbing was similarly generalized so providers such as MiniMax / Moonshot / Z.AI can expose documented Anthropic-compatible endpoints with small provider modules.
- Provider-specific normalization remains inside the provider modules, not in downstream apps.

Validation

- `cargo check -p rig-core`
- targeted provider tests for ChatGPT, GitHub Copilot, MiniMax, Moonshot, and Z.AI
- live end-to-end validation through Con for:
  - ChatGPT Subscription OAuth + streaming completions
  - GitHub Copilot OAuth + streaming completions
- live manual validation for MiniMax China endpoints:
  - OpenAI-compatible mode
  - Anthropic-compatible mode

Reference alignment

- ChatGPT Subscription and Copilot behavior was aligned against live backend behavior and LiteLLM’s current subscription-provider behavior.
- MiniMax / Moonshot / Z.AI were checked against local LiteLLM and AI SDK references.
- The only material gap found in audit was Moonshot-specific request shaping for `tool_choice=required` and assistant `reasoning_content`; that is handled in this PR.

Scope note

This is intentionally one cohesive provider/compatibility change because the subscription providers required small shared compatibility extensions. If maintainers prefer, I can split follow-up work into:
1. shared OpenAI / Anthropic compatibility refactors
2. MiniMax / Moonshot / Z.AI providers
3. GitHub Copilot
4. ChatGPT Subscription


wey-gu:
btw i am building a rust desktop app and verified above providers manually :)

RileyMathews:
I was also working on trying to make a custom provider using OAuth codex login. So loved seeing this PR already open! This would be amazing to see added here. I’m also working on some agent tools and would love to be able to just reuse my existing codex subscription

wey-gu:
> I was also working on trying to make a custom provider using OAuth codex login. So loved seeing this PR already open! This would be amazing to see added here. I’m also working on some agent tools and would love to be able to just reuse my existing codex subscription

Thanks @RileyMathews !

Glad this is helpful for you, too!

gold-silver-copper:
Thanks for the PR. Overall this looks good, but I have a couple of follow-up concerns about the auth flow.

Device flow polling: the OAuth polling logic is currently hardcoded to a fixed sleep/attempt count and only retries on authorization_pending. GitHub’s device flow docs mention `interval` and `slow_down`, so I think we should either handle those explicitly or document why that is not necessary here.

Cached token recovery: if a cached access token exists on disk but is stale or revoked, it looks like we return it directly and then fail when the Copilot token bootstrap rejects it. Is there a reason we don’t invalidate the cached token and fall back to re-auth in that case? Ideally this should recover automatically so the user does not need to manually clear cached tokens later.

wey-gu:
> Thanks for the PR. Overall this looks good, but I have a couple of follow-up concerns about the auth flow.
> 
> Device flow polling: the OAuth polling logic is currently hardcoded to a fixed sleep/attempt count and only retries on authorization_pending. GitHub’s device flow docs mention `interval` and `slow_down`, so I think we should either handle those explicitly or document why that is not necessary here.
> 
> Cached token recovery: if a cached access token exists on disk but is stale or revoked, it looks like we return it directly and then fail when the Copilot token bootstrap rejects it. Is there a reason we don’t invalidate the cached token and fall back to re-auth in that case? Ideally this should recover automatically so the user does not need to manually clear cached tokens later.

Thanks @gold-silver-copper for the great suggestions!

Now addressed both concerns in the latest push.

- GitHub Copilot device flow now uses the server-provided `interval`, honors `slow_down` by increasing the polling delay, and uses `expires_in` rather than a fixed attempt count.
- If a cached GitHub access token is stale or revoked, and Copilot token bootstrap rejects it with `401`/`403`, Rig now invalidates the cached token, re-runs device auth, and retries bootstrap automatically once.

Also added targeted tests around the new polling and stale-token recovery behavior.

Thanks again :)

gold-silver-copper:
I'm going to work on fixing the final errors and merging this with #1415 so we can get this in by the end of today.

wey-gu:
thanks @gold-silver-copper!

did a small patch to make ci happy

gold-silver-copper:
@wey-gu Hey there, I appreciate it but I'm going to revert it, I've already handled that on my local branch, sorry about that. I'm currently finishing up this PR, it'll be merged by tonight, don't worry about it :)

wey-gu:
> @wey-gu Hey there, I appreciate it but I'm going to revert it, I've already handled that on my local branch, sorry about that. I'm currently finishing up this PR, it'll be merged by tonight, don't worry about it :)

sure! sorry for that :-P thought i was helping follow the sun, didnt realize breaking your work :D

gold-silver-copper:
It's okay, I should've been communicating more :p

PR #1626, opened 2026-04-13
fix(bedrock): preserve all parallel tool calls in completion response

aleksmeshr:
The TryFrom<AwsConverseOutput> implementation had an early return that extracted only the first ToolCall from the response using find_map, wrapping it in OneOrMany::one() and discarding all other content blocks. This meant that when the model returned multiple parallel tool calls, only the first one was passed through.

The `choice` variable already contains all content blocks (text, reasoning, and all tool calls) correctly converted from the AWS response. Removing the early return lets the full set pass through.


gold-silver-copper:
Thank you!

PR #1639, opened 2026-04-16
fix(copilot): add missing cache_creation_input_tokens field

DAMEK86:
The Copilot provider's Usage construction was missing the cache_creation_input_tokens field that was added to the completion module's Usage struct. This caused a compilation error (E0063) on main after the Copilot provider was merged.


joshua-mo-143:
@gold-silver-copper think we missed one

Also for clarification: I don't work at Playgrounds anymore, Stephen who I have tagged is the lead maintainer now 🙂 apologies for the confusion!

DAMEK86:
Sad to hear. wish u the best @joshua-mo-143 !!

PR #1645, opened 2026-04-17
fix(type): fix filter types

### LINKED ISSUE #1644: bug: type mismatch on SurrealDB integration

## Bug Report

```
type mismatch resolving <SurrealSearchFilter as SearchFilter>::Value == Value
surrealdb::surrealdb_types::Value and serde_json::value::Value have similar names, but are actually distinct types
required for SurrealVectorStore<Db, rig::providers::openai::EmbeddingModel> to implement VectorStoreIndexDyn
```

## Reproduction

```rust
let base_url = std::env::var("OPENAI_BASE_URL").expect("OPENAI_BASE_URL not set");
let key = std::env::var("OPENAI_API_KEY").expect("OPENAI_API_KEY not set");
let client = openai::Client::builder()
    .api_key(key)
    .base_url(base_url)
    .build()
    .unwrap();
let model = client.embedding_model(MINIMAX_M2);
let db = surrealdb::Surreal::new::<Mem>(()).await.unwrap();
let vector = SurrealVectorStore::with_defaults(model, db);

let agent = client
    .agent(MINIMAX_M2_7)
    .preamble("You are a dice rolling AI. Always use tools to roll dice.")
    .tools(tools())
    .temperature(0.7)
    .max_tokens(4096)
    .dynamic_context(3, vector)
    .build();
```

## Expected behavior

Pass the compilation

## Additional context

`SurrealSearchFilter::Value` seems expect to be `serde_json::Value`

fu050409:
Resolves: #1644

### DIFF
--- a/rig-integrations/rig-surrealdb/examples/vector_store.rs
+++ b/rig-integrations/rig-surrealdb/examples/vector_store.rs
@@ -1,0 +1,68 @@
+use rig::client::{CompletionClient, EmbeddingsClient, ProviderClient};
+use rig::providers::openai;
+use rig::{Embed, embeddings::EmbeddingsBuilder, vector_store::InsertDocuments};
+use rig_surrealdb::{Mem, SurrealVectorStore};
+use serde::{Deserialize, Serialize};
+use surrealdb::Surreal;
+
+// A vector search needs to be performed on the `definitions` field, so we derive the `Embed` trait for `WordDefinition`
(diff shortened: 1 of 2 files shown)


gold-silver-copper:
Surrealdb already has a native Value::from_t(...) for converting between sql Values and json Values. But this wasn't the issue ultimately, I created a different PR that fixed the actual issue in the filter

Issue #1434, opened 2026-02-25
bug(deepseek): reasoning_content dropped from assistant messages with tool_calls, breaking multi-turn tool loops


kawayiYokami:
## Bug Description

In `providers/deepseek.rs`, the `TryFrom<message::Message> for Vec<Message>` implementation splits a single rig `Message::Assistant` (containing `Reasoning` + `ToolCall` content) into two separate DeepSeek `Message::Assistant` entries:

1. First: `reasoning_content = Some(...)`, `tool_calls = vec![]`
2. Second: `tool_calls = [...]`, `reasoning_content = None`

This causes DeepSeek API to return **400 Bad Request**:
> Missing reasoning_content field in the assistant message at message index ...

PR #1333 added a workaround that merges the last reasoning into the last assistant message during request construction. But this only fixes the **last** assistant message — in multi-turn tool loops, intermediate messages still lack `reasoning_content`.

## Root Cause

The core issue is that the conversion layer **discards** the `reasoning_content` from the message that carries `tool_calls`:

```rust
// line 339-345: tool_calls message always gets reasoning_content: None
if !tool_calls.is_empty() {
    messages.push(Message::Assistant {
        content: "".to_string(),
        tool_calls,
        reasoning_content: None,  // ← always None, reasoning is lost
    });
}
```

## Suggested Fix

The SDK should not try to cache or manage reasoning content itself. The caller (application code managing the tool loop) is responsible for passing `AssistantContent::Reasoning` back in `chat_history`. The SDK just needs to faithfully serialize it.

The fix is straightforward — put `reasoning_content` on the message that carries `tool_calls`, since that's what DeepSeek expects:

```rust
message::Message::Assistant { content, .. } => {
    let mut text_content = String::new();
    let mut reasoning_content = String::new();
    let mut tool_calls = Vec::new();

    for item in content.iter() {
        match item {
            message::AssistantContent::Text(text) => {
                text_content.push_str(text.text());
            }
            message::AssistantContent::Reasoning(reasoning) => {
                reasoning_content.push_str(&reasoning.display_text());
            }
            message::AssistantContent::ToolCall(tool_call) => {
                tool_calls.push(ToolCall::from(tool_call.clone()));
            }
            _ => {}
        }
    }

    let reasoning = if reasoning_content.is_empty() {
        None
    } else {
        Some(reasoning_content)
    };

    // Single message — reasoning_content stays with tool_calls
    Ok(vec![Message::Assistant {
[...]

PR #1649, opened 2026-04-19
fix(openai): carry reasoning_content on assistant tool-call messages

indrazm:
## Bug Description

OpenAI-compatible providers serving thinking models (Kimi K2.5, DeepSeek-R1, GLM-4.6, Qwen3-Thinking) reject the second round of a multi-turn tool loop with:                                                                           
                                                                                                                                            
> 400 Bad Request: thinking is enabled but reasoning_content is missing in assistant tool call message at index N

```bash
Error: CompletionError: ProviderError: Invalid status code 400 Bad Request with message: {"error":{"message":"thinking is enabled but reasoning_content is missing in assistant tool call message at index 4","type":"invalid_request_error"}}
```
                                               
The Chat Completions serializer in `providers/openai/completion/mod.rs` silently dropped `AssistantContent::Reasoning` blocks. This PR attaches reasoning text to the same assistant message that carries `tool_calls`, which is what those providers require.                   
  
Same bug shape as #1434 (DeepSeek-native, fixed in #1459), extended to the OpenAI Chat Completions path that most OpenAI-compatible providers go through.                                                                
                                                                                                                                            
  ## Changes                                                                           
                                               
- `Message::Assistant` gains `reasoning_content: Option<String>` (`#[serde(skip_serializing_if = "Option::is_none")]`, so requests to canonical OpenAI are byte-identical to before).
- Outbound `TryFrom<OneOrMany<AssistantContent>> for Vec<Message>` collects reasoning text and attaches it to the message carrying `tool_calls`.                                                                                                                             
- Inbound `TryFrom<Message>` and `TryFrom<CompletionResponse>` reconstruct an `AssistantContent::Reasoning` block when the provider returns one, so structured-history replay is symmetric.                                                                                   
- OpenRouter's `Message::Assistant` maps the new field onto its existing `reasoning` field at the OpenAI↔OpenRouter conversion boundary.
- Trivial literal updates in `providers/groq.rs` and the openai client test.                                                              
                                                                                                                                            
## Tests                                                                                                                                  
                                                                                                                                            
New tests in `providers/openai/completion/mod.rs`:                                   

- `assistant_reasoning_alone_is_dropped`, reasoning-only turn still produces empty output (prior behavior preserved).                    
- `assistant_reasoning_is_attached_to_tool_call_message`, regression test asserting outbound JSON has `reasoning_content` populated when Reasoning + ToolCall are serialized together.
- `assistant_reasoning_content_roundtrips_back_to_rig_message`, inbound conversion reconstructs the Reasoning block.

## Test plan                                                                         
                                                                                                                                            
- [x] `cargo fmt --check` clean                                                      
- [x] `cargo clippy --lib -- -D warnings` clean
- [x] `cargo test -p rig-core --lib` 445 passed, 0 failed (unchanged from main)                                                         
- [x] Verified end-to-end on a live multi-turn tool loop against an OpenAI-compatible thinking-model endpoint
                                                                                                                                            
## Reproduction (without this patch)                                                                                                      
                                                                                                                                            
1. Configure an OpenAI-compatible endpoint serving a thinking model with thinking enabled.                                                
2. Run an agent with tools.                  
3. Prompt something that requires ≥2 rounds of tool calls.                                                                                
4. Round 2 returns 400. With this patch: round 2 succeeds.

### DIFF
--- a/rig/rig-core/src/providers/openai/completion/mod.rs
+++ b/rig/rig-core/src/providers/openai/completion/mod.rs
@@ -156,7 +156,8 @@ pub enum Message {
             serialize_with = "serialize_assistant_content_vec"
         )]
         content: Vec<AssistantContent>,
-        // provided by llama.cpp
+        // OpenAI-compatible providers expose hidden reasoning on this non-standard
+        // field, and some require it to be echoed back on assistant tool-call turns.
         #[serde(skip_serializing_if = "Option::is_none", rename = "reasoning_content")]
         reasoning: Option<String>,
         #[serde(skip_serializing_if = "Option::is_none")]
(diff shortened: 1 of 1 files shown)


gold-silver-copper:
Thanks for the PR, i rebased it with main and merged.

PR #1657, opened 2026-04-21
fix(rig-1283): handle llama.cpp reasoning_content as content

### LINKED ISSUE #1656: bug: llama.cpp reasoning - CompletionError: ResponseError: Response contained no message or tool call (empty)
## Bug Report

When llama.cpp responds with `reasoning_content` in completion we get the error:
`CompletionError: ResponseError: Response contained no message or tool call (empty)`

## Reproduction

```json
{
    "choices": [
        {
            "finish_reason": "stop",
            "index": 0,
            "message": {
                "role": "assistant",
                "content": "",
                "reasoning_content": "Now I understand the structure better. I need to: ..."
            }
        }
    ],
  ...
}
```

## Expected behavior

I think llama.cpp is not strictly conforming to the openai api here. But it would be nice if the `reasoning_content` could just be assumed as `content`.

inqode-lars:
closes https://github.com/0xPlaygrounds/rig/issues/1656

### DIFF
--- a/rig/rig-core/src/providers/llamafile.rs
+++ b/rig/rig-core/src/providers/llamafile.rs
@@ -249,6 +249,7 @@ fn llamafile_message_value(message: openai::Message) -> Result<Value, Completion
         openai::Message::Assistant {
             content,
             refusal,
+            reasoning: _,
             audio,
             name,
             tool_calls,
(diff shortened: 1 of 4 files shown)


gold-silver-copper:
lgtm

PR #1666, opened 2026-04-24
fix(telemetry): emit OTel-standard gen_ai.usage.cache_read.input_tokens across providers

alwayys-afk:
## Summary
- Rename the tracing/OTel attribute `gen_ai.usage.cached_tokens` to `gen_ai.usage.cache_read.input_tokens` across every provider that emits it.
- `gen_ai.usage.cache_read.input_tokens` is the canonical attribute name listed in the [OpenTelemetry GenAI semantic-conventions registry](https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/). The previous name was non-standard and would not be picked up by OTel-aware backends.
- Every instance — span declarations and `span.record(...)` call sites, streaming and non-streaming paths — has been updated so a single provider never emits two different attribute names.

## Providers touched
openai (chat + responses_api, streaming + non-streaming), openai chat-completions-compatible shared helper, anthropic (via the existing `cached_input_tokens` plumbing), azure, chatgpt, cohere, copilot, deepseek, galadriel, gemini (completion, streaming, interactions_api), groq, huggingface, hyperbolic, llamafile, mira, mistral, moonshot, ollama, openrouter, perplexity, together, xai.

## Notes
- No provider wire-format changes. Provider response structs that happen to deserialize a JSON field named `cached_tokens` (e.g. DeepSeek's `prompt_tokens_details.cached_tokens`, OpenAI's `input_tokens_details.cached_tokens`) keep those names — those are the provider's API, not our telemetry.
- Per OTel guidance, `cache_read.input_tokens` SHOULD be included in `gen_ai.usage.input_tokens`. Existing providers already follow that convention; this PR does not change any token-accounting logic.
- This is a breaking change for any downstream dashboards/alerts that queried the old attribute name.

## Test plan
- [x] `cargo fmt -- --check`
- [x] `cargo clippy -p rig-core --all-targets --all-features` — clean
- [x] `cargo test -p rig-core --all-features --lib` — 555 passed, 0 failed, 8 ignored
- [x] `grep -rn "gen_ai.usage.cached_tokens" rig/` returns no matches

### DIFF
--- a/rig/rig-core/src/providers/azure.rs
+++ b/rig/rig-core/src/providers/azure.rs
@@ -709,7 +709,7 @@ where
                 gen_ai.response.model = tracing::field::Empty,
                 gen_ai.usage.output_tokens = tracing::field::Empty,
                 gen_ai.usage.input_tokens = tracing::field::Empty,
-                gen_ai.usage.cached_tokens = tracing::field::Empty,
+                gen_ai.usage.cache_read.input_tokens = tracing::field::Empty,
             )
         } else {
             tracing::Span::current()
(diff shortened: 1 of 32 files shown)


anish-kristipati (approved):
Good catch on the caching telemetry issues

PR #1681, opened 2026-04-28
feat(core): add ToolCallHookAction::ContinueWith for tool argument mutation

### LINKED ISSUE #1680: feat: allow `on_tool_call` hook to modify tool arguments before execution

## Feature Request

Add argument mutation support to the `on_tool_call` prompt hook. Today the hook can observe, reject, or terminate — but it cannot modify the arguments that will be passed to the tool. This forces callers into workarounds that bypass rig's tool dispatch, lose tracing/observability, and make `on_tool_result` inconsistent.

### Motivation

Three concrete use cases require argument mutation before tool execution:

1. **Argument sanitization** — A policy layer wants to strip file paths outside a sandbox, clamp numeric ranges, or redact PII from tool inputs before the tool sees them. Today the only options are `Continue` (run with original args) or `Skip` (don't run at all).

2. **Argument enrichment** — A hook wants to inject a default parameter the LLM omitted (e.g. adding `timezone: "UTC"` to a time query, or adding a `limit` to an unbounded database query).

3. **Argument normalization** — Converting between formats or fixing common LLM mistakes in tool arguments (e.g. normalizing date formats) without rejecting the call.

**Current workaround:** Callers return `Skip` from `on_tool_call`, manually call the tool inside the hook with modified args, and stuff the real result into the `reason` field of `Skip { reason }`. This works but has real downsides:
- The hook takes on tool dispatch, error handling, and tracing that rig normally owns
- `on_tool_result` never fires for the call (rig thinks it was skipped), breaking logging/auditing hooks downstream
- The workaround is invisible to any other hook in the chain

### Proposal

Add a `ContinueWith` variant to `ToolCallHookAction` that carries replacement arguments:

```rust
pub enum ToolCallHookAction {
    Continue,
    ContinueWith { args: String },  // new
    Skip { reason: String },
    Terminate { reason: String },
}
```

With a convenience constructor:

```rust
impl ToolCallHookAction {
    pub fn continue_with(args: impl Into<String>) -> Self {
        Self::ContinueWith { args: args.into() }
    }
}
```

The dispatch code in both `mod.rs` and `streaming.rs` would check for `ContinueWith` and use the replacement args for `call_tool` instead of the originals. The replacement args would also flow into `on_tool_result` and the tracing span, so downstream hooks and observability see the actual args that were executed.

**Semver note:** `ToolCallHookAction` is not currently `#[non_exhaustive]`, so adding a variant is a breaking change. Consider adding `#[non_exhaustive]` at the same time (this is already done for `MultiTurnStreamItem`, `PromptResponse`, and other public enums in rig-core).

### Alternatives

1. **Wrapper/middleware pattern around `Tool`** — Callers could wrap each tool to intercept and modify args at the tool level. This works but is per-tool rather than cross-cutting, and requires touching every tool registration rather than adding one hook.

2. **Do nothing** — The `Skip` workaround exists, but it silently breaks `on_tool_result` and tracing, which makes the hook system unreliable for auditing.

mathpal:
## Summary

- Add `ContinueWith { args: String }` variant to `ToolCallHookAction` so prompt hooks can modify tool arguments before execution
- Both non-streaming and streaming dispatch paths shadow the original args with the replacement, ensuring `call_tool`, `on_tool_result`, and tracing spans all see the actual arguments used
- Mark `ToolCallHookAction` and `HookAction` as `#[non_exhaustive]` for future extensibility

Fixes #1680

## Motivation

Today `on_tool_call` can observe, reject (`Skip`), or abort (`Terminate`) — but it cannot modify the arguments that will be passed to the tool. This forces callers into workarounds (returning `Skip` and manually calling the tool inside the hook) that bypass rig's dispatch, lose tracing, and make `on_tool_result` inconsistent.

Use cases: argument sanitization (sandbox file paths), argument enrichment (inject defaults the LLM omitted), argument normalization (fix common LLM mistakes in tool args).

### `hooks.rs`
- New `ContinueWith { args: String }` variant on `ToolCallHookAction`
- `continue_with()` convenience constructor
- `#[non_exhaustive]` on both `ToolCallHookAction` and `HookAction`
- Updated `on_tool_call` doc to list all four return variants
- 3 unit tests for the new constructor

### `mod.rs` (non-streaming dispatch)
- Refactored `if let` chain to exhaustive `match` on `ToolCallHookAction`
- `args` is shadowed with replacement when `ContinueWith` is returned
- Moved `tool_span.record("gen_ai.tool.call.arguments")` after the hook so it records the actual args used

### `streaming.rs` (streaming dispatch)
- Same refactor — `tool_args` is shadowed, downstream `call_tool`, `on_tool_result`, and tracing all see the replacement

### Tests
- **`tests/core/hook_continue_with.rs`** — 7 mock-based CI-runnable tests (no API key needed):
  - `continue_passes_original_args` — regression: `Continue` still works
  - `continue_with_replaces_args_in_call_tool` — contract: tool computes 100+200=300, not 1+2=3
  - `continue_with_replacement_args_reach_on_tool_result` — contract: `on_tool_result` sees replacement args
  - `terminate_still_cancels_loop` — regression: `Terminate` still works after refactor
  - `skip_returns_reason_and_bypasses_on_tool_result` — regression: `Skip` still works, `on_tool_result` does NOT fire
  - `continue_with_invalid_json_does_not_panic` — boundary: invalid JSON args → graceful tool error
  - `continue_with_wrong_fields_triggers_tool_error` — error path: missing fields → deser error becomes tool result
- **`tests/openai/arg_rewrite_hook.rs`** — integration test (requires API key) verifying end-to-end arg rewrite

## Breaking Changes

- `ToolCallHookAction` and `HookAction` are now `#[non_exhaustive]` — downstream crates with exhaustive `match` on these enums will need a wildcard arm
- New `ContinueWith` variant — not breaking by itself since the enums are now `#[non_exhaustive]`

## AI Disclosure

This PR was significantly assisted by AI (Claude). All code has been reviewed, tested, and verified by the contributor.

### DIFF
--- a/rig/rig-core/src/agent/prompt_request/hooks.rs
+++ b/rig/rig-core/src/agent/prompt_request/hooks.rs
@@ -35,7 +35,9 @@ where
     ///
     /// # Returns
     /// - `ToolCallHookAction::Continue` - Allow tool execution to proceed
+    /// - `ToolCallHookAction::ContinueWith { args }` - Allow tool execution with replacement arguments
     /// - `ToolCallHookAction::Skip { reason }` - Reject tool execution; `reason` will be returned to the LLM as the tool result
+    /// - `ToolCallHookAction::Terminate { reason }` - Terminate the agent loop early
     fn on_tool_call(
         &self,
         _tool_name: &str,
@@ -93,9 +95,16 @@ impl<M> PromptHook<M> for () where M: CompletionModel {}
 
 /// Control flow action for tool call hooks. This is different from the regular [`HookAction`] in that tool call executions may be skipped for one or more reasons.
 #[derive(Debug, Clone, PartialEq, Eq)]
+#[non_exhaustive]
 pub enum ToolCallHookAction {
     /// Continue tool execution as normal.
     Continue,
+    /// Continue tool execution with replacement arguments.
+    ///
+    /// The tool will be invoked with the provided `args` instead of the
+    /// original arguments from the LLM. Downstream hooks (`on_tool_result`)
+    /// and tracing spans will also see the replacement arguments.
+    ContinueWith { args: String },
     /// Skip tool execution and return the provided reason as the tool result.
     Skip { reason: String },
     /// Terminate agent loop early
@@ -106,6 +115,11 @@ impl ToolCallHookAction {
     /// Continue the agentic loop as normal
     pub fn cont() -> Self {
         Self::Continue
+    }
+
+    /// Continue tool execution with replacement arguments.
+    pub fn continue_with(args: impl Into<String>) -> Self {
+        Self::ContinueWith { args: args.into() }
     }
 
     /// Skip a given tool call (with a provided reason).
@@ -125,6 +139,7 @@ impl ToolCallHookAction {
 
 /// Control flow action for hooks.
 #[derive(Debug, Clone, PartialEq, Eq)]
+#[non_exhaustive]
 pub enum HookAction {
     /// Continue agentic loop execution as normal.
     Continue,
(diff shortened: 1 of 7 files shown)


gold-silver-copper:
This has been implemented in #1963

PR #1709, opened 2026-05-01
fix(openai): send PDF Documents as file parts in chat completions

### LINKED ISSUE #1708: bug: openai chat completions silently corrupts binary PDF documents

## Bug Report

The OpenAI **Chat Completions** provider silently corrupts `UserContent::Document` payloads carrying binary PDFs. The serializer at `crates/rig-core/src/providers/openai/completion/mod.rs:497-505` flattens the base64-encoded PDF bytes into a `text` content part, instead of emitting OpenAI's `{"type":"file","file":{...}}` content type. The model receives a raw base64 blob and hallucinates plausible content rather than reading the PDF.

Note: #446 explicitly describes OpenAI as *"Correctly converts it to text content and works as expected"*, but that holds only for *text* documents. Binary PDFs hit the same arm and are silently destroyed. The same path also affects LiteLLM-style proxies routed through rig's OpenAI provider.

Affected: rig-core 0.36.0 

## Reproduction

No API key needed. With `rig-core = "0.36"` and `serde_json = "1"`:

```rust
use rig::completion::message::{Document, DocumentMediaType, DocumentSourceKind, UserContent as MsgContent};
use rig::providers::openai::UserContent as WireContent;

fn main() {
    let doc = MsgContent::Document(Document {
        data: DocumentSourceKind::Base64("JVBERi0xLjQK".into()),
        media_type: Some(DocumentMediaType::PDF),
        additional_params: None,
    });
    let wire: WireContent = doc.try_into().unwrap();
    println!("{}", serde_json::to_string(&wire).unwrap());
}
```

**Actual:** `{"type":"text","text":"JVBERi0xLjQK"}`
**Expected:** `{"type":"file","file":{"file_data":"data:application/pdf;base64,JVBERi0xLjQK","filename":"document.pdf"}}`

## Expected behavior

`UserContent::Document` with `media_type: Some(PDF)` and `DocumentSourceKind::Base64(...)` should serialize as OpenAI's file content part so the model actually receives the PDF.

## Additional context

The wire enum at `completion/mod.rs:236-249` has no `File` variant, so the existing `TryFrom` has nothing to route PDFs into. The Responses API provider already does this correctly (`responses_api/mod.rs:1614-1621`). Related: #446 (the broader Document-handling family this issue narrows on).

fangkangmi:
# fix(openai): send PDF Documents as file parts in chat completions

## Description

Previously, `UserContent::Document` carrying a base64-encoded PDF was flattened into a plain `text` content part on the OpenAI Chat Completions path. The model never received the PDF and, with grounded prompts, would hallucinate plausible content matching the system prompt.

This PR adds a `File { file: FileData }` variant to the chat-completions `UserContent` enum matching OpenAI's `{"type":"file","file":{...}}` content type, and routes `UserContent::Document` with `media_type: Some(PDF)` into it. Mirrors the pattern already used by rig's Responses API provider and the OpenRouter provider (added in #1404 / commit `9cb2d9ca`). Non-PDF Documents continue to serialize as text, so existing markdown/text callers are unaffected.

The OpenRouter bridge (`From<openai::UserContent> for openrouter::UserContent`) is also extended to preserve `file_id` end to end. `openrouter::FileContent` gains a `file_id: Option<String>` field, matching the upstream OpenRouter OpenAPI schema (which accepts `file_id` on `type: "file"` content alongside `file_data`). Both `file_data` and `file_id`-only payloads now produce a valid OpenRouter `File` content part.

This unblocks PDF use both on native OpenAI chat completions and on LiteLLM-style proxies (Gemini, Anthropic, vLLM, etc.) routed through rig's OpenAI provider, since they share this serializer.

Fixes #1708
Refs #446

## Type of change

- [x] Bug fix

## Testing

8 new unit tests in `crates/rig-core/src/providers/openai/completion/mod.rs`:

OpenRouter test additions:

- New `test_user_content_file_id_serialization` pinning `file_id` on the wire
- Existing `file_deserialization` test extended to assert `file_id` deserializes
- Cross-conversion test extended to assert `openai::UserContent::File { file_id }` round-trips through to OpenRouter's `File` content with `file_id` preserved

Verification:

- [x] `cargo fmt --check`
- [x] `cargo clippy --all-targets --all-features -- -D warnings`
- [x] `cargo test --workspace` (539 lib tests + integration binaries, all green)
- [x] Manual canary verification with a PDF containing a unique token; the model returns the canary verbatim instead of hallucinating

## Checklist

- [x] I have made corresponding changes to the documentation (doc comments on new public items: `UserContent::File`, `FileData`, all fields; `FileContent::file_id` doc comment + struct doc updated)
- [x] I have added tests that prove my fix is effective

## Notes

Two additive public API surface changes:

1. New variant `File { file: FileData }` on the public `openai::completion::UserContent` enum (re-exported as `rig::providers::openai::UserContent`). The enum is not `#[non_exhaustive]`, so downstream code that exhaustively `match`es on it will need to add a new arm. Workspace-internal: only `openrouter/completion.rs` had such a match; patched in this PR.
2. New field `file_id: Option<String>` on the public `openrouter::completion::FileContent` struct. The struct is also not `#[non_exhaustive]`, so downstream construction-by-literal will need to add the field. Workspace-internal: 4 construction sites (`UserContent::file_url`, `UserContent::file_base64`, two arms in `TryFrom<message::UserContent>`); all patched in this PR.

Given Rig is pre-1.0 and the alternative is leaving binary PDFs silently corrupted (and `file_id` unreachable on OpenRouter), I went additive rather than retroactively adding `#[non_exhaustive]`. Happy to mark either type `#[non_exhaustive]` in a follow-up if maintainers prefer.

**Out of scope (separate RFC).** A natural follow-up is `DocumentSourceKind::FileId(String)` as a cross-provider primitive for OpenAI / Anthropic / Gemini Files APIs and Bedrock S3. The wire `FileData` and `FileContent` already carry `file_id`, so the high-level abstraction can land later without re-churning users.

### DIFF
--- a/crates/rig-core/src/providers/openai/completion/mod.rs
+++ b/crates/rig-core/src/providers/openai/completion/mod.rs
@@ -246,6 +246,14 @@ pub enum UserContent {
     Audio {
         input_audio: InputAudio,
     },
+    /// File content part for documents such as PDFs.
+    ///
+    /// Maps to OpenAI's `{"type":"file","file":{...}}` content type. Either
+    /// `file_data` (a base64 data URI like `data:application/pdf;base64,...`)
+    /// or `file_id` (a previously uploaded file reference) must be set.
+    File {
+        file: FileData,
+    },
 }
 
 #[derive(Debug, Serialize, Deserialize, PartialEq, Clone)]
@@ -259,6 +267,25 @@ pub struct ImageUrl {
 pub struct InputAudio {
     pub data: String,
     pub format: AudioMediaType,
+}
+
+/// File payload for [`UserContent::File`].
+///
+/// At least one of `file_data` or `file_id` must be set for the content part
+/// to be accepted by OpenAI's chat completions API. `filename` is optional
+/// but recommended.
+#[derive(Debug, Serialize, Deserialize, PartialEq, Clone)]
+pub struct FileData {
+    /// Inline file data as a base64 data URI, e.g.
+    /// `data:application/pdf;base64,JVBERi0xLjQK...`.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub file_data: Option<String>,
+    /// Identifier of a previously uploaded file (OpenAI Files API).
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub file_id: Option<String>,
+    /// Display name of the file. Recommended for inline `file_data`.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub filename: Option<String>,
 }
 
 #[derive(Debug, Serialize, Deserialize, PartialEq, Clone)]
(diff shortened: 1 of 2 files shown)


fangkangmi:
Reopened as #1732 with clean history (no force-push). Same commit `21af5f88`, same content, same Linear ticket RIG-1296. Apologies for the churn.

gold-silver-copper:
No problem, I'll be reviewing this today, thanks for your hard work

PR #1712, opened 2026-05-02
fix(cli_chatbot): added a trailing newline after streamed agent response. 

### LINKED ISSUE #1711: bug: cli_chatbot closing separator renders on same line as streamed response

## Bug Report

When I was using ChatBotBuilder with an Agent (streaming mode), I found that the closing === separator is printed on the same line as the last token of the assistant's response.

## Reproduction

>Note: I am using rig version 0.36.0
Use `ChatBotBuilder::new().agent(agent).build().run()` with any model. The closing ================================================================ line will appear glued to the end of the response text.

My exact code snippet:

```rust
use rig::client::{CompletionClient, ProviderClient};
use rig::integrations::cli_chatbot::ChatBotBuilder;
use rig::providers::openai;

#[tokio::main]
async fn main() -> Result<(), Box<dyn std::error::Error>> {
    let client = openai::Client::from_env()?;

    let agent = client
        .agent("gpt-5.4-mini")
        .preamble("you are a helpful assistant")
        .build();

    ChatBotBuilder::new().agent(agent).build().run().await?;
    Ok(())
}
```

## Expected behavior

The separator should always start on a new line.

## Screenshots

Image

## Additional context

To my knowledge - In `src/integrations/cli_chatbot.rs`, the `AgentImpl::request` method streams tokens with `print!("{}", text)` (no trailing newline). After the stream ends, the `run()` loop immediately prints the closing
separator; but if the model's final chunk doesn't end with \n, the separator gets concatenated on the same terminal line as the response.

I can confirm that the `ChatImpl` path uses `println!` and renders correctly; the `AgentImpl` (streaming) path uses print! and doesn't. So basically the two code paths in the same module have inconsistent behavior 
for the same visual output.

Should I go ahead and open a PR with a fix for this?

Shaurya-Sethi:
## Summary

Fixes [issue #1711](https://github.com/0xPlaygrounds/rig/issues/1711)

This pull request makes a minor improvement to the CLI chatbot integration by adding a newline when the response stream ends, ensuring cleaner output formatting.

Before:
image

After fix:
image

### DIFF
--- a/crates/rig-core/src/integrations/cli_chatbot.rs
+++ b/crates/rig-core/src/integrations/cli_chatbot.rs
@@ -78,6 +78,7 @@ where
 
         loop {
             let Some(chunk) = response_stream.next().await else {
+                println!();
                 break Ok(acc);
             };
 


gold-silver-copper:
Nice, thanks for the fix

Issue #1717, opened 2026-05-03
feat: Add structured outputs for Open Router


Mnwa:
- [x] I have looked for existing issues (including closed) about this

## Feature Request

Support structured outputs for the OpenRouter provider.

Currently, using structured outputs with OpenRouter in `rig` returns:

```text
Structured outputs currently not supported for OpenRouter
````

OpenRouter now supports structured outputs for compatible models through the `response_format` request parameter, using `type: "json_schema"` and a `json_schema` object containing the schema. Their documentation also notes that structured outputs are supported for both non-streaming and streaming responses. 
[Structured Outputs documentation](https://openrouter.ai/docs/guides/features/structured-outputs)

### Motivation

I would like to use `rig` with OpenRouter while still benefiting from `rig`’s structured output / typed extraction APIs.

This is useful for agent workflows where the application expects a reliable typed response, such as:

* extracting structured data from text
* returning typed tool/planning results
* validating model output against a JSON Schema
* using the same `rig` structured output code across providers

OpenRouter is especially useful because it routes to many model providers, including models that already support structured outputs. OpenRouter’s docs describe structured outputs as a way to enforce JSON Schema validation, get consistent type-safe outputs, avoid parsing errors, and simplify response handling.

### Proposal

Add structured output support to the OpenRouter provider by translating `rig`’s structured output schema into OpenRouter’s `response_format` format.

OpenRouter expects requests like:

```json
{
  "messages": [
    { "role": "user", "content": "What's the weather like in London?" }
  ],
  "response_format": {
    "type": "json_schema",
    "json_schema": {
      "name": "weather",
      "strict": true,
      "schema": {
        "type": "object",
        "properties": {
          "location": {
            "type": "string",
            "description": "City or location name"
          },
          "temperature": {
            "type": "number",
            "description": "Temperature in Celsius"
          },
          "conditions": {
            "type": "string",
            "description": "Weather conditions description"
          }
        },
        "required": ["location", "temperature", "conditions"],
        "additionalProperties": false
      }
    }
  }
}
```

Possible implementation details:

* Reuse the existing `rig` schema generation used by providers that already support structured outputs.
* For OpenRouter requests, map the generated JSON Schema into:

```json
{
  "response_format": {
    "type": "json_schema",
    "json_schema": {
      "name": "<schema name>",
      "strict": true,
      "schema": { ... }
    }
  }
}
```

[...]

PR #1718, opened 2026-05-03
feat(openrouter): added a support of structured output

Mnwa:
Based on that [issue](https://github.com/0xPlaygrounds/rig/issues/1717).
I test that realisation on my own project which depends on `rig-core` and it's working well.


gold-silver-copper:
Отличная работа, благодарим!

PR #1725, opened 2026-05-04
fix(mistral): expose cached and audio token fields in Usage

### LINKED ISSUE #1724: Mistral provider drops cached / audio token fields from Usage
## Summary

`mistral::client::Usage` only deserializes `prompt_tokens`, `completion_tokens`, and `total_tokens` — the additional fields Mistral now returns under `UsageInfo` are silently dropped.

Per the official Mistral OpenAPI spec (`https://docs.mistral.ai/openapi.yaml`, `UsageInfo` schema):

```yaml
UsageInfo:
  required: [prompt_tokens, completion_tokens, total_tokens]
  properties:
    prompt_tokens:         integer
    completion_tokens:     integer
    total_tokens:          integer
    prompt_audio_seconds:  integer | null
    num_cached_tokens:     integer | null
    prompt_tokens_details: { cached_tokens: integer } | null
    prompt_token_details:  { cached_tokens: integer } | null   # singular alias
```

## Impact

- Cached prompt-token information from Mistral never reaches downstream consumers.
- `GetTokenUsage for mistral::CompletionResponse` (`crates/rig-core/src/providers/mistral/completion.rs:477`) and the `TryFrom` mapping at `completion.rs:541` both hardcode `cached_input_tokens: 0` because the source struct can't carry the value, even when the API returns it.
- OpenAI's Responses API in this same crate properly surfaces cached tokens via `ResponsesUsage.input_tokens_details.cached_tokens`. Mistral has the equivalent shape (`prompt_tokens_details.cached_tokens`) but rig discards it.

Verified empirically against `https://api.mistral.ai/v1/chat/completions` with a repeated large prompt: the second call returned `prompt_tokens_details.cached_tokens: 3968`, which is currently lost on the way through rig.

## Proposed fix

Add the optional fields to `Usage` (with a `prompt_token_details` serde alias for the singular form), introduce a `PromptTokensDetails` struct, and thread `cached_tokens` through both `GetTokenUsage` and the `TryFrom<CompletionResponse>` mapping. The change is purely additive — existing required fields are unchanged and serde ignores unknown fields by default, so no consumer breaks.

byQuexo:
## Summary

Adds the missing optional fields from Mistral's `UsageInfo` schema to `mistral::client::Usage` so cached and audio token information stops being silently dropped.

Per the official OpenAPI spec at `https://docs.mistral.ai/openapi.yaml`:

```yaml
UsageInfo:
  required: [prompt_tokens, completion_tokens, total_tokens]
  properties:
    prompt_tokens:         integer
    completion_tokens:     integer
    total_tokens:          integer
    prompt_audio_seconds:  integer | null
    num_cached_tokens:     integer | null
    prompt_tokens_details: { cached_tokens: integer } | null
    prompt_token_details:  { cached_tokens: integer } | null   # singular alias
```

## Changes

- `mistral::client::Usage` gains `prompt_audio_seconds`, `num_cached_tokens`, and `prompt_tokens_details: Option<PromptTokensDetails>` (with `serde(alias = "prompt_token_details")` to handle Mistral's singular variant).
- New `Usage::cached_tokens()` helper — prefers `prompt_tokens_details.cached_tokens`, falls back to `num_cached_tokens`, returns `0` if neither is present.
- `GetTokenUsage` impl and the `TryFrom<CompletionResponse>` mapping at `mistral/completion.rs:541` now thread cached tokens through to `cached_input_tokens` instead of hardcoding `0`.

The change is purely additive: existing required fields are unchanged, all new fields are `Option` with `#[serde(default)]`, and serde ignores unknown fields by default. No public-API breakage.

## Tests

Five new unit tests in `mistral::completion::tests`:

- `test_usage_deserializes_prompt_tokens_details_cached_tokens`
- `test_usage_accepts_singular_prompt_token_details_alias`
- `test_usage_falls_back_to_num_cached_tokens`
- `test_usage_prefers_prompt_tokens_details_over_num_cached_tokens`
- `test_token_usage_threads_cached_tokens_into_completion_usage`

The existing `test_response_deserialization` was adjusted (it exhaustively destructured `Usage`) to validate the new fields default to `None`.

```
cargo fmt && cargo clippy --all-targets --all-features  # clean
cargo test -p rig-core --lib  # 535 passed, 0 failed
```

Also verified end-to-end against a downstream consumer of `rig-core`: a repeated oversized prompt to `mistral:codestral-latest` now surfaces `cached_tokens: 3968` on the second call (was always `0` before the fix).

Closes #1724

### DIFF
--- a/crates/rig-core/src/providers/mistral/client.rs
+++ b/crates/rig-core/src/providers/mistral/client.rs
@@ -81,11 +81,57 @@ impl ProviderClient for Client {
     }
 }
 
+/// In-depth details on prompt tokens.
+///
+/// Mirrors Mistral's `PromptTokensDetails` schema. The Mistral API also exposes
+/// the same shape under the singular field name `prompt_token_details`; the
+/// `Usage` field accepts either form via `serde(alias = ...)`.
+#[derive(Clone, Debug, Default, Deserialize, Serialize)]
+pub struct PromptTokensDetails {
+    /// Number of tokens served from the prompt cache.
+    #[serde(default)]
+    pub cached_tokens: u64,
+}
+
+/// Token usage returned by Mistral's chat completions and embeddings endpoints.
+///
+/// See <https://docs.mistral.ai/api/> (`UsageInfo` schema). The three counts are
+/// always present; the remaining fields are populated by Mistral on a best-effort
+/// basis (e.g. cached-token information appears once a prompt is large enough to
+/// be cached).
 #[derive(Clone, Debug, Deserialize, Serialize)]
 pub struct Usage {
     pub completion_tokens: usize,
     pub prompt_tokens: usize,
     pub total_tokens: usize,
+    /// Duration in seconds of audio tokens in the prompt (audio-input models only).
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub prompt_audio_seconds: Option<u64>,
+    /// Total cached prompt tokens reported at the top level. Some Mistral
+    /// responses populate this in addition to (or instead of)
+    /// `prompt_tokens_details.cached_tokens`.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub num_cached_tokens: Option<u64>,
+    /// In-depth breakdown of prompt token usage (currently only cached tokens).
+    #[serde(
+        default,
+        alias = "prompt_token_details",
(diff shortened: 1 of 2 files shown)


byQuexo:
I will do a Follow-up for streaming soon.

gold-silver-copper:
Thank you 🙏

PR #1726, opened 2026-05-04
Change rig_qdrant::filter module to public

bguo068:
The `filter` module in rig-qdrant is private, making it inaccessible. Changing it to public appears to work.


gold-silver-copper:
Thank you for bringing this up, I fixed this with a slightly more comprehensive PR #1727

PR #1731, opened 2026-05-05
feat(core)!: take &mut Vec<Message> in Chat trait so reasoning and tool calls round-trip

### LINKED ISSUE #1556: Should the `Chat` trait take `&mut Vec<Message>` instead of `Vec<Message>`
i am implemeting a Chat Bot using rig. However, reasoning and tool calls are lost when using the high-level Chat trait. 

These issues could be solved by passing a mutable reference to the chat_history. 

The underlying implemetation of `Chat` already does this.

```

#[allow(refining_impl_trait)]
impl<M, P> Chat for Agent<M, P>
where
    M: CompletionModel,
    P: PromptHook<M> + 'static,
{
    #[tracing::instrument(skip(self, prompt, chat_history), fields(agent_name = self.name()))]
    async fn chat(
        &self,
        prompt: impl Into<Message> + WasmCompatSend,
        mut chat_history: Vec<Message>,
    ) -> Result<String, PromptError> {
        PromptRequest::from_agent(self, prompt)
            .with_history(&mut chat_history)
            .await
    }
}
```

Why not expose this?

SarthakB11:
## Summary

The `Chat` trait consumed history as `IntoIterator<Item: Into<Message>>` and returned only the final assistant `String`. Any reasoning blocks, tool calls, or tool results produced during the turn were dropped, so a chatbot built on `Chat::chat` had no way to round-trip those messages back into its persistent history. `PromptRequest::extended_details()` already builds them internally, this PR just exposes them through the trait.

The new signature takes `chat_history: &mut Vec<Message>` and appends the prompt plus every assistant and tool message produced during the turn before returning. Persisting history across turns is now just keeping the same `Vec<Message>` around.

Closes #1556

## API contract

`chat` owns the appending of the prompt. Callers must NOT pre-push the user message into `chat_history` before calling, `chat` will append it for you, along with any assistant and tool messages produced during the turn.

```rust
// before
let response = agent.chat(prompt, history).await?;

// after
let response = agent.chat(prompt, &mut history).await?;
// history now contains: [...prior turns, User(prompt), Assistant(...), (optional tool turns)]
```

## Changes

- `Chat::chat` now has the signature:
  ```rust
  fn chat(
      &self,
      prompt: impl Into<Message> + WasmCompatSend,
      chat_history: &mut Vec<Message>,
  ) -> impl Future<Output = Result<String, PromptError>> + WasmCompatSend;
  ```
- `Chat for Agent<M, P>` drives `extended_details().send()` and extends `chat_history` with `PromptResponse::messages` (which is `[prompt, ...assistant/tool turns]`, no prior history), so callers see the full turn on return with no double-append.
- `cli_chatbot`:
  - `CliChat::request` now takes `history: &mut Vec<Message>` instead of an owned `Vec` that was silently discarded.
  - `ChatImpl::request` forwards the caller's history straight into `Chat::chat`, so reasoning and tool messages reach `ChatBot::run`'s persistent buffer.
  - `AgentImpl::request` (streaming path) appends `Message::user(prompt)` and `Message::assistant(acc)` after a successful stream. The streaming path does not go through `Chat::chat`, so the contract is mirrored manually. Streaming intermediate reasoning and tool deltas are not captured here, that matches prior streaming behavior and is out of scope.
  - The outer `ChatBot::run` loop drops its hand-rolled `history.push(Message::user(input))` / `Message::assistant(response)` lines, the trait now handles both.
- `discord_bot::handle_thread_message`:
  - Drops the pre-call `history.push(User { ... })` and the post-call `history.push(Assistant { ... })`. The latter also incidentally fixed a pre-existing bug where `msg.content` was being stored as the assistant's content.
  - Now calls `agent.chat(&msg.content, &mut history)` and writes the round-tripped `history` back into `state.conversations`, so threads accumulate full reasoning and tool history.
  - Drops the now-unused `OneOrMany`, `AssistantContent`, `UserContent` imports.
- `examples/multi_agent.rs`, the `agent` module doc-comment example, and 9 provider reasoning round-trip tests updated to pass `&mut Vec<Message>`.

## Tests

New `chat_appends_prompt_and_assistant_to_history` in `tests/core/prompt_response_messages.rs`. It uses the existing `SimpleTextModel` mock and asserts:

- `agent.chat("hi", &mut history)` leaves `history.len() == 2`, `history[0]` as `Message::User("hi")`, `history[1]` as `Message::Assistant("hello from mock")`.
- A second `agent.chat("again", &mut history)` grows history to 4 entries.

This locks in the appending contract per CONTRIBUTING.md's "new functionality must be tested" requirement.

## Breaking change

`Chat` is a public trait. There is one first-party impl (`Agent<M, P>`), so the in-tree footprint is the call sites listed above. Downstream users need to:

1. Pass `&mut history` (or `&mut Vec::<Message>::new()` for an empty buffer) instead of an owned iterable.
2. Stop pre-pushing the user message before calling, `chat` does that now.

## Verification

```
cargo check -p rig-core --all-targets   # clean
cargo clippy -p rig-core --all-targets  # clean
cargo test --test core chat_appends     # 1 passed
```

## Notes

AI assistance was used while preparing this change. Every line was reviewed by hand and the trait change matches the maintainer sketch in #1556.

### DIFF
--- a/crates/rig-core/src/agent/completion.rs
+++ b/crates/rig-core/src/agent/completion.rs
@@ -292,18 +292,19 @@ where
     P: PromptHook<M> + 'static,
 {
     #[tracing::instrument(skip(self, prompt, chat_history), fields(agent_name = self.name()))]
-    async fn chat<I, T>(
+    async fn chat(
         &self,
         prompt: impl Into<Message> + WasmCompatSend,
-        chat_history: I,
-    ) -> Result<String, PromptError>
-    where
-        I: IntoIterator<Item = T>,
-        T: Into<Message>,
-    {
-        PromptRequest::from_agent(self, prompt)
-            .with_history(chat_history)
-            .await
+        chat_history: &mut Vec<Message>,
+    ) -> Result<String, PromptError> {
+        let response = PromptRequest::from_agent(self, prompt)
+            .with_history(chat_history.clone())
+            .extended_details()
+            .await?;
+        if let Some(messages) = response.messages {
+            chat_history.extend(messages);
+        }
+        Ok(response.output)
     }
 }
 
--- a/crates/rig-core/src/agent/mod.rs
+++ b/crates/rig-core/src/agent/mod.rs
@@ -37,7 +37,7 @@
 //!
 //! // Use the agent for completions and prompts
 //! // Generate a chat completion response from a prompt and chat history
-//! let chat_response = agent.chat("Prompt", Vec::<rig_core::completion::Message>::new()).await?;
+//! let chat_response = agent.chat("Prompt", &mut Vec::<rig_core::completion::Message>::new()).await?;
 //!
 //! // Generate a prompt completion response from a simple prompt
 //! let prompt_response = agent.prompt("Prompt").await?;
--- a/crates/rig-core/src/completion/request.rs
+++ b/crates/rig-core/src/completion/request.rs
@@ -248,14 +248,16 @@ pub trait Chat: WasmCompatSend + WasmCompatSync {
     /// is returned as a string.
     ///
     /// If the tool does not exist, or the tool call fails, then an error is returned.
-    fn chat<I, T>(
+    ///
+    /// `chat_history` is taken by mutable reference: the prompt and any
+    /// assistant/tool messages produced during the turn are appended to it so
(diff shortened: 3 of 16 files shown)


gold-silver-copper:
Closed in favor of #1733

PR #1702, opened 2026-04-30
feat(memory): Rig-managed conversation memory + rig-memory companion crate

ForeverAngry:
# PR: Rig-managed conversation memory

Fixes #1701

> **AI assistance note:** the test suites in this PR (the `#[cfg(test)]`
> modules in `rig-core` and `rig-memory`) were AI-generated. The
> implementation, public API design, error model, and integration with
> the agent loop were authored by hand.

## Motivation

Rig agents currently require callers to pass `with_history(...)` on every prompt,
which forces every application to reinvent persistence, isolation, and shaping of
prior turns. This PR introduces first-class, Rig-managed conversation memory:

- A `ConversationMemory` trait that abstracts load / append / clear over a
  conversation id.
- A default in-process backend (`InMemoryConversationMemory`) with optional
  history-shaping via a closure-based `MessageFilter`.
- Per-agent and per-request entry points
  (`AgentBuilder::memory`, `AgentBuilder::conversation_id`,
  `PromptRequest::conversation`, `PromptRequest::without_memory`) that mirror
  the existing prompt builder ergonomics for both streaming and non-streaming.
- A new companion crate `rig-memory` shipping reusable, named history-shaping
  policies (`SlidingWindowMemory`, `TokenWindowMemory`) that callers can drop
  into the backend's filter slot.

The split keeps `rig-core` minimal — only the trait, the in-memory backend,
and the agent integration live in core — while named, opinionated policies and
their dependencies (eg. token counters) live in the companion crate.

## Architecture

```
rig-core::memory
├── ConversationMemory (trait)
├── MemoryError + MemoryBackendError
├── MessageFilter (trait + closure blanket impl)
└── InMemoryConversationMemory
        ├── new()
        └── with_filter<F: MessageFilter + 'static>(self, F)

rig-memory (companion crate)
├── re-exports ConversationMemory / InMemoryConversationMemory / MemoryError
├── MemoryPolicy (trait) + IntoFilter (blanket extension → MessageFilter)
├── NoopMemoryPolicy
├── SlidingWindowMemory::last_messages(n)
├── TokenCounter (trait) + Fn-blanket impl
└── TokenWindowMemory::new(max_tokens, counter)
```

Both windowing policies share a private `drop_leading_orphan_tool_result`
helper to avoid sending an orphaned tool result as the first message after
truncation (the leading turn would otherwise reference a tool call no longer
in context).

### `rig-core` (crate root → `rig::memory`)

- `pub mod memory;`
[...]

PR #1737, opened 2026-05-06
Add demotion hooks for bounded conversation memory

### LINKED ISSUE #1739: Bounded conversation memory: evicted turns have no path into long-lived memory
PR #1702 landed Rig-managed conversation memory with `ConversationMemory`, `MessageFilter`, `InMemoryConversationMemory`, and the `rig-memory` companion crate (`SlidingWindowMemory`, `TokenWindowMemory`, `MemoryPolicy`, `IntoFilter`).

While integrating that work downstream, one gap surfaced that we missed during #1702: when a windowing policy trims old turns, those messages are silently discarded. There is no seam for a downstream memory backend (persistent store, vector index, archival sink) to receive evicted turns, even though the rest of the stack — `ConversationMemory` for active history, prompt hooks for per-turn persistence, vector stores for recall — is already in place.

This is the missing handoff between "active, bounded conversation window" and "long-lived memory":

- `MemoryPolicy::apply` returns only the kept messages; the dropped prefix is unobservable.
- Companion crates that implement durable storage have no stable place to receive evictions.
- Without this seam, every adapter has to wrap the policy itself or reimplement windowing, which fragments the memory story.

Scope of the fix:

- A small trait in `rig-core` for "messages were demoted from active history" (no storage opinions).
- A policy adapter in `rig-memory` that composes any `ConversationMemory` + `MemoryPolicy` + demotion sink.
- Idempotency contract documented for adapters that need to survive process restarts.

Out of scope:

- Specific storage backends (memvid, sqlite, vector stores, etc.) — those live in their own companion crates and only need to implement the hook.
- LLM-backed summarisation as a demotion path — separate trait surface.

Tracking PR: #1737.

ForeverAngry:
## Summary

Follow-up to #1702. Adds the missing seam for evicted conversation turns so bounded active history can hand off to long-lived memory without `rig-core` taking on a storage opinion.

Fixes #1739.

Three small additions:

- `HeuristicTokenCounter` — zero-dep default so `TokenWindowMemory` is usable out of the box.
- `DemotionHook` (+ `NoopDemotionHook`) in `rig-core` — trait seam for messages removed from active history.
- `DemotingPolicyMemory` in `rig-memory` — composes any `ConversationMemory`, any `MemoryPolicy`, and any demotion hook. Tracks per-conversation in-process watermarks so repeated `load` calls do not redeliver the same demoted prefix; documents the idempotency contract for hooks that need to survive restarts.

## Why

#1702 gave `rig-core` first-class conversation memory and gave `rig-memory` named windowing policies. What was missing — and what we noticed only when wiring downstream backends — was the path for turns that fall *out* of the window. They were silently dropped.

This keeps that seam minimal: `rig-core` only defines the trait, `rig-memory` owns the policy adapter, storage crates choose whether to implement the hook.

## Notes for review

- The demotion watermark is intentionally in-process only. Durable hooks must be idempotent on `(conversation_id, messages)` because restarts or freshly constructed adapters can redeliver prior demotions.
- Concurrent `load` calls for the same conversation are gated at the demotion seam so only one call delivers to the hook at a time.
- `MemoryPolicy::apply_with_demoted` is the canonical method; `apply` remains the convenience wrapper for callers that only need the kept history.

## Verification

- `cargo test -p rig-core -p rig-memory`
- `cargo clippy -p rig-core -p rig-memory --all-targets --all-features`
- `cargo doc --no-deps -p rig-core -p rig-memory`

### DIFF
--- a/crates/rig-core/CHANGELOG.md
+++ b/crates/rig-core/CHANGELOG.md
@@ -36,6 +36,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `DemotingPolicyMemory<M, P, H>` adapter lives in `rig-memory`. Includes a
   forwarding `impl<H: DemotionHook + ?Sized> DemotionHook for Arc<H>` so
   hooks can be shared across multiple adapters.
+- `MemoryError::Internal(String)` variant for in-process invariant
+  violations (e.g. poisoned mutex guards), distinct from
+  `MemoryError::Backend` which is reserved for failures of the underlying
+  conversation store. `MemoryError` is now `#[non_exhaustive]` so future
+  variants are not breaking changes.
 
 ## [0.36.0](https://github.com/0xPlaygrounds/rig/compare/rig-core-v0.35.0...rig-core-v0.36.0) - 2026-04-28
 
--- a/crates/rig-core/CHANGELOG.md
+++ b/crates/rig-core/CHANGELOG.md
@@ -42,6 +42,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   conversation store. `MemoryError` is now `#[non_exhaustive]` so future
   variants are not breaking changes.
 
+  **Note for downstream crates:** `MemoryError` was previously a plain
+  enum, so any existing `match` against it without a wildcard arm will
+  now warn (and may need a wildcard arm if it was upgraded to a hard
+  error elsewhere). Adding `_ => ...` is forward-compatible with future
+  variants.
+
 ## [0.36.0](https://github.com/0xPlaygrounds/rig/compare/rig-core-v0.35.0...rig-core-v0.36.0) - 2026-04-28
 
 ### Added
--- a/crates/rig-core/CHANGELOG.md
+++ b/crates/rig-core/CHANGELOG.md
@@ -29,6 +29,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Examples: `agent_with_memory.rs` and `agent_with_memory_streaming.rs`.
   - Named history-shaping policies (sliding window, token budget) live in the
     new companion crate `rig-memory`.
+- `rig::memory::DemotionHook` trait and `NoopDemotionHook` no-op default.
+  Side-channel for messages that a memory policy or adapter removes from
+  active history. Defined in `rig-core` so any memory backend can implement
+  it without a `rig-memory` dependency; the composing
+  `DemotingPolicyMemory<M, P, H>` adapter lives in `rig-memory`. Includes a
+  forwarding `impl<H: DemotionHook + ?Sized> DemotionHook for Arc<H>` so
+  hooks can be shared across multiple adapters.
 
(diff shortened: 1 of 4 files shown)


gold-silver-copper:
Awesome work, once you upgrade it from Draft status, I'll give it a deeper look.

ForeverAngry:
> Awesome work, once you upgrade it from Draft status, I'll give it a deeper look.

Gotcha, yeah i was trying to get more reps in composing agents with the changes i made, so that i didnt overlook anything else that might make it tough for end users to understand.  I should have the PR ready in later on tonight 8 or 9PM EST.

I really appreciate your responsiveness!

ForeverAngry:
@gold-silver-copper to help add a bit more context around these commits, this is how i have been thinking about the memory feature path, let me know what you think:

| Layer    | Name                          | What's new vs prior layer                     |
|----------|-------------------------------|-----------------------------------------------|
| #1702    | windowing / policy memory     | bounded active context                        |
| #1737    | eviction hook / demotion seam | observe what falls out                        |
| next     | compaction (rolling summary)  | replace evicted prefix with derived Message   |
| after    | recall / archival memory      | retrieve evicted content back per turn        |
| eventual | tiered / hierarchical memory  | router across all of the above                |

ForeverAngry:
re: copilot review. Pushed as `7773f861`. Per-comment:

1. **`MemoryError` non-exhaustive callout** — agreed, added an explicit downstream-impact note in `crates/rig-core/CHANGELOG.md` so the breaking surface (existing exhaustive `match` arms warn) is visible without having to read the diff.

2. **`SlidingWindowMemory` `first_ref()`** — applied. `OneOrMany::first_ref` was already in `rig-core`, so the orphan-tool-result check no longer clones the leading content block.

3. **`TokenWindowMemory` `first_ref()`** — same fix, applied.

4. **`chars().count()` → `str::len()`** — applied, plus the configurable ratio is renamed `chars_per_token` → `bytes_per_token` so the API matches the implementation. Bytes are arguably a better proxy for BPE tokenizers anyway, and for non-ASCII text the counter now slightly *over*-estimates, which is the safe direction for a hard budget. Doc updated to spell this out. Unreleased so no migration cost.

5. **In-flight failure-visibility doc** — applied. Added a "Failure visibility" paragraph to `DemotingPolicyMemory`'s concurrency docs explaining that hook errors are only returned to the caller that drove delivery; concurrent loads that short-circuited on `in_flight` see `Ok(kept)` even if the in-flight delivery later failed. The watermark stays unchanged so the next `load` retries — durability-sensitive callers should surface failures through the hook's own observability rather than the `load` return value.

6. **`entry(to_string())` allocation per load** — applied. The hot path now uses `HashMap::get_mut(conversation_id)` and only allocates a new `String` key when first recording state for a previously-unseen conversation *and* there is demotion work to track. For an active conversation with a steady stream of `load`s the per-call cost drops to a single `HashMap::get_mut` plus the work the gate already did.

All `cargo fmt --all -- --check`, `cargo test -p rig-core --lib`, `cargo test -p rig-memory`, and `cargo build -p rig-core -p rig-memory --examples` are clean.

gold-silver-copper:
LGTM and yeah that sounds like a solid plan. Thank you for working on this, we appreciate it ❤️

PR #1747, opened 2026-05-08
feat(ollama): Enhance `think` parameter with string levels

cobaltburn:
Extends the `think` parameter in Ollama completion requests to accept string values "low", "medium", or "high", in addition to boolean. This provides more granular control over Ollama's internal thought process reporting.
documentation for thinking: https://docs.ollama.com/capabilities/thinking


gold-silver-copper:
Awesome work, thank you!

PR #1748, opened 2026-05-09
feat(memory): add Compactor trait, CompactingMemory adapter, and TemplateCompactor

ForeverAngry:
## Summary

Adds a `Compactor` trait to `rig_core::memory` and a composing
`CompactingMemory` adapter (with a zero-dependency `TemplateCompactor`
reference implementation) in `rig-memory`. Where #1737's `DemotionHook`
is a one-way drain that *observes* messages a policy truncates out of
active history, `Compactor` is the inverse: it returns a single
`Message`-shaped artifact derived from the evicted prefix, and the
adapter splices that artifact back at the front of the loaded history.
The resulting prompt shape is `[summary, ...recent_window]` — the
canonical recursive-summary pattern for long-running agents.

## Motivation

`DemotingPolicyMemory` (from #1737) gives backends a side channel for
durable persistence of demoted turns, but the *active* history a policy
returns is still a verbatim suffix. For long conversations this throws
away the head context entirely once a window slides past it. Existing
ecosystem code papers over this either by stuffing summaries into the
preamble (drifts as the conversation grows) or by writing custom
`ConversationMemory` impls (no shared idempotency / locking story).

The compaction seam needs the same composability story `PolicyMemory`
and `DemotingPolicyMemory` already have:

- a trait that lives in `rig_core` so any backend can implement it
  without taking a `rig-memory` dependency,
- a composing adapter that lives in `rig-memory` next to the policies,
- a reference implementation that does not pull in an LLM provider, so
  doctests, examples, and integration tests can exercise the full
  shape with no network.

### `rig_core::memory::Compactor`

```rust
pub trait Compactor: WasmCompatSend + WasmCompatSync {
    type Artifact: Into<Message> + Clone + WasmCompatSend + WasmCompatSync;

    fn compact<'a>(
        &'a self,
        conversation_id: &'a str,
        evicted: &'a [Message],
        carry_over: Option<&'a Self::Artifact>,
    ) -> WasmBoxedFuture<'a, Result<Self::Artifact, MemoryError>>;
}
```

- `Artifact: Into<Message>` keeps the trait `rig-core`-pure; no
  dependency on a specific summary representation.
- `carry_over` threads the previous summary so impls can do
  rolling-summary semantics without state of their own.
- Per-conversation watermarks are in-process only — implementations
  with durable side effects (LLM calls, vector-store writes) must
  deduplicate per the same idempotency contract documented for
  `DemotionHook`.

### `rig_memory::CompactingMemory<M, P, C>`

Wraps an inner `ConversationMemory` and a `MemoryPolicy`. On every
`load`:

1. Read inner history.
2. Apply the policy → `(kept, evicted_prefix)`.
3. If `evicted_prefix` is non-empty, invoke the compactor with the
   stored `carry_over` for that `conversation_id`; the returned
   artifact replaces `carry_over` and is converted into a `Message`
   spliced at position 0 of `kept`.
4. If `evicted_prefix` is empty, the previous `carry_over` (if any) is
   still spliced — so the summary survives later loads that don't
   evict anything new.

**Concurrency at the seam.** Concurrent loads on the same
`conversation_id` are serialised by an in-flight gate (a per-id
watcher). Only one load at a time invokes the compactor; others
observe the gate and immediately return the *previously*-stored
summary spliced in front of `kept`, without re-running the compactor.
This matters because compactors are typically expensive (LLM calls).
The contract is exercised by
`compacting_concurrent_with_clear_does_not_resurrect_state`: a
`clear` that lands while compaction is in flight is not overwritten
by the post-await state update.

**`clear` semantics.** `clear` drops the carry-over for that
conversation, so a freshly-populated backend re-compacts from scratch
rather than inheriting a stale summary from a deleted conversation.

### `rig_memory::TemplateCompactor`

Zero-dependency, no-LLM rollup. Produces a `TextSummary(String)` that
converts into a `Message::User` containing:

- a `[Conversation summary so far]` header,
- the previous summary (if any),
- one `role: text` line per evicted message, with an `[assistant tool
  call: name]` marker for tool-call assistant turns.

`with_max_bytes` provides a deterministic cap (truncating from the
middle with a `[…truncated…]` marker so the header and the most-recent
turns survive). Useful as a default, in doctests, and as the body of
integration tests that exercise the seam without needing a model.

## Behavioural changes

- New public surface in `rig_core::memory`: the `Compactor` trait. No
  existing types changed.
- New public surface in `rig_memory`: `CompactingMemory`,
  `TemplateCompactor`, `TextSummary`. Re-exports `Compactor` from
  `rig_core::memory` for symmetry with `DemotionHook`.
- No changes to `ConversationMemory`, `MemoryPolicy`, `DemotionHook`,
  `PolicyMemory`, or `DemotingPolicyMemory`. `CompactingMemory`
  composes with all of them.

## Tests

- 5 unit tests for `TemplateCompactor` covering: artifact-into-message
  conversion, system-message rendering, tool-call marker, carry-over
  threading, and `with_max_bytes` (including the
  `with_max_bytes(0) == unbounded` and unbounded-by-default edges).
- 1 integration test (`compacting_summary_stays_bounded_across_rolls`)
  driving multiple turns through a sliding window + `TemplateCompactor`
  to confirm summary size stays bounded across rolls.
- 1 concurrency test
  (`compacting_concurrent_with_clear_does_not_resurrect_state`) using
  a gated mock compactor + `tokio::sync::Notify` to verify the
  in-flight gate honours an interleaved `clear`.
- 1 doctest on `CompactingMemory` and 1 on `TemplateCompactor`,
  covering the public-facing usage shape.

`cargo test -p rig-core --lib memory` and `cargo test -p rig-memory`
green. `cargo clippy -p rig-memory --all-targets -- -D warnings` clean.
`RUSTDOCFLAGS="-D warnings" cargo doc -p rig-memory --no-deps` clean.

## Out of scope

- An LLM-backed `Compactor` — that wants a model handle, a prompt
  template, and a retry/backoff story; it lives downstream of this PR
  in user code (or a follow-up that opts into a provider feature).
- Persisting `carry_over` across process restarts — the trait
  contract documents that watermarks are in-process; durable carry-over
  is a Compactor-implementation concern.
- Wiring `CompactingMemory` into the agent builder as a sugar method.
  The existing `.memory(...)` builder takes any `ConversationMemory`,
  which `CompactingMemory` implements; no new builder surface needed.

### DIFF
--- a/crates/rig-core/CHANGELOG.md
+++ b/crates/rig-core/CHANGELOG.md
@@ -9,6 +9,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- `rig::memory::Compactor` trait. A `Compactor` produces a single
+  `Message`-shaped `Artifact` from a slice of messages a memory policy
+  has evicted, optionally combining it with the previous summary
+  (`carry_over`) for rolling-summary semantics. Where `DemotionHook`
+  is a one-way drain that observes evicted messages, `Compactor` is
+  the inverse: it returns an artifact that the composing adapter
+  splices back into the active history, so the loaded prompt shape is
+  `[summary, ...recent_window]` instead of a verbatim suffix. The
+  trait lives in `rig_core` so any memory backend can implement it
+  without taking a `rig-memory` dependency; the composing
+  `CompactingMemory` adapter and a zero-dependency `TemplateCompactor`
+  reference implementation live in `rig-memory`. Implementations with
+  durable side effects (LLM calls, vector-store writes) must
+  deduplicate per the same idempotency contract as `DemotionHook`,
+  since per-conversation watermarks are in-process only.
+
 - Rig-managed conversation memory:
   - New `rig::memory` module with the `ConversationMemory` trait,
     `MemoryError` (with a `MemoryBackendError` source type),
--- a/crates/rig-core/src/memory.rs
+++ b/crates/rig-core/src/memory.rs
@@ -207,6 +207,79 @@ where
     }
 }
 
+/// Derives a single [`Message`]-shaped artifact from a slice of messages
+/// that a memory policy has evicted from the active window.
+///
+/// Where a [`DemotionHook`] is a one-way drain — observe what fell out and
+/// return `()` — a `Compactor` is the inverse: it takes the evicted prefix
+/// (and optionally the previous summary) and produces a derived artifact
+/// that the composing adapter splices *back into* the active history. The
+/// resulting prompt is no longer a verbatim suffix of the conversation; it
+/// is `[summary, ...recent_window]`.
+///
+/// Implementations typically wrap an LLM call (`LlmCompactor<M>`) or a
+/// pure template rollup. They run inline on the load path whenever the
+/// policy demotes new messages, so a slow compactor delays the agent's
+/// next turn — keep them fast or offload to a cached/background pipeline.
+///
+/// # Rolling summaries
+///
+/// `carry_over` is the artifact produced by the previous compaction for
+/// this conversation, if any. Implementations that want a *recursive*
+/// summary (the canonical pattern for long-running agents) should
+/// summarize `evicted` *together with* `carry_over` so context lost in
+/// earlier compactions is preserved transitively. Stateless implementations
+/// can ignore `carry_over` and produce a fresh summary of `evicted` alone.
+///
+/// # Idempotency contract
+///
+/// Composing adapters track per-conversation in-process delivery so the
+/// same `evicted` slice is not compacted twice within a process lifetime,
+/// but those watermarks are not persisted across restarts. Implementations
+/// that have side effects (writing summaries to a vector store, billing an
+/// LLM call) should deduplicate by conversation id and content hash, the
+/// same way [`DemotionHook`] implementations do.
+pub trait Compactor: WasmCompatSend + WasmCompatSync {
+    /// The summary value produced by [`Compactor::compact`].
+    ///
+    /// `Into<Message>` is required so the composing adapter can splice the
+    /// artifact at the front of the loaded history. `Clone` is required so
+    /// the adapter can keep a private copy as `carry_over` for the next
+    /// compaction.
+    type Artifact: Into<Message> + Clone + WasmCompatSend + WasmCompatSync + 'static;
+
+    /// Produce a summary artifact for `evicted`, optionally combining it
+    /// with the previous summary in `carry_over`.
+    ///
+    /// `evicted` is in original conversation order. Errors are surfaced as
+    /// [`MemoryError::Backend`] (or any other variant the implementation
+    /// chooses) by the composing adapter.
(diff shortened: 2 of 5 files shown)


ForeverAngry:
Pushed `0fb9bef3` addressing all three Copilot review comments.

| # | File:line | Fix |
| - | --------- | --- |
| 1 | `crates/rig-memory/src/lib.rs:965` | Replaced the manual `in_flight = true` / post-await clear with an RAII `InFlightGuard<'_, A>` whose `Drop` clears the gate unless explicitly disarmed after the post-await state update. Future drops (caller cancellation, `tokio::select!`, panic in the compactor) now release the gate so subsequent loads can retry. Missing-entry case is a no-op so a concurrent `clear` + `forget` is still respected. New regression: `compacting_dropped_load_releases_in_flight_gate` aborts a `JoinHandle` while the compactor is parked on a `Notify` and asserts the next load enters the compactor a second time. |
| 2 | `crates/rig-memory/src/lib.rs:1136` | `truncate_summary` now locates the header/body boundary by scanning `buf` for the first `\n`, instead of trusting `header.len() + 1`. A header containing embedded newlines no longer mis-identifies the body. The `header` argument was removed since the assembled buffer is now self-describing. New regression: `template_compactor_caps_summary_with_multiline_header`. |
| 3 | `crates/rig-core/src/memory.rs:256` | Reworded `Compactor::compact` rustdoc: errors are propagated unchanged by composing adapters, with `MemoryError::Backend` and `MemoryError::Internal` called out as conventions (not as wrappers the adapter applies). |

Verification (local):

- `cargo fmt --all -- --check` clean
- `cargo test -p rig-core --lib memory` — 19 passed
- `cargo test -p rig-memory --lib` — 57 passed (2 new)
- `cargo clippy -p rig-memory -p rig-core --all-targets -- -D warnings` clean for the touched crates
- `RUSTDOCFLAGS="-D warnings" cargo doc -p rig-memory -p rig-core --no-deps` clean

### CI clippy note

The red `stable / clippy` check on the previous run is unrelated to this PR. It fires on `IndexStrategy` in `crates/rig-core/src/vector_store/in_memory_store.rs`:

```
error: this `impl` can be derived
   --> crates/rig-core/src/vector_store/in_memory_store.rs:262:1
    |
    = note: `-D clippy::derivable-impls` implied by `-D warnings`
```

Reproduced locally on `upstream/main` (commit `e9a4cc22`) under clippy 1.91.0 — the lint was promoted in a recent clippy version and the PR diff does not touch that file. Happy to fold a one-line `#[derive(Default)]` cleanup in here if maintainers prefer it as part of this PR; otherwise it's a separate baseline fix.

gold-silver-copper:
Hey there, please get clippy passing, we are planning on doing a rig release soon and would like to get this merged by then!

ForeverAngry:
Will do!

ForeverAngry:
> Hey there, please get clippy passing, we are planning on doing a rig release soon and would like to get this merged by then!

Should be good to merge, now.

PR #1755, opened 2026-05-12
docs: fix "a ancient" grammar in glarb-glarb sample text

abhicris:
Fixes a grammar error duplicated across 17 files where the sample document text reads "A glarb-glarb is a ancient tool" — the indefinite article should be `an` before the vowel-initial word `ancient`. Applied uniformly to all 25 occurrences across `examples/`, every per-provider example, and integration test fixtures so the canonical demo string stays consistent.

### Why

This text is the canonical RAG-demo sample shown to every new user landing on rig examples; the grammar slip is the first thing they see.


gold-silver-copper:
Thanks for catching this and providing the fix!

Issue #1750, opened 2026-05-10
feat: support TTS and STT (transcription) for openrouter provider


fversaci:
- [X] I have looked for existing issues (including closed) about this

## Feature Request
Openrouter api now supports both TTS and STT, it would be great to add it to rig.

### Proposal
I have already implemented this functionality in a fork [fversaci/rig](https://github.com/fversaci/rig/tree/openrouter-audio-gen), with assistance from MiniMax 2.7. I've attached a markdown doc describing the implementation. I'd be happy to submit a PR.

[openrouter-tts-stt.md](https://github.com/user-attachments/files/27569157/openrouter-tts-stt.md)

gold-silver-copper:
It would be great to support this. A PR from your fork would be lovely. Grazie!

fversaci:
Thanks for the quick merge! It's now used by [PolyRig](https://github.com/fversaci/polyrig), my voice-chat project (just published)!

PR #1757, opened 2026-05-12
feat(openrouter): add transcription (STT) and audio generation (TTS) support

fversaci:
PR to solve issue #1750 

Add  transcription.rs  and  audio_generation.rs  to the OpenRouter provider, implementing STT (speech-to-text) via
OpenRouter's  /api/v1/audio/transcriptions  endpoint and TTS via  /audio/speech . STT uses JSON with base64-encoded audio (differing from OpenAI's multipart transport) and exposes well-known Whisper, GPT-4o Transcribe, and Chirp 3 model constants. TTS defaults to MP3 output and supports provider-specific passthrough via  additional_params . Update  mod.rs with module declarations and re-exports, and change client capabilities in  client.rs  from  Nothing  to  Capable  for both transcription and audio generation (behind the  audio  feature gate).
[pr-description.md](https://github.com/user-attachments/files/27628786/pr-description.md)

### DIFF
--- a/crates/rig-core/src/providers/openrouter/audio_generation.rs
+++ b/crates/rig-core/src/providers/openrouter/audio_generation.rs
@@ -61,7 +61,11 @@ where
         .into_iter()
         .collect();
 
-         if let Some(obj) = request.additional_params.as_ref().and_then(|p| p.as_object()) {
+        if let Some(obj) = request
+            .additional_params
+            .as_ref()
+            .and_then(|p| p.as_object())
+        {
             for (k, v) in obj {
                 body_map.insert(k.clone(), v.clone());
             }
--- a/crates/rig-core/src/providers/openrouter/audio_generation.rs
+++ b/crates/rig-core/src/providers/openrouter/audio_generation.rs
@@ -1,0 +1,97 @@
+use crate::audio_generation::{
+    self, AudioGenerationError, AudioGenerationRequest, AudioGenerationResponse,
+};
+use crate::http_client::{self, HttpClientExt};
+use crate::providers::openrouter::Client;
+use bytes::Bytes;
+use serde_json::json;
+
+// ================================================================
+// Model constants
+// ================================================================
+
+/// The `openai/gpt-4o-mini-tts-2025-12-15` model.
+pub const GPT_4O_MINI_TTS: &str = "openai/gpt-4o-mini-tts-2025-12-15";
+/// The `mistralai/voxtral-mini-tts-2603` model.
+pub const VOXTRAL_MINI_TTS: &str = "mistralai/voxtral-mini-tts-2603";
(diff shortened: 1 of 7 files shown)


gold-silver-copper:
Per favore, correggi gli errori di linting 🙏

fversaci:
Fixed it is!

fversaci:
I'll check the issues :+1:

gold-silver-copper:
Sorry if you saw some spam, AI went rogue 😓 

Here is the proper review:

Requested changes:

1. **`infer_format_from_filename()`**

   Please use the actual file extension rather than `filename.split('.').nth(1)`. That currently reads the segment after the first dot, so a filename like `meeting.final.mp3` resolves to `final` and falls back to `wav`.

   `Path::extension()` would be the clearest fix. Please also keep the accepted formats aligned with OpenRouter’s documented STT formats: `wav`, `mp3`, `flac`, `m4a`, `ogg`, `webm`, and `aac`.

2. **`TranscriptionRequest::prompt`**

   Please handle this explicitly instead of silently ignoring it. Rig’s shared transcription builder exposes `.prompt()`, and users would reasonably expect it to either be sent or rejected.

   OpenRouter’s STT schema does not expose a top-level `prompt`; prompt-like behavior is routed through provider-specific options under `provider.options.<provider>.prompt`. If there is no safe provider-agnostic mapping, setting `.prompt()` should return a `TranscriptionError::RequestError` with guidance to use `additional_params`.

3. **`additional_params`**

   Please make non-object `additional_params` fail fast in both STT and TTS instead of silently dropping the value. This matches the safer behavior used by existing transcription providers and avoids making user-supplied provider configuration disappear without feedback.

4. **TTS bounds**

   Please replace the raw `Send` bound in the TTS implementation with Rig’s wasm-compatible bounds so wasm/custom HTTP clients are not over-constrained. Use WasmCompatSend and WasmCompatSync

5. **Feature gating / PR description**

   Please align the PR description with the implementation. Right now transcription is exposed unconditionally, while TTS is behind the `audio` feature. Either the description should reflect that, or the code should change if STT is intended to be gated as well.

6. **Response parsing**

   One additional response-shape issue: `TranscriptionUsage.seconds` is currently required in the response struct, but OpenRouter’s schema does not appear to require individual usage fields. Please make it optional to avoid brittle response parsing, or document the reason for requiring it.

fversaci:
Here is how the issues have been fixed: [fix-issues.md](https://github.com/user-attachments/files/27639396/fix-issues.md)
The specs of openrouter tts and stt: [openrouter-audio.md](https://github.com/user-attachments/files/27639398/openrouter-audio.md)
The current description of the PR: [pr-description.md](https://github.com/user-attachments/files/27639399/pr-description.md)

gold-silver-copper:
Grazie!

PR #1776, opened 2026-05-16
feat(gemini): expose finish_reason and model_version on StreamingCompletionResponse

mateobelanger:
## Summary

- Adds `finish_reason: Option<FinishReason>` and `model_version: Option<String>` to `gemini::streaming::StreamingCompletionResponse` — both were deserialized from the SSE stream but dropped before the `FinalResponse` was yielded to callers
- Adds `model_version: Option<String>` to `gemini::interactions_api::streaming::StreamingCompletionResponse`, extracted from `Interaction.model`
- Fixes a subtle bug: when the final Gemini chunk carries `finishReason` with no `content` (a common protocol pattern), the stream now correctly breaks instead of continuing to wait for more events
- The Interactions API has no `FinishReason` equivalent (it uses `InteractionStatus` lifecycle values); only `model_version` is added for that path, with a doc comment explaining why

## Motivation

Ryzome's PostHog LLM Analytics forwarder needs `$ai_stop_reason` (from `finish_reason`) and the resolved model string `$ai_model` (from `model_version`, which is the exact model like `gemini-2.5-pro-preview-05-06` rather than the alias the caller requested).

## Test Plan

- New unit test `test_streaming_completion_response_has_finish_reason_and_model_version` in `gemini/streaming.rs` asserts both fields populate and roundtrip through serde correctly
- New unit test `test_streaming_completion_response_has_model_version` in `interactions_api/streaming.rs` asserts `model_version` populates and roundtrips

Closes RIG-1315

### DIFF
--- a/crates/rig-core/src/providers/gemini/interactions_api/streaming.rs
+++ b/crates/rig-core/src/providers/gemini/interactions_api/streaming.rs
@@ -25,6 +25,11 @@ use serde_json::{Map, Value};
 pub struct StreamingCompletionResponse {
     pub usage: Option<InteractionUsage>,
     pub interaction: Option<Interaction>,
+    /// Resolved model identifier (e.g. `gemini-2.5-pro-preview-05-06`), extracted from
+    /// `Interaction.model`. The Interactions API has no `FinishReason` field; use
+    /// `interaction.status` for lifecycle state.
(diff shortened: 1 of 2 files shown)


gold-silver-copper:
LGTM, gonna add a live regression test as a follow up PR

PR #1778, opened 2026-05-16
feat(anthropic): support document citations

### LINKED ISSUE #1767: feat: support Anthropic document citations

## Feature Request

Rig's Anthropic provider supports document content blocks, but it does not currently expose Claude's Messages API citations surface.

Anthropic supports enabling citations on document blocks with:

```json
"citations": { "enabled": true }
```

When enabled, Claude returns citation metadata on response text blocks, including location types such as `char_location`, `page_location`, and `content_block_location`. Streaming responses can also include `citations_delta`.

Current Rig behavior in `rig-core` v0.37.0:

- `crates/rig-core/src/providers/anthropic/completion.rs` models `Content::Document` with only `source` and `cache_control`.
- `message::Document` has `additional_params`, but the Anthropic conversion currently ignores those provider-specific fields when building `Content::Document`.
- Anthropic response conversion maps text blocks into generic `AssistantContent::Text`, which has no place to preserve citation metadata.

Related prior work: #1358 added plain text document support for Anthropic, but explicitly did not enable citations.

### Motivation

Provider-native citations are useful for document QA, RAG, grading/review workflows, and any application that needs grounded model output with machine-readable links back to source documents.

Without this support, downstream users must bypass Rig and call Anthropic or Anthropic-compatible proxies directly to enable citations and parse the citation-bearing response shape. That loses Rig's normal provider abstraction, response conversion, telemetry, and usage handling.

### Proposal

Add Anthropic citation support for document content blocks and responses.

Suggested scope:

1. Add an Anthropic citation config type, e.g. `citations: { enabled: true }`, on supported Anthropic document content blocks.
2. Ensure generic `Document::additional_params` is either intentionally forwarded for Anthropic document fields or provide a typed Anthropic-specific builder/API for citations.
3. Preserve returned citation metadata on assistant text blocks instead of dropping it during conversion.
4. Support streaming `citations_delta` events if possible, or document that only non-streaming citations are supported initially.
5. Add serialization/deserialization tests covering:
   - request document serializes `citations.enabled = true`
   - response text block with `char_location` citation is preserved
   - PDF/page citation and custom content block citation shapes are represented or explicitly handled

This should stay scoped to the Anthropic provider and avoid changing other provider APIs to model fields they do not support.

### Alternatives

A downstream raw HTTP call works, but bypasses Rig's provider abstraction, response types, telemetry, and usage handling.

## Relevant source/docs:

  - https://platform.claude.com/docs/en/build-with-claude/citations

I'm happy to work on a PR for this

temrjan:
## What

Adds Anthropic document citations end-to-end:

- **Request:** `Content::Document` gains optional `title`, `context`, and typed `citations: CitationsConfig` fields. `TryFrom<message::Document>` forwards `additional_params` so users going through the generic surface can opt in.
- **Response:** `Content::Text` carries typed `citations: Vec<Citation>` with all three Anthropic location types (`char_location`, `page_location`, `content_block_location`). Conversion to `message::AssistantContent` preserves them via `message::Text.additional_params`.
- **Streaming:** new `ContentDelta::CitationsDelta` variant. Also added `#[serde(other)] Unknown` so any future delta type Anthropic adds does not break stream deserialization.
- **New public helper** `anthropic_citations(&message::Text) -> Result<Vec<Citation>, serde_json::Error>` for typed extraction from the generic `message::AssistantContent` surface.

## Why

The Messages API exposes citation metadata (RAG, document QA, auditable model output). Rig previously dropped these fields silently both on the request side (when set via `message::Document.additional_params`) and on the response side (the citation arrays returned on text blocks). This PR closes the gap without forcing other providers to model fields they don't support.

## Compatibility note

`message::Text` gains `additional_params: Option<serde_json::Value>` mirroring `Image`/`Audio`/`Video`/`Document` in the same module — `Text` was the only content type missing this. The change is non-breaking at the JSON wire level (default `None`, `#[serde(flatten, skip_serializing_if = "Option::is_none")]`) but downstream code using the struct literal `message::Text { text }` should migrate to `Text::new(text)` or add `additional_params: None`. A constructor is provided. Same generic mechanism would let future providers (logprobs, reasoning notes, ...) attach metadata to text blocks without a new abstraction per provider.

## Testing

Full workspace baseline green:

- `cargo build --workspace`
- `cargo test --workspace` — 610 tests pass; added 9 in `anthropic/completion.rs` (document serialization with citations, three location-type round-trips, helper extraction in both empty and populated paths, `AssistantContent` round-trip preserving citations, end-to-end `additional_params` forwarding on a request) and 2 in `anthropic/streaming.rs` (citations_delta event deserialization, unknown-delta fallback).
- `cargo clippy --workspace --all-targets --all-features -- -Dwarnings`
- `cargo +nightly fmt --all --check`
- `cargo +nightly doc --workspace --all-features --no-deps --document-private-items` with `RUSTDOCFLAGS="-D warnings"`

## Notes

- Streaming consumer currently drops `CitationsDelta` (returns `None` from `handle_event`); citations are accumulated per-content-block and returned with the assembled final `Content::Text`. Real-time citation streaming through `RawStreamingChoice` is out of scope here and would be a follow-up.
- Citation validation (the API requires all-or-none citations across documents in a request) is left to the server — adding client-side validation is scope creep.

## AI assistance disclosure

Implemented with a two-stage AI-assisted process:

1. **Claude Opus 4.7 (Anthropic)** wrote the implementation, tests, and documentation, following the project's idiomatic Rust style (`Result`-based error propagation, `#[derive(Default)]` on the extended `message::Text` so downstream struct-literal call-sites stay ergonomic).
2. **Kimi K2.6 (Moonshot)** ran two independent reviews:
   - **Pre-coding architectural audit** of the implementation plan — caught missing `impl From<String>`/`impl Display` updates and the real scope of ~22 call-sites that would break with the new `additional_params` field before any code was written.
   - **Final diff review** — caught compile failures in `rig-bedrock` and `rig-vertexai` not surfaced by `cargo check -p rig-core`.

I reviewed every line and take responsibility for correctness and maintainability. Full workspace baseline is green: `cargo build`, `test`, `clippy --all-targets --all-features -- -Dwarnings`, `+nightly fmt --check`, and `+nightly doc -D warnings`.

Fixes #1767

### DIFF
--- a/crates/rig-bedrock/src/types/assistant_content.rs
+++ b/crates/rig-bedrock/src/types/assistant_content.rs
@@ -126,7 +126,7 @@ impl TryFrom<aws_bedrock::ContentBlock> for RigAssistantContent {
     fn try_from(value: aws_bedrock::ContentBlock) -> Result<Self, Self::Error> {
         match value {
             aws_bedrock::ContentBlock::Text(text) => {
-                Ok(RigAssistantContent(AssistantContent::Text(Text { text })))
+                Ok(RigAssistantContent(AssistantContent::Text(Text::new(text))))
             }
             aws_bedrock::ContentBlock::ToolUse(call) => Ok(RigAssistantContent(
                 completion::AssistantContent::tool_call(
--- a/crates/rig-bedrock/src/types/completion_request.rs
+++ b/crates/rig-bedrock/src/types/completion_request.rs
@@ -243,9 +243,7 @@ mod tests {
             model: None,
             preamble: None,
             chat_history: OneOrMany::one(Message::User {
-                content: OneOrMany::one(UserContent::Text(Text {
-                    text: "test".to_string(),
-                })),
+                content: OneOrMany::one(UserContent::Text(Text::new("test".to_string()))),
             }),
             documents: vec![],
             tools: vec![],
@@ -508,9 +506,7 @@ mod tests {
             chat_history: OneOrMany::many(vec![
                 Message::system("History system instruction"),
                 Message::User {
-                    content: OneOrMany::one(UserContent::Text(Text {
-                        text: "test".to_string(),
-                    })),
+                    content: OneOrMany::one(UserContent::Text(Text::new("test".to_string()))),
                 },
             ])
             .expect("history should be non-empty"),
@@ -566,9 +562,7 @@ mod tests {
             chat_history: OneOrMany::many(vec![
                 Message::system("History system instruction"),
                 Message::User {
-                    content: OneOrMany::one(UserContent::Text(Text {
-                        text: "test".to_string(),
-                    })),
+                    content: OneOrMany::one(UserContent::Text(Text::new("test".to_string()))),
                 },
             ])
             .expect("history should be non-empty"),
@@ -607,9 +601,9 @@ mod tests {
         let request = CompletionRequest {
             chat_history: OneOrMany::many(vec![
                 Message::User {
-                    content: OneOrMany::one(UserContent::Text(Text {
-                        text: "user prompt".to_string(),
-                    })),
+                    content: OneOrMany::one(UserContent::Text(Text::new(
+                        "user prompt".to_string(),
+                    ))),
                 },
                 Message::Assistant {
                     id: None,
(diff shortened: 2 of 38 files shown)


fangkangmi:
Thanks @temrjan

PR #1781, opened 2026-05-17
fix: default OpenAI base64 image detail

### LINKED ISSUE #1780: bug: openai base64 image content rejects default image detail

## Bug Report

The OpenAI chat-completions converter rejects base64 image content when
`Image.detail` is `None`, even though `rig::completion::message::Image` exposes
`detail` as `Option<ImageDetail>` and `ImageDetail::default()` is `Auto`.

URL-backed images already treat `None` as `ImageDetail::Auto`, but base64-backed
images still return this conversion error:

```text
OpenAI image URI must have image detail
```

That makes a type-valid image fail only at request-conversion time.

## Reproduction

```rust
use rig::completion::message::{DocumentSourceKind, Image, ImageMediaType, UserContent};
use rig::providers::openai;

fn main() -> Result<(), Box<dyn std::error::Error>> {
    let image = Image {
        data: DocumentSourceKind::Base64("iVBORw0KGgo=".to_string()),
        media_type: Some(ImageMediaType::PNG),
        detail: None,
        additional_params: None,
    };

    let content = UserContent::Image(image);

    // This is the OpenAI provider conversion path used when building a chat
    // completions request from Rig message content.
    let result = openai::completion::UserContent::try_from(content);

    assert!(result.is_ok());
    Ok(())
}
```

Today the conversion fails with:

```text
OpenAI image URI must have image detail
```

The same `detail: None` value is accepted for URL-backed image content because
that path uses `detail.unwrap_or_default()`.

## Expected behavior

Base64 image content should behave the same as URL-backed image content:

```rust
detail.unwrap_or_default()
```

So `detail: None` should serialize as OpenAI image detail `auto`.

## Additional context

This came up while building a downstream multimodal grading flow. Our message
builder constructed image attachments with `detail: None`, which is valid at the
Rig type level, but OpenAI request conversion failed at runtime.

The local workaround is to always set:

```rust
detail: Some(ImageDetail::Auto)
```

But since `ImageDetail` already defaults to `Auto`, the OpenAI converter can
avoid this footgun by defaulting `None` in the base64 image path too.

fangkangmi:
## Description

Fixes #1780

OpenAI chat-completions conversion already defaulted URL-backed image detail with `unwrap_or_default()`, but base64-backed images returned a conversion error when `detail` was `None`. This changes the base64 image path to use the same `ImageDetail::Auto` default while still requiring `media_type` to build the data URI.

No new dependencies.

## Type of change

- [x] Bug fix

## Testing

- [x] `cargo fmt`
- [x] `cargo fmt --check`
- [x] `git diff --check`
- [x] `cargo test -p rig-core --lib providers::openai::completion::tests::base64_image_without_detail_defaults_to_auto`
- [x] `cargo test -p rig-core --lib providers::openai::completion::tests`
- [x] `cargo test -p rig-core --lib`

## Checklist

## Notes

AI assistance was used to locate the inconsistent converter branch and prepare this small patch. I reviewed the final diff and verified it locally.

### DIFF
--- a/crates/rig-core/src/providers/openai/completion/mod.rs
+++ b/crates/rig-core/src/providers/openai/completion/mod.rs
@@ -503,9 +503,7 @@ impl TryFrom<message::UserContent> for UserContent {
                         data
                     );
 
-                    let detail = detail.ok_or(message::MessageError::ConversionError(
-                        "OpenAI image URI must have image detail".into(),
-                    ))?;
+                    let detail = detail.unwrap_or_default();
 
                     Ok(UserContent::Image {
                         image_url: ImageUrl { url, detail },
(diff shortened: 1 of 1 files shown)


gold-silver-copper:
awesome PR thanks!

PR #1789, opened 2026-05-19
fix: stream ToolCallDelta in prompt_request

notV4l:


gold-silver-copper:
Thanks for catching this!

PR #1795, opened 2026-05-21
fix: metadata field does not handle null values (#1538)

### LINKED ISSUE #1538: bug: metadata field does not handle null values

tao501:
Fixes https://github.com/0xPlaygrounds/rig/issues/1538

The `metadata` field in `AdditionalParameters` was typed as `Map<String, Value>` without null handling. When OpenAI-compatible providers (e.g., vLLM) return `metadata: null` in the response, deserialization fails.

**Changes:**
- Changed `metadata` type to `Option<Map<String, Value>>` in responses_api/mod.rs
- Updated `.metadata.clear()` to `.metadata = None` in chatgpt/mod.rs

All existing tests pass (74 openai + 13 chatgpt).

### DIFF
--- a/crates/rig-core/src/providers/chatgpt/mod.rs
+++ b/crates/rig-core/src/providers/chatgpt/mod.rs
@@ -385,7 +385,7 @@ where
         }
 
         request.additional_parameters.background = None;
-        request.additional_parameters.metadata.clear();
+        request.additional_parameters.metadata = None;
         request.additional_parameters.parallel_tool_calls = None;
         request.additional_parameters.service_tier = None;
         request.additional_parameters.store = Some(false);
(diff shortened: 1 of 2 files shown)


gold-silver-copper:
Fixed in #1993

PR #1803, opened 2026-05-23
fix(tools): normalize null tool call arguments to empty object

### LINKED ISSUE #1802: fix(tools): ToolDyn::call fails when LLM sends null for all-optional tool arguments
## Summary

Several LLMs (e.g. Claude, GPT-4o) send `null` as the `arguments` field when invoking a tool whose parameters are all optional and none are needed. `serde_json::from_str::<T>("null")` fails for struct-typed args even when every field is `Option<_>`, because JSON `null` does not deserialize to an empty struct.

## Reproduction

```rust
#[derive(Deserialize)]
struct MyArgs {
    query: Option<String>,
}

// Fails: called with args = "null"
serde_json::from_str::<MyArgs>("null").unwrap();
// Error: invalid type: null, expected struct MyArgs
```

Any tool registered via `#[tool]` or `impl Tool` with an all-optional args struct will return a `ToolError` when the model omits all arguments.

## Expected behavior

`ToolDyn::call("null")` should succeed for any args type that also succeeds with `"{}"`—i.e. one whose fields are all optional.

## Notes

- `#1438` fixed `""` (empty string) for the OpenRouter-specific `stringified_json` deserializer — a different path.
- `#1786` (`null_or_default` in `json_utils.rs`) handles null in streaming tool-call deltas — also a different path.
- `ToolDyn::call` in `rig-core/src/tool/mod.rs` is not covered by either.

I have a fix ready.

jimmiebfulton:
## Summary

Normalizes `"null"` to `"{}"` in `ToolDyn::call` before deserialization so tools with all-optional argument fields work correctly when an LLM sends no arguments.

Several models (Claude, GPT-4o) send `null` as the `arguments` field when invoking a tool and all parameters are optional. `serde_json::from_str::<T>("null")` fails for struct-typed args even when every field is `Option<_>`, because JSON `null` does not deserialize to an empty struct.

## Changes

- **`rig-core/src/tool/mod.rs`** — normalize `args.trim() == "null"` → `"{}"` at the top of `ToolDyn::call`, before the `serde_json::from_str` call.
- Adds `null_args_are_normalized_to_empty_object` test using a struct with one `Option<String>` field.

## Related

- Complements #1438 (fixes `""` in OR `stringified_json` deserializer — different path).
- Complements #1786 (`null_or_default` for streaming deltas — different path).

## AI Assistance

This fix was developed with significant AI assistance (Claude). The bug was identified from real-world tool invocation failures, the fix and test were reviewed and validated by the author before submission.

## Issue

Fixes #1802

### DIFF
--- a/crates/rig-core/src/tool/mod.rs
+++ b/crates/rig-core/src/tool/mod.rs
@@ -200,6 +200,16 @@ impl<T: Tool> ToolDyn for T {
 
     fn call<'a>(&'a self, args: String) -> WasmBoxedFuture<'a, Result<String, ToolError>> {
         Box::pin(async move {
+            // LLMs frequently send `null` for tools whose arguments are all optional.
+            // `serde_json::from_str::<T>("null")` fails for struct types even when
+            // every field is `Option<_>`, because JSON null does not deserialize to an
+            // empty object. Normalise to `{}` so callers do not need to wrap their
+            // entire args type in `Option<T>` just to handle the no-argument case.
+            let args = if args.trim() == "null" {
+                "{}".to_string()
+            } else {
+                args
+            };
             match serde_json::from_str(&args) {
                 Ok(args) => <Self as Tool>::call(self, args)
                     .await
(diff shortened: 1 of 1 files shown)


gold-silver-copper:
Had to make a small fix and wasnt able to edit this PR for some reason (something to do with it being part of an organization perhaps), so I created stacked pr #1814 which builds on this pr, I'm going to do the same with the other PRs

PR #1810, opened 2026-05-23
feat(openrouter): add prompt-caching support for Anthropic models

### LINKED ISSUE #1809: feat(openrouter): add prompt-caching support for Anthropic models
## Summary

OpenRouter supports Anthropic-style prompt caching for Anthropic models: adding `cache_control: {"type": "ephemeral"}` to the system prompt tells Anthropic to cache that prefix at ~10% of the normal input-token cost. Subsequent turns sharing the same system prompt hit the cache and are billed at the reduced rate.

There is currently no way to enable this through the `rig` OpenRouter client.

## Proposed Solution

Add a `with_prompt_caching()` builder on `CompletionModel` that post-processes the outgoing request body to attach `cache_control` to the system prompt. Non-Anthropic backends routed through OpenRouter ignore the field, so it is safe to enable regardless of model.

```rust
let model = client
    .completion_model("anthropic/claude-3.5-sonnet")
    .with_prompt_caching();
```

I have an implementation ready.

jimmiebfulton:
## Summary

Adds opt-in Anthropic-style prompt caching for Anthropic models routed through OpenRouter. When enabled, the outgoing request body is post-processed to attach `cache_control: {"type": "ephemeral"}` to the system prompt so Anthropic caches the prefix at ~10% of the normal input-token cost. Non-Anthropic backends ignore the field gracefully.

## Changes

- **`apply_prompt_caching(body: &mut Value)`** — post-processes the serialized request body: string system content is converted to a structured text block with `cache_control`; array content has `cache_control` added to the last block (non-text blocks such as images are preserved unchanged).
- **`CompletionModel`** gains `prompt_caching: bool` and `with_prompt_caching()` builder.
- Wired into both the non-streaming `completion()` path (`completion.rs`) and the streaming `stream()` path (`streaming.rs`).
- Four tests: string system message, array system message (last-block only), non-text block preservation, and no-system-message no-op.

## AI Assistance

This implementation was developed with significant AI assistance (Claude). The design and tests were reviewed and validated by the author before submission.

## Issue

Fixes #1809

### DIFF
--- a/crates/rig-core/src/providers/openrouter/completion.rs
+++ b/crates/rig-core/src/providers/openrouter/completion.rs
@@ -1603,6 +1603,63 @@ pub enum ToolChoiceFunctionKind {
     Function { name: String },
 }
 
+/// Apply Anthropic-style prompt-caching markers to an already-serialized
+/// OpenRouter request body (automatic mode).
+///
+/// Finds the first system message in `messages` and converts its `content`
+/// to a structured text block with `cache_control: {"type": "ephemeral"}`.
+/// This tells Anthropic (via OpenRouter) to cache the system prompt so
+/// subsequent turns that share the same prefix are billed at the cache-hit rate.
+///
+/// Non-Anthropic models routed through OpenRouter typically ignore the
+/// `cache_control` field, so enabling this is safe regardless of the model.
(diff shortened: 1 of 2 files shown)


gold-silver-copper:
merged as #1832

PR #1812, opened 2026-05-23
feat(anthropic): add cache_control field to ToolDefinition

### LINKED ISSUE #1811: feat(anthropic): add cache_control field to ToolDefinition
## Summary

Anthropic's prompt-caching API supports up to 4 cache breakpoints per request across three layers: the system prompt, the tools array, and conversation history. The system prompt and message content layers are already supported via `SystemContent` and `Content` variants — both have `cache_control` fields.

The **tools layer** is missing: `ToolDefinition` has no `cache_control` field, so callers have no way to cache the tools array independently of the system prompt. Setting `cache_control` on the last `ToolDefinition` in the array is how Anthropic caches all tools up to that point.

## Proposed change

Add `pub cache_control: Option<CacheControl>` to `ToolDefinition`, following the exact same pattern already used by `SystemContent` and the message `Content` variants. The field is `skip_serializing_if = "Option::is_none"` so it is absent from requests that don't use caching.

I have an implementation ready.

jimmiebfulton:
## Summary

Anthropic supports caching the tools layer by setting `cache_control` on the last `ToolDefinition` in the array. `ToolDefinition` currently has no such field — this is the only cacheable layer not yet exposed in rig.

`SystemContent` and the message `Content` variants already have `cache_control`. This brings `ToolDefinition` in line with the same pattern.

## Changes

- Add `pub cache_control: Option<CacheControl>` to `ToolDefinition` with `#[serde(skip_serializing_if = "Option::is_none")]`.
- Initialize `cache_control: None` at the two `ToolDefinition` construction sites in the `TryFrom` impls (`completion.rs` and `streaming.rs`).

No behaviour change for existing callers — the field defaults to `None` and is omitted from serialization when unset.

## AI Assistance

This implementation was developed with significant AI assistance (Claude). The change and rationale were reviewed and validated by the author before submission.

## Issue

Fixes #1811

### DIFF
--- a/crates/rig-core/src/providers/anthropic/completion.rs
+++ b/crates/rig-core/src/providers/anthropic/completion.rs
@@ -150,6 +150,11 @@ pub struct ToolDefinition {
     pub name: String,
     pub description: Option<String>,
     pub input_schema: serde_json::Value,
+    /// Cache breakpoint marker. Set on the last tool in the array to cache
+    /// the tools layer independently of the system prompt. Anthropic accepts
+    /// up to 4 `cache_control` markers per request.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub cache_control: Option<CacheControl>,
 }
 
 /// TTL for a cache control breakpoint.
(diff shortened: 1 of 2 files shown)


gold-silver-copper:
Stacked this pr into #1815

PR #1813, opened 2026-05-24
Add ToolCallHookAction::Replace for mutating tool arguments

### LINKED ISSUE #1744: Feature request: add ToolCallHookAction::Replace { args } variant for arg-mutating hooks
## Summary

`ToolCallHookAction` (in `crates/rig-core/src/agent/prompt_request/hooks.rs`) currently has three variants:

- `Continue`
- `Skip { reason: String }`
- `Terminate { reason: String }`

There is no `Replace { args: serde_json::Value }` variant, which means a `PromptHook` implementation cannot intercept and rewrite a tool call's arguments mid-flight. Hooks can only let the call proceed unchanged, skip it, or terminate the loop.

This is a real limitation for any middleware that needs to mutate (not just gate) tool arguments before execution — for example, guardrail / policy / DLP libraries that want to redact PII, normalize parameters, or substitute scoped credentials before the tool runs.

## Proposed change

Add a fourth variant to `ToolCallHookAction`:

```rust
pub enum ToolCallHookAction {
    Continue,
    Skip { reason: String },
    Terminate { reason: String },
    Replace { args: serde_json::Value },   // ← new
}
```

When the agent loop processes a `Replace { args }` action, it would substitute the provided `args` for the original tool-call arguments and continue with the call.

Parikalp-Bhardwaj:
## Summary

This PR adds `ToolCallHookAction::Replace { args }`, allowing prompt hooks to modify tool-call arguments before the tool is executed.

This is useful for guardrails, DLP, policy checks, argument normalization, and cases where middleware needs to sanitize or adjust tool inputs instead of only allowing, skipping, or terminating the call.

## Changes

- Added `ToolCallHookAction::Replace { args: serde_json::Value }`
- Added a `ToolCallHookAction::replace(args)` helper
- Updated normal tool execution to use replaced arguments before calling the tool
- Updated streaming tool execution to use replaced arguments before calling the tool
- Passed the final tool arguments into `on_tool_result`

## Testing

Ran:

```bash
cargo fmt
cargo check
cargo test
```

Fixes #1744

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -145,6 +145,7 @@ When modifying hook behavior, preserve the intended control flow:
 - `ToolCallHookAction::Continue`
 - `ToolCallHookAction::Skip`
 - `ToolCallHookAction::Terminate`
+- `ToolCallHookAction::Replace`
 
 Check both streaming and non-streaming paths.
 
--- a/crates/rig-core/src/agent/prompt_request/hooks.rs
+++ b/crates/rig-core/src/agent/prompt_request/hooks.rs
@@ -8,6 +8,8 @@ use crate::{
     wasm_compat::{WasmCompatSend, WasmCompatSync},
 };
 
+use serde_json::Value;
+
 /// Trait for per-request hooks to observe tool call events.
 pub trait PromptHook<M>: Clone + WasmCompatSend + WasmCompatSync
 where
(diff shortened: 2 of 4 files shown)


gold-silver-copper:
This has been now implemented by #1963

PR #1819, opened 2026-05-25
feat(ollama): support tool_choice in completion requests

pablof7z:
## Summary

The Ollama `/api/chat` endpoint accepts `tool_choice` (`none`/`auto`/`required`) in versions that support tool calling. Previously rig warned and silently dropped this field:

```
WARN rig::providers::ollama: WARNING: `tool_choice` not supported for Ollama
```

This fired on every call through the `Extractor` API (which sets `tool_choice: Required` internally), meaning structured output extraction relied entirely on prompt instructions rather than API-level enforcement.

## Changes

- Adds a `ToolChoice` enum for the Ollama provider mapping rig's `message::ToolChoice` variants (`Auto` → `"auto"`, `None` → `"none"`, `Required` → `"required"`)
- Wires `tool_choice` through `OllamaCompletionRequest` serialization with `skip_serializing_if = "Option::is_none"` (no change to requests that don't set it)
- Removes the `tracing::warn!` that fired on every extractor call
- `Specific` variant returns a `CompletionError::ProviderError` (Ollama does not support named-function selection)
- Adds 3 unit tests covering `Required`, `Auto`, and absent `tool_choice` serialization

## Test plan

- [ ] `cargo test -p rig-core --lib -- providers::ollama::tests` — all 17 tests pass
- [ ] `cargo test --test ollama` — 7 tests pass (live-Ollama tests skipped without a running instance)

### DIFF
--- a/crates/rig-core/Cargo.toml
+++ b/crates/rig-core/Cargo.toml
@@ -15,7 +15,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 workspace = true
 
 [lib]
-name = "rig_core"
+name = "rig"
 path = "src/lib.rs"
 doctest = true
 
--- a/crates/rig-core/Cargo.toml
+++ b/crates/rig-core/Cargo.toml
@@ -15,7 +15,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 workspace = true
 
 [lib]
-name = "rig"
+name = "rig_core"
 path = "src/lib.rs"
 doctest = true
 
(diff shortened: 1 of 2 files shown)


gold-silver-copper:
Hey there, I'm generally for adding tool_choice support to ollama, but I can't find any good documentation that says it is supported for api/chat. I was able to find that /v1/chat/completions does mention it but seemingly does not support it. Various open issues in the ollama github also mention missing tool_choice support. If you could add a live ignored ollama test under `/tests`, that showcases proper usage of tool_choice with api/chat in this PR , that would be really helpful. Or if you have some other useful information, that'd be great. Thank you.

References:
https://docs.ollama.com/api/chat
https://docs.ollama.com/api/openai-compatibility
https://github.com/ollama/ollama-python/issues/663
https://github.com/ollama/ollama/issues/11171
 https://github.com/onyx-dot-app/onyx/pull/7156

PR #1822, opened 2026-05-27
fix(openai): tolerate object-form tool-call `arguments` in streaming

### LINKED ISSUE #1829: bug: OpenAI-compatible gateway streams tool-call `arguments` as object, breaking streaming tool calls

## Bug Report

**TL;DR:** When streaming, object-form `arguments` in a tool call (e.g. `{}` instead of `"{}"`) are silently dropped, and the turn ends with an empty assistant response.

When an OpenAI-compatible gateway streams `tool_calls[].function.arguments` as a JSON object (e.g. `{}`) instead of the spec-mandated JSON-encoded string (`"{}"`), the tool call never surfaces, and the turn ends with empty assistant text.

`StreamingFunction.arguments` is typed `Option<String>` (`crates/rig-core/src/providers/openai/completion/streaming.rs`, line 22), so serde fails with `invalid type: map, expected a string`. 
The SSE payload is deserialized into `StreamingCompletionChunk` inside `normalize_chunk` (~L167–172 of the same file). 
On a parse error, the code logs at `error` level and then returns `Ok(None)` — the error is not propagated to the caller:
```rust
let data = match serde_json::from_str::<StreamingCompletionChunk>(data) {
    Ok(data) => data,
    Err(error) => {
        tracing::error!(?error, message = data, "Failed to parse SSE message");
        return Ok(None); // chunk dropped, error NOT propagated
    }
};
```

So even though the failure is logged at `error` level, the caller receives no `Err`: the chunk is dropped, the tool call never surfaces, and the turn ends silently with empty assistant text.

**Affected:** 
`rig-core` `main` @ commit `a0cd8a8f5` (2026-05-27), provider: `openai` (streaming chat completions).

**Context:** 
Encountered on an internal gateway that exposes an OpenAI-compatible API while backing onto a Mistral model, so it goes through Rig's openai provider despite the underlying model being Mistral. 
The endpoint is private, and the exact gateway software is unknown, but the failure reproduces standalone from the raw SSE chunk below (no gateway access needed).

## Reproduction

The gateway emits:
```json
"function": { "name": "list_dir", "arguments": {} }
```

instead of the spec form:
```json
"function": { "name": "list_dir", "arguments": "{}" }
```

Minimal reproducing SSE chunk, deserializing this payload into `StreamingCompletionChunk` (the type `normalize_chunk` parses) triggers the failure:
```json
{"choices":[{"index":0,"delta":{"role":"assistant","content":"","tool_calls":[{"index":0,"id":"call_x","type":"function","function":{"name":"list_dir","arguments":{}}}]},"finish_reason":null}],"usage":null}
```

`StreamingCompletionChunk` is a crate-internal (module-private) type and cannot be deserialized from outside the crate, so the runnable artifact is the in-crate test `test_streaming_function_object_arguments` added in #1822.

## Expected behavior

The chunk should deserialize successfully, and the tool call should surface. 
The `arguments` field should be parsed tolerantly: a string is taken verbatim, and any other JSON value (object, array) is re-serialized to its compact JSON-string form. 
A malformed-but-recoverable `arguments` value should not cause the entire chunk, and thus the tool call to be silently discarded.

## Screenshots

N/A — reproducible from the SSE payload above.

## Additional context

Fix proposed in #1822: deserialize `arguments` tolerantly (string verbatim, any other JSON value re-serialized to a compact JSON string).

xavierforge:
Fixes #1829 

## Problem
Some OpenAI-compatible gateways stream `tool_calls[].function.arguments` as a JSON object (e.g. `"arguments": {}`) instead of the spec-mandated JSON-encoded string (`"arguments": "{}"`).

`StreamingFunction.arguments` is typed `Option<String>`, so serde fails with `invalid type: map, expected a string`. 
In `normalize_chunk` the parse error is only logged, and the whole SSE chunk is dropped (`Ok(None)`), so the tool call never surfaces, and the turn ends with empty assistant text, and then the caller sees a blank response, with no error unless `rig` debug logs are enabled.

Minimal reproducing chunk:
```json
{"choices":[{"index":0,"delta":{"role":"assistant","content":"","tool_calls":[{"index":0,"id":"call_x","type":"function","function":{"name":"list_dir","arguments":{}}}]},"finish_reason":null}],"usage":null}
```
## Fix
Deserialize `arguments` with a tolerant helper: a string is taken verbatim; any other JSON value is re-serialized to its compact JSON-string form. `null`/missing stays `None`. 
Fully backward compatible with the spec-compliant string form.

Moves the tolerant deserializer into a shared `json_utils::deserialize_json_string_or_value` (built on the existing `value_to_json_string`) and applies it to `StreamingFunction.arguments` via `#[serde(deserialize_with = …)]`, so other providers can reuse it. 
Adds unit tests in `json_utils` covering string / object / nested / null / missing.

## Tests
Adds `test_streaming_function_object_arguments` (empty + nested object) and `test_streaming_function_null_arguments` (null / missing). 
All existing streaming tests still pass.

### DIFF
--- a/crates/rig-core/src/providers/openai/completion/streaming.rs
+++ b/crates/rig-core/src/providers/openai/completion/streaming.rs
@@ -19,7 +19,31 @@ use crate::streaming;
 #[derive(Default, Deserialize, Debug)]
 pub(crate) struct StreamingFunction {
     pub(crate) name: Option<String>,
+    #[serde(default, deserialize_with = "de_tool_call_arguments")]
     pub(crate) arguments: Option<String>,
+}
+
+/// Deserialize `tool_calls[].function.arguments`, tolerating providers that send
+/// a JSON object/array instead of the spec-mandated JSON-encoded string.
+///
+/// The OpenAI Chat Completions spec defines `arguments` as a string (e.g.
+/// `"{\"city\":\"Paris\"}"`), and that is the only form rig produced before. Some
+/// OpenAI-compatible gateways instead emit a raw object (`"arguments": {}`), which
+/// previously failed deserialization with "invalid type: map, expected a string",
+/// causing the whole SSE chunk (and thus the tool call) to be silently dropped.
+///
+/// We now accept either: a string is taken verbatim; any other JSON value is
+/// re-serialized to its compact JSON-string form.
+fn de_tool_call_arguments<'de, D>(deserializer: D) -> Result<Option<String>, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    let value = Option::<serde_json::Value>::deserialize(deserializer)?;
+    Ok(match value {
+        None | Some(serde_json::Value::Null) => None,
+        Some(serde_json::Value::String(s)) => Some(s),
+        Some(other) => Some(other.to_string()),
+    })
 }
 
 #[derive(Deserialize, Debug)]
@@ -235,6 +259,32 @@ mod tests {
             function.arguments.as_ref().unwrap(),
             r#"{"location":"Paris"}"#
         );
+    }
+
+    #[test]
+    fn test_streaming_function_object_arguments() {
+        // Some OpenAI-compatible gateways send `arguments` as a JSON object
+        // instead of the spec-mandated JSON-encoded string. Accept it by
+        // re-serializing to the string form rather than dropping the chunk.
+        let json = r#"{"name": "list_dir", "arguments": {}}"#;
+        let function: StreamingFunction = serde_json::from_str(json).unwrap();
+        assert_eq!(function.name, Some("list_dir".to_string()));
+        assert_eq!(function.arguments.as_ref().unwrap(), "{}");
+
+        let json = r#"{"name": "get_weather", "arguments": {"city": "London"}}"#;
+        let function: StreamingFunction = serde_json::from_str(json).unwrap();
+        assert_eq!(function.arguments.as_ref().unwrap(), r#"{"city":"London"}"#);
+    }
+
+    #[test]
+    fn test_streaming_function_null_arguments() {
+        let json = r#"{"name": "list_dir", "arguments": null}"#;
+        let function: StreamingFunction = serde_json::from_str(json).unwrap();
+        assert!(function.arguments.is_none());
+
+        let json = r#"{"name": "list_dir"}"#;
+        let function: StreamingFunction = serde_json::from_str(json).unwrap();
+        assert!(function.arguments.is_none());
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


crates/rig-core/src/providers/openai/completion/streaming.rs
```diff
@@ -19,9 +19,33 @@ use crate::streaming;
...
+///
+/// The OpenAI Chat Completions spec defines `arguments` as a string (e.g.
+/// `"{\"city\":\"Paris\"}"`), and that is the only form rig produced before. Some
+/// OpenAI-compatible gateways instead emit a raw object (`"arguments": {}`), which
+/// previously failed deserialization with "invalid type: map, expected a string",
+/// causing the whole SSE chunk (and thus the tool call) to be silently dropped.
+///
+/// We now accept either: a string is taken verbatim; any other JSON value is
+/// re-serialized to its compact JSON-string form.
+fn de_tool_call_arguments<'de, D>(deserializer: D) -> Result<Option<String>, D::Error>
```
Shaurya-Sethi: This logic duplicates `json_utils::value_to_json_string` in `crates/rig-core/src/json_utils.rs`, which already does string vs non-string normalization:
```rust
pub fn value_to_json_string(value: &serde_json::Value) -> String {
    match value {
        serde_json::Value::String(s) => s.clone(),
        other => other.to_string(),
    }
}
```
We could reuse that (or add a small shared deserializer in `json_utils`) instead of maintaining a second copy here

Shaurya-Sethi (changes requested):
thanks for the clear write-up and focused fix.

I've verified the deserialization failure on `main` using the minimal `json` from your pr (object-form `"arguments": {}` fails with `invalid type: map, expected a string`), but haven't reproduced it against a live openai-compatible gateway.

**Before merge**
1. **Issue + real repro** : Please open an issue for this bug on the actual OpenAI-compatible gateway that emits object-form `arguments` in streaming (provider name, endpoint if relevant, and steps to reproduce). Link it from this PR (`Fixes #<n>`). 

The new unit tests are a good regression guard, but we will still need a reproducible report on the actual gateway. 

2. **Deduplication** : As noted inline on `de_tool_call_arguments`, please consider reusing the existing `json_utils::value_to_json_string`  helper.

Once you document and link the issue, a maintainer will likely follow-up with additional feedback before approving :)

crates/rig-core/src/providers/openai/completion/streaming.rs
xavierforge: Thanks for catching this. 
Moved it into `json_utils::deserialize_json_string_or_value` (on top of the existing `value_to_json_string`) and applied via `#[serde(deserialize_with = …)]`. Added unit tests in `json_utils`; the streaming tests stay as integration coverage.

xavierforge:
Thanks for taking the time to verify the repro. 
Opened #1829, it's an internal company gateway fronting a Mistral model (backing software unknown to me, endpoint private), but the raw SSE chunk in the issue reproduces the deserialization failure standalone.
Linked via Fixes #1829.

gold-silver-copper:
Thanks for the PR! And if you want, you can mention your company in the "who's using rig" section of our readme!

xavierforge:
Thanks! Happy to contribute, and thanks for the offer about the README!

PR #1828, opened 2026-05-27
Fix parsing of streamed function-call argument deltas

### LINKED ISSUE #1827: Streaming Responses function-call argument deltas can be silently dropped

## Bug Report

Rig can silently drop streamed OpenAI Responses API function-call argument events because the parser expects fields that are not present in the real wire shape.

The existing cassette `tests/cassettes/openai/streaming_tools/streaming_tools_smoke.yaml` contains real streamed function-call argument events without content_index, which get silently dropped.

### Expected behavior

Rig should parse streamed response.function_call_arguments.delta and response.function_call_arguments.done events using the real Responses API wire shape:

- item_id is on the outer event / ItemChunk
- content_index may be absent
- function-call argument deltas should emit ToolCallDelta events with the correct tool-call id

### Actual behavior

The events fail to deserialize and are silently skipped in the streaming path.

Because the completed tool call may still arrive later in response.output_item.done, end-to-end tool-call tests can still pass while incremental function-call argument deltas are missing.

### Proposed fix
https://github.com/0xPlaygrounds/rig/pull/1828

geraschenko:
Fixes https://github.com/0xPlaygrounds/rig/issues/1827, including adding a test demonstrating the issue using the existing cassette.

The real wire shape for response.function_call_arguments.delta and .done places item_id at the outer event level and may omit content_index. Rig’s parser expected those fields in the inner structs, causing these chunks to fail deserialization and be silently skipped.

The completed tool call can still arrive later via response.output_item.done, so existing end-to-end tool tests could pass while incremental ToolCallDelta events were missing.

### Changes

- Parse response.function_call_arguments.delta without requiring content_index.
- Parse response.function_call_arguments.done without requiring content_index.
- Use outer ItemChunk.item_id when emitting ToolCallDelta.
- Update the accumulator to receive the full ItemChunk.
- Add cassette coverage asserting recorded OpenAI streaming tool SSE events all parse as StreamingCompletionChunk.
- Add focused unit tests for real function-call argument wire shapes.

### Verification

```bash
cargo test -p rig --test openai \
  streaming_tools_smoke_cassette_sse_events_parse \
  -- --nocapture
```
This fails prior to the fix (second commit) like this:
```
thread 'openai::cassette::streaming_tools::streaming_tools_smoke_cassette_sse_events_parse' panicked at tests/providers/openai/cassette/streaming_tools.rs:82:5:
all cassette SSE data events should parse as StreamingCompletionChunk; failures:
interaction 0, body line 10: data did not match any variant of untagged enum StreamingCompletionChunk
{"delta":"{\"","item_id":"fc_REDACTED_1","obfuscation":"obfuscation_REDACTED_1","output_index":0,"sequence_number":3,"type":"response.function_call_arguments.delta"}
...
interaction 0, body line 37: data did not match any variant of untagged enum StreamingCompletionChunk
{"arguments":"{\"x\":2,\"y\":5}","item_id":"fc_REDACTED_1","output_index":0,"sequence_number":12,"type":"response.function_call_arguments.done"}
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
test openai::cassette::streaming_tools::streaming_tools_smoke_cassette_sse_events_parse ... FAILED
```
It passes after the second commit.

### DIFF
--- a/crates/rig-core/src/providers/copilot/mod.rs
+++ b/crates/rig-core/src/providers/copilot/mod.rs
@@ -1015,15 +1015,17 @@ where
                                         yield Ok(RawStreamingChoice::Message(delta.delta.clone()))
                                     }
                                     ItemChunkKind::FunctionCallArgsDelta(delta) => {
-                                        let internal_call_id = tool_call_internal_ids
-                                            .entry(delta.item_id.clone())
-                                            .or_insert_with(|| nanoid::nanoid!())
-                                            .clone();
-                                        yield Ok(RawStreamingChoice::ToolCallDelta {
-                                            id: delta.item_id.clone(),
-                                            internal_call_id,
-                                            content: streaming::ToolCallDeltaContent::Delta(delta.delta.clone())
-                                        })
+                                        if let Some(item_id) = chunk.item_id.as_ref() {
+                                            let internal_call_id = tool_call_internal_ids
+                                                .entry(item_id.clone())
+                                                .or_insert_with(|| nanoid::nanoid!())
+                                                .clone();
+                                            yield Ok(RawStreamingChoice::ToolCallDelta {
+                                                id: item_id.clone(),
+                                                internal_call_id,
+                                                content: streaming::ToolCallDeltaContent::Delta(delta.delta.clone())
+                                            })
+                                        }
                                     }
                                     _ => continue,
                                 }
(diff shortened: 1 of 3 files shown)


gold-silver-copper:
Awesome PR, THANK YOU

PR #1830, opened 2026-05-28
Preserve OpenAI Responses prompt cache params

Kade-Powell:
Adds OpenAI Responses API support for passing prompt_cache_key and prompt_cache_retention through additional_params.

Includes a focused regression test.


kernitus:
I’m working on moving VTCode to use Rig’s OpenAI Responses request types more directly. Having `prompt_cache_key` and `prompt_cache_retention` modelled here removes one of the local VTCode overlays we currently need. For coding-assistant workloads, these cache controls matter because long tool and repository contexts are expensive and need stable cache routing and retention to work well, as outlined in https://openai.com/index/unrolling-the-codex-agent-loop/

gold-silver-copper:
Thank you!

PR #1837, opened 2026-05-29
fix(rig-core): Expose tools added via ToolServerHandle::append_toolset

### LINKED ISSUE #1836: bug: `ToolServerHandle::append_toolset` silently hides tools from the LLM

## Bug Report

Tools added via `ToolServerHandle::append_toolset` are registered in the executable toolset but never added to static_tool_names, which get_tool_defs enumerates. The LLM never sees them. The single-tool path, `ToolServerHandle::add_tool`, updates both correctly, however.

## Reproduction

```rust
use rig_core::test_utils::MockAddTool;
use rig_core::tool::{ToolSet, server::ToolServer};

#[tokio::test]
async fn append_toolset_makes_tools_visible() {
    let handle = ToolServer::new().run();

    let mut toolset = ToolSet::default();
    toolset.add_tool(MockAddTool);
    handle.append_toolset(toolset).await.unwrap();

    let defs = handle.get_tool_defs(None).await.unwrap();
    assert_eq!(defs.len(), 1); // defs is empty
}
```

## Expected behavior

`ToolServerHandle::append_toolset` and serial `ToolServerHandle::add_tool` calls with the same tools in the same order should have the same result.

## Additional context

Seems that this wasn't caught as there are no callers of the function across the codebase. It was introduced originally in #916 and not addressed in the refactor in #1607

mccormickt:
Tools added via ToolServerHandle::append_toolset were merged into the internal ToolSet but absent from static_tool_names, so they wouldnt' be returned by `get_tool_defs` and the LLM would never see them.

Fixes #1836

### DIFF
--- a/crates/rig-core/src/tool/server.rs
+++ b/crates/rig-core/src/tool/server.rs
@@ -121,9 +121,14 @@ impl ToolServerHandle {
         Ok(())
     }
 
-    /// Merge an entire toolset into the server.
+    /// Merge an entire toolset into the server. Tool names from `toolset`
+    /// are appended to the static-tool list, so the tools become visible
+    /// to the LLM via [`Self::get_tool_defs`].
     pub async fn append_toolset(&self, toolset: ToolSet) -> Result<(), ToolServerError> {
         let mut state = self.0.write().await;
+        state
+            .static_tool_names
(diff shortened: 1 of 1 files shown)


gold-silver-copper:
Thank you for the fix!

PR #1855, opened 2026-06-02
feat(openai-responses): preserve unknown Output payloads

### LINKED ISSUE #1861: OpenAI Responses Output::Unknown discards unrecognized output-item payloads at decode
## Context

Building observability/telemetry over the OpenAI Responses API, I need to see
provider-native hosted-tool items (`web_search_call`, `file_search_call`,
`computer_use_call`, `code_interpreter_call`) that come back in the response
`output[]` array. These tools execute inside the provider and never round-trip
the agent tool loop, so `PromptHook::on_tool_call` never fires for them. The
typed `Output` decode is the only place their data can surface.

## Observed

`Output::Unknown` is a fieldless unit variant fed by `#[serde(other)]`
(`crates/rig-core/src/providers/openai/responses_api/mod.rs`). Any output item
whose `type` tag isn't one of the modeled variants decodes to a payloadless
`Unknown` — the raw item object is discarded at decode. Every hosted-tool
result is dropped on the floor, in both the non-streaming response and the
streaming path (`OutputItemDone(StreamingItemDoneOutput { item: Output })`).

## Expected

The catch-all should keep the bytes, not just swallow the variant. The
`#[serde(other)] Unknown` exists for a real invariant — an unknown item type
must not fail deserialization of the whole `CompletionResponse` (the regression
that previously caused streaming token usage to be silently dropped). That
invariant is satisfied just as well by *preserving* the payload as by discarding
it. "Don't break decode" was implemented as "discard the bytes" when "keep the
bytes" meets the same bar and unblocks downstream inspection.

## Repro

```rust
use rig::providers::openai::responses_api::Output;
use serde_json::json;

let item = json!({
    "type": "web_search_call",
    "id": "ws_001",
    "status": "completed",
    "action": { "type": "search", "queries": ["rig framework"] },
});

let output: Output = serde_json::from_value(item).unwrap();
// output is Output::Unknown with no data — the item object is gone.
```

## Hypothesis

`#[serde(other)]` only applies to a unit variant, so `Unknown` cannot both be
the catch-all and carry data under the derive. The fix is to hand-write
`Output`'s (de)serialization, dispatching the known tags through a private
mirror enum and capturing any other tag as `Unknown(serde_json::Value)`. Known
variants keep their internally tagged (`type`) wire shape byte-for-byte; a known
tag with a malformed body still errors rather than silently degrading to
`Unknown`.

## Scope

- In scope: stop discarding unrecognized `Output` item payloads; preserve them
  verbatim as `Value`.
- Out of scope: typed modeling of specific hosted tools (can land later on top
  of the preserved `Value`), and surfacing hosted tools through `on_tool_call`
  (a separate seam in the agent loop).

Tracked by #1855.

ForeverAngry:
Fixes #1861

## Summary

The OpenAI Responses `Output::Unknown` variant was fieldless, so unrecognized
output items — including all provider-native hosted tools — were discarded at
decode. This makes `Unknown` carry the raw `serde_json::Value` so the payload
survives.

## Motivation

Hosted tools (`web_search_call`, `file_search_call`, `computer_use_call`,
`code_interpreter_call`) execute inside the provider and never round-trip the
agent tool loop, so `PromptHook::on_tool_call` never fires for them. The typed
`Output` decode was the only place their data could surface — and it dropped it.

The `#[serde(other)] Unknown` catch-all exists for a real reason: keep an
unknown item type from failing deserialization of the whole
`CompletionResponse` (which previously caused streaming token usage to be
silently dropped). That invariant is preserved here. The bug is that "don't
break decode" was implemented as "discard the bytes," when "keep the bytes"
satisfies the same invariant and unblocks observability.

## Design

`#[serde(other)]` only applies to a unit variant, so `Unknown` cannot both be
the catch-all and carry data. The (de)serialization is therefore hand-written
and dispatches through a private `KnownOutput` mirror of the modeled variants:

- Deserialize: decode to `Value`, read the `type` tag. Known tag → decode via
  `KnownOutput` (a malformed body still errors). Any other tag →
  `Unknown(value)`.
- Serialize: known variants round-trip through `KnownOutput` (wire shape is
  byte-for-byte identical); `Unknown(value)` emits the raw value.

Only one enum changes. Hosted-tool items reach `Output` in both the
non-streaming response and the streaming path (via
`OutputItemDone(StreamingItemDoneOutput { item: Output })`), so preserving
`Output` covers both. `ItemChunkKind::Unknown` is intentionally left as-is —
it catches unknown streaming *event wrappers*, not output items.

Alternatives rejected:
- `#[serde(untagged)]` with an `Unknown(Value)` tail — flips the enum to
  untagged, making a malformed *known* variant silently fall through to
  `Unknown` instead of erroring. Loses strictness for no gain.
- Typed variants per hosted tool — couples rig-core to OpenAI's hosted-tool
  churn. The catch-all's whole point is to not track that surface; keep it
  open, just stop discarding it.

## Public-API delta

Breaking. `Output::Unknown` → `Output::Unknown(serde_json::Value)`. The enum
also gains `#[non_exhaustive]` (the doc already called it "currently
non-exhaustive"; this backs the prose so the next unknown-variant addition
isn't itself breaking). Downstream exhaustive matches on `Output` need the
one-token `Unknown(_)` update; since the variant was fieldless, no caller can
be reading data out of it today.

## Tests

- `web_search_call` item decodes to `Unknown(value)` holding the full object.
- Round-trip is byte-stable through serialize/deserialize.
- A known tag (`function_call`) with a malformed body still errors.
- A `CompletionResponse` mixing an unknown item with a real message decodes,
  and token usage is intact (guards the original `#[serde(other)]` reason).

## Out of scope

- Typed modeling of specific hosted tools — can land later on top of the
  preserved `Value`.
- Surfacing hosted tools through `on_tool_call` — a separate seam in the agent
  loop; this only stops rig-core from destroying the data that seam needs.

### DIFF
--- a/crates/rig-core/src/providers/copilot/mod.rs
+++ b/crates/rig-core/src/providers/copilot/mod.rs
@@ -1003,7 +1003,7 @@ where
                                         StreamingItemDoneOutput { item: responses_api::Output::Message(msg), .. } => {
                                             yield Ok(RawStreamingChoice::MessageId(msg.id.clone()));
                                         }
-                                        StreamingItemDoneOutput { item: responses_api::Output::Unknown, .. } => {}
+                                        StreamingItemDoneOutput { item: responses_api::Output::Unknown(_), .. } => {}
                                     },
                                     ItemChunkKind::OutputTextDelta(delta) => {
                                         yield Ok(RawStreamingChoice::Message(delta.delta.clone()))
--- a/crates/rig-core/src/providers/openai/responses_api/mod.rs
+++ b/crates/rig-core/src/providers/openai/responses_api/mod.rs
@@ -1277,10 +1277,41 @@ pub enum Include {
 }
 
 /// A currently non-exhaustive list of output types.
-#[derive(Clone, Debug, Deserialize, Serialize, PartialEq)]
-#[serde(tag = "type")]
-#[serde(rename_all = "snake_case")]
+///
+/// Unrecognized output items — notably provider-native hosted tools such as
+/// `web_search_call`, `file_search_call`, `computer_use_call`, and
+/// `code_interpreter_call` — decode to [`Output::Unknown`], which preserves
+/// the verbatim item object so callers can inspect or forward it. This keeps
+/// unknown item types from breaking deserialization of the entire
+/// `CompletionResponse` (the invariant that previously caused streaming token
+/// usage to be silently dropped) without discarding the payload along the way.
+#[derive(Clone, Debug, PartialEq)]
+#[non_exhaustive]
 pub enum Output {
+    Message(OutputMessage),
+    FunctionCall(OutputFunctionCall),
+    Reasoning {
+        id: String,
+        summary: Vec<ReasoningSummary>,
+        encrypted_content: Option<String>,
+        status: Option<ToolStatus>,
+    },
+    /// Catch-all for output item types this version does not model. Holds the
+    /// raw item object exactly as it appeared in the provider's `output[]`
+    /// array, so hosted-tool payloads survive the typed decode.
+    Unknown(Value),
+}
+
+/// Serde mirror of the modeled [`Output`] variants.
+///
+/// `Output`'s (de)serialization is hand-written so [`Output::Unknown`] can
+/// carry a raw [`Value`]; `#[serde(other)]` only applies to a unit variant and
+/// would otherwise force the payload to be dropped. Both impls dispatch through
+/// this helper so the internally tagged (`type`) wire shape of the known
+/// variants stays byte-for-byte identical to the previous derived behavior.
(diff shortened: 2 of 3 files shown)


kernitus:
I’m working on moving VTCode to rely on Rig’s Responses support as much as possible. This change is important cause coding assistants can receive provider-hosted or custom output items that Rig may not model yet, and dropping or rejecting those payloads makes valid streams fail or lose tool and state data. Preserving unknown `Output` payloads lets downstreams use Rig on the happy path while still handling newer Responses item shapes safely.

I think the pr needs to be updated off of the main branch and the comments resolved

gold-silver-copper:
Merged as #1950

PR #1856, opened 2026-06-03
feat(vector-store): Add Pinecone vector store integration

### LINKED ISSUE #37: feat: Add support for Pinecone vector store
Pinecone is a popular cloud based, fully managed vector database designed specifically for RAG applications.

### Resources
The Pinecone API provides a user-friendly way to [insert](https://docs.pinecone.io/reference/api/2024-07/data-plane/upsert) and [query](https://docs.pinecone.io/reference/api/2024-07/data-plane/query) embeddings. This should make the integration pretty straighforward. 

Note that users will require a Pinecone API key to be able to use Pinecone. Moreover, the responsibility of creating a collection and an index on that collection is outside the scope of Rig (the integration should assume that the user has already setup Pinecone correctly).

Interestingly, Pinecone is a pure vector store database, in the sense that it does not allow users to store documents, only vectors (and optional metadata for filtering). Performing vector search only returns the IDs of the closest vectors and not full documents.

Another interesting observation is that Pinecone has an endpoint to generate embeddings. This means that Pinecone could potentially also be integrated as an embedding model provider.

sosal123tyu1:
This PR adds Pinecone vector store integration for Rig.

It implements `PineconeVectorStore` supporting:
- Document insertions with UUID generation and custom document metadata framing
- Vector search via `top_n` and `top_n_ids`
- Filter translation (Eq, Gt, Lt, And, Or) mapping to Pinecone query syntax

An integration test suite has been added under `crates/rig-pinecone/tests/pinecone_tests.rs` using `httpmock` to verify the end-to-end integration and translation without needing real API keys.

Closes #37

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8982,6 +8982,7 @@ dependencies = [
  "rig-milvus",
  "rig-mongodb",
  "rig-neo4j",
+ "rig-pinecone",
  "rig-postgres",
  "rig-qdrant",
  "rig-s3vectors",
@@ -9234,6 +9235,21 @@ dependencies = [
  "tokio",
  "tracing",
  "tracing-subscriber",
+]
+
+[[package]]
+name = "rig-pinecone"
+version = "0.38.1"
+dependencies = [
+ "anyhow",
+ "httpmock",
+ "reqwest 0.13.3",
+ "rig-core",
+ "serde",
+ "serde_json",
+ "thiserror 2.0.18",
+ "tokio",
+ "uuid",
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


gold-silver-copper:
Hello, recently RIg has been taking a stronger stance in regards to external integrations. It is a lot of work for us to verify the proper functioning of each new integration, and more so to maintain it. From now on external integrations are best implemented as personal projects. We will be creating an extra repository for keeping track of such external integrations, so you can submit yours there: https://github.com/0xPlaygrounds/awesome-rig

PR #1857, opened 2026-06-03
fix(openai): make token usage details optional in responses API

### LINKED ISSUE #1569: I can haz `mistralrs` provider?
## Background

> WRT mistral: We have our own provider integration specifically for it, you shouldn't need to be using the OpenAI Chat Completions API. It looks like you are? I'm a bit confused here as to what you are doing

*Originally posted by @joshua-mo-143 in *[*#1179*](<https://github.com/0xPlaygrounds/rig/issues/1179#issuecomment-3673016743>)

## Intro

**Hi there! Great project. Just a few questions/thoughts...**

The OP on [RIG-1103](https://linear.app/playgrounds-ai/issue/RIG-1103/featbug-geminimistralrs-dynamic-context-is-ignoredleads-to-an-error) (@christoph-wulf) mentioned `mistral.rs`, and I'd assert he meant [this `mistral.rs`](<https://github.com/EricLBuehler/mistral.rs>) / [mistralrs-cli](<https://crates.io/crates/mistralrs-cli>) **not** [Mistral AI](<https://mistral.ai/>) (?).

However, @joshua-mo-143 responded with the existing `mistral` provider... which was actually helpful, but first let me start with the step before that...

## Trying the `openai` provider with a custom URL pointing to a `mistralrs serve` instance

I, too, have been trying to make `rig-core` work with [mistral.rs](<http://mistral.rs>), which is a similar setup as ollama, except default endpoint should be `http://localhost:1234/v1`. @christoph-wulf also mentioned [mistral.rs](<http://mistral.rs>)' [http server providing an "OpenAI-compatible" API](<https://github.com/EricLBuehler/mistral.rs/blob/master/docs/HTTP.md>), but it's not compatible enough apparently that the `openai` provider works out of the box simply by providing a custom URL.

The error I see when trying that with a modified version (diff below) of [the agent example](<https://github.com/0xPlaygrounds/rig/blob/main/rig/rig-core/examples/agent.rs>) is:

```text
Error: CompletionError: JsonError: missing field `output_tokens_details` at line 1 column 3411

Caused by:
    0: JsonError: missing field `output_tokens_details` at line 1 column 3411
    1: missing field `output_tokens_details` at line 1 column 3411
```

```diff
-    let client = openai::Client::from_env();
+    let client = openai::Client::builder()
+        .api_key("")
+        .base_url("http://localhost:1234/v1")
+        .build()?;
```

## Trying the `mistral` provider with a custom URL pointing to a `mistralrs serve` instance

Surprisingly this worked! I don't know why/how... (perhaps [mistral.rs](<http://mistral.rs>)' protocol is close enough to mistral's? no idea...) but I'm able to call both `prompt` and `stream_prompt` using the same `agent` example... it sends the request to [mistral.rs](<http://mistral.rs>) and receives the response. However, the `mistral` provider suffers from a couple deficiencies, at least that I ran into:

1. No real streaming capability
   * Can call `stream_prompt` on the `agent`, but it doesn't really stream
2. No support for reasoning/thinking:
   * Appears to be neutered?

## Summary

Why use [mistral.rs](<http://mistral.rs>)? It's faster than ollama and it's Rust. It runs locally. True, we could just use [mistral.rs](<http://mistral.rs>) directly... but I believe that use case generally would involve loading the model on each run rather than client/server and afaict its agent, integrations, etc capabilities aren't anything like [rig.rs](<http://rig.rs>)'.

I just wanted to provide more details/background on this to help the [rig.rs](<http://rig.rs>) project understand the idea: want to "fork" or combine the `mistral` provider (because it works but flesh out the streaming and reasoning mechanics) and/or the ollama or openai providers (because they have streaming/reasoning) to provide a working out-of-the-box UX using [rig.rs](<http://rig.rs>) with [mistral.rs](<http://mistral.rs>). I've personally put a few hours into it and found it to be a relatively non-trivial prospect but I'm a noob to [rig.rs](<http://rig.rs>) source... perhaps another community or core member might be able to provide additional info, assist, etc?

Thanks, again @joshua-mo-143 et al for all your work. Love [rig.rs](<http://rig.rs>) so far.

sosal123tyu1:
This PR resolves issue #1569 where using the `openai` provider with compatible backends (such as `mistral.rs` or `llama.cpp`) results in a deserialization error due to missing token usage details.

Changes:
- Made `output_tokens_details` in `ResponsesUsage` optional (`Option<OutputTokensDetails>`) and marked it with `#[serde(skip_serializing_if = "Option::is_none")]`.
- Updated usage extraction to gracefully handle missing values (falling back to default/0).
- Updated existing test cases to wrap test token usage structures accordingly.

Closes #1569

### DIFF
--- a/crates/rig-core/src/providers/openai/responses_api/mod.rs
+++ b/crates/rig-core/src/providers/openai/responses_api/mod.rs
@@ -709,21 +709,19 @@ impl Add for ResponsesUsage {
 
     fn add(self, rhs: Self) -> Self::Output {
         let input_tokens = self.input_tokens + rhs.input_tokens;
-        let input_tokens_details = self.input_tokens_details.map(|lhs| {
-            if let Some(tokens) = rhs.input_tokens_details {
-                lhs + tokens
-            } else {
-                lhs
-            }
-        });
+        let input_tokens_details = match (self.input_tokens_details, rhs.input_tokens_details) {
+            (Some(lhs), Some(rhs)) => Some(lhs + rhs),
+            (Some(lhs), None) => Some(lhs),
+            (None, Some(rhs)) => Some(rhs),
+            (None, None) => None,
+        };
         let output_tokens = self.output_tokens + rhs.output_tokens;
-        let output_tokens_details = self.output_tokens_details.map(|lhs| {
-            if let Some(tokens) = rhs.output_tokens_details {
-                lhs + tokens
-            } else {
-                lhs
-            }
-        });
+        let output_tokens_details = match (self.output_tokens_details, rhs.output_tokens_details) {
+            (Some(lhs), Some(rhs)) => Some(lhs + rhs),
+            (Some(lhs), None) => Some(lhs),
+            (None, Some(rhs)) => Some(rhs),
+            (None, None) => None,
+        };
         let total_tokens = self.total_tokens + rhs.total_tokens;
         Self {
             input_tokens,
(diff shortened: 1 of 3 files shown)


gold-silver-copper:
Отличная работа! Спасибо!

sosal123tyu1:
Hi! Great to see this PR merged and used in the project. I'm reaching out to follow up on the bounty payout procedure for this contribution. Could you please let me know if there's a specific form I need to fill out, or if you need me to provide a wallet address directly here? Thanks!

sosal123tyu1:
Also, just to clarify, I submitted this contribution because I'm genuinely interested in the project. If there's no bounty associated with this specific PR, that's completely fine. Keep up the great work!

PR #1858, opened 2026-06-03
feat(agent): add `push_notification` to inject messages during multi-turn prompts

lazytanuki:
Hi!

This PR adds a `push_notification` method to `Agent` that allows injecting user messages into an already-running multi-turn prompt, similar to how OpenCode makes it possible to send input to an active agent mid-conversation.

## How it works

- `Agent::push_notification(conversation_id, message)` enqueues a message into a per-conversation pending queue.
- At the start of each loop iteration (both streaming and non-streaming), the queue is drained and messages are injected as `User` messages into the conversation history.
- The queue is stored behind `Arc<Mutex<...>>`, so cloning the agent shares the same queue, enabling cross-task patterns like spawning a background watcher that calls `push_notification` when an external event occurs.

## Example

```rust
let agent = openai::Client::from_env()?
    .agent(openai::GPT_5_2)
    .conversation_id("my-conv")
    .toot(some_tool)
    .build();
let agent_clone = agent.clone();

let prompt_task = tokio::spawn(async move {
    // Start a long running multi-turn prompt.
    let resp = agent.prompt("Do this and that with your available tools.").await.unwrap();
    println!("{resp}");
})

let some_watcher_task = tokio::spawn(async move {
    // ... wait for some condition ...
    agent_clone.push_notification("my-conv", "Continue, but you should know that ...".to_string());
});

// Wait on both tasks...
```

## Disclosure

- Feature code was handwritten.
- Tests were AI-generated then checked.

Thanks!

### DIFF
--- a/crates/rig-core/src/agent/builder.rs
+++ b/crates/rig-core/src/agent/builder.rs
@@ -1,4 +1,7 @@
-use std::{collections::HashMap, sync::Arc};
+use std::{
+    collections::HashMap,
+    sync::{Arc, Mutex},
+};
 
 use schemars::{JsonSchema, Schema, schema_for};
 
@@ -533,6 +536,7 @@ where
             output_schema: self.output_schema,
             memory: self.memory,
             default_conversation_id: self.default_conversation_id,
+            pending_notifications: Arc::new(Mutex::new(HashMap::new())),
         }
     }
 }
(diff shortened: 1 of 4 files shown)


gold-silver-copper:
This should be possible to do already through hooks or the ecs runtime

PR #1859, opened 2026-06-03
feat(rig-core): expose provider error response inspection

### LINKED ISSUE #1820: feat: Expose errors returned by the providers
Any plans to add better support for error handling?

At the moment if there's an error returned by a provider it's exposed just as a string.

Example:
```
ProviderError: Invalid status code 400 Bad Request with message: {
  "error": {
    "message": "The image data you provided does not represent a valid image. Please check your input and try again with one of the supported image formats: ['image/jpeg', 'image/png', 'image/gif', 'image/webp'].",
    "type": "invalid_request_error",
    "param": "input",
    "code": "invalid_value"
  }
}
```

I would like to be able to dig deeper and have access to the error.
Even the raw response from the provider if it would be returned it would be helpful. I could JSON parse it.

Shaurya-Sethi:
Fixes #1820

## Summary

This PR adds provider error response inspection across Rig capability errors.

Users can read a preserved provider response body, parse it as JSON, and get the HTTP status when Rig captured one, through uniform helpers on capability errors:

- `provider_response_body`
- `provider_response_json`
- `provider_response_status`

## Approach

Preserved provider responses are stored separately from `ProviderError(String)` in the type system:

- A shared `ProviderResponseError { status: Option<StatusCode>, body: String }` type, re-exported at the crate root.
- A `ProviderResponse(ProviderResponseError)` variant on `CompletionError`, `EmbeddingError`, `TranscriptionError`, `ImageGenerationError`, `AudioGenerationError`, and `VerifyError`.

The helpers only report data from two sources that genuinely represent a provider response:

- `ProviderResponse`, for provider-authored API error envelopes.
- `HttpError(InvalidStatusCodeWithMessage(status, body))`, for non-success HTTP responses.

`ProviderError(String)` is never treated as a provider response body because it also carries Rig-generated and internal diagnostics. The helpers return `None` for it.

`PromptError` and `StructuredOutputError` forward these helpers when they wrap a completion failure.

## What Improved

- Uniform helpers across rig-core completion, prompt, structured output, embedding, transcription, image generation, audio generation, and verify errors.
- 2xx API error envelopes are preserved as `ProviderResponse` with captured status and raw body for the touched OpenAI-family and OpenAI-compatible providers.
- Paths that previously flattened `"{status}: {body}"` into `ProviderError` now use `HttpError(InvalidStatusCodeWithMessage(status, body))` in the touched providers, so `provider_response_status()` remains recoverable.
- OpenAI Responses API non-success responses, the default OpenAI completion path, now preserve inspectable status and body.
- `verify()` non-success arms preserve status and body via `InvalidStatusCodeWithMessage`; 401/403 still map to `InvalidAuthentication`.
- OpenAI Chat Completions-compatible streams and OpenAI Responses streaming now surface HTTP transport failures and in-band provider error events with inspectable raw bodies.
- Multipart-capable `RecordingHttpClient` coverage makes transcription provider response branches testable.

## Notes And Limits

- `AudioGenerationError::ProviderResponse` is forward-looking for symmetry. No audio provider path currently constructs it. Only Azure audio failures currently surface as inspectable `HttpError`; other audio providers still return `ProviderError` and remain follow-up work.
- Some providers and transports still have bespoke error envelopes or event formats that should be handled separately: Anthropic, Gemini, Cohere, xAI Responses-style envelopes, Mira, provider-specific SSE event payloads, WebSocket transports, Ollama streaming, and companion `rig-*` provider crates.
- `provider_response_status()` can return a 2xx status for provider-authored error envelopes sent over successful HTTP responses.

## Breaking Changes

- The new `ProviderResponse` variant breaks downstream exhaustive matches on `CompletionError`, `EmbeddingError`, and `VerifyError`. These enums, along with `ImageGenerationError` and `AudioGenerationError`, are now `#[non_exhaustive]`, matching `TranscriptionError`.
- Provider paths that previously returned `ProviderError` for preserved bodies or non-success statuses now return `ProviderResponse` or `HttpError`, which changes which variant downstream `match` arms observe for those failures.
- Streaming HTTP failures now surface as `CompletionError::HttpError` instead of `ProviderError` with a formatted string.

## Tests

- Non-2xx provider responses via `RecordingHttpClient::with_error_response(status, raw_json_body)` for rewritten branches that receive a response object.
- 2xx error envelope coverage retained as a secondary case, now asserting `ProviderResponse` and captured status.
- End-to-end `verify()` failure through `RecordingHttpClient`.
- OpenAI Responses parser, buffered stream, and live stream error-event coverage.
- OpenAI Chat Completions-compatible in-band streaming error coverage.
- Transcription non-success paths through multipart-capable `RecordingHttpClient`.
- Unit tests assert that `ProviderError(String)` and Rig-generated stream transport diagnostics are not reported as provider response bodies.

## Test Plan

- [x] `cargo fmt`
- [x] `cargo clippy -p rig-core --all-targets --all-features`
- [x] `cargo test -p rig-core --all-features` (4 pre-existing epub loader failures unrelated to this change)
- [x] `cargo check --workspace --all-targets`
- [x] `cargo doc -p rig-core --no-deps --all-features`

Latest focused checks after final review cleanup:

- [x] `cargo test -p rig-core provider_response`
- [x] `cargo test -p rig-core streaming_error_event`
- [x] `cargo test -p rig-core test_completion_response_from_sse_body_falls_back_to_streamed_text`
- [x] `cargo test -p rig-core transcription_http_non_success`

## AI Disclosure

I have used an LLM for writing most of the rustdocs and some tests. The shared capability-error helper API and prompt forwarding were implemented by me line by line. For provider raw-body preservation, an agent helped identify low-risk provider paths and talk through the edits, but I made changes myself.

Update: Changes integrated into this PR after review are agent-driven with direction from me.

### DIFF
--- a/crates/rig-core/src/audio_generation.rs
+++ b/crates/rig-core/src/audio_generation.rs
@@ -14,10 +14,11 @@ use thiserror::Error;
 /// [`Self::provider_response_json`], and [`Self::provider_response_status`].
 ///
 /// Note: no provider path currently constructs [`Self::ProviderResponse`] for
-/// audio generation; real audio failures surface as [`Self::HttpError`], which
-/// the helpers already read. The variant is kept for symmetry with the other
-/// capability errors and for future provider paths that preserve a 2xx error
-/// envelope.
+/// audio generation; coverage is limited today — only Azure audio failures
+/// surface as [`Self::HttpError`], which the helpers read, while other audio
+/// providers still emit [`Self::ProviderError`]. The variant is kept for
+/// symmetry with the other capability errors and for future provider paths that
+/// preserve a 2xx error envelope.
 #[derive(Debug, Error)]
 #[non_exhaustive]
 pub enum AudioGenerationError {
--- a/crates/rig-core/src/audio_generation.rs
+++ b/crates/rig-core/src/audio_generation.rs
@@ -12,7 +12,14 @@ use thiserror::Error;
 ///
 /// Inspect provider failures with [`Self::provider_response_body`],
 /// [`Self::provider_response_json`], and [`Self::provider_response_status`].
+///
+/// Note: no provider path currently constructs [`Self::ProviderResponse`] for
+/// audio generation; real audio failures surface as [`Self::HttpError`], which
+/// the helpers already read. The variant is kept for symmetry with the other
+/// capability errors and for future provider paths that preserve a 2xx error
+/// envelope.
 #[derive(Debug, Error)]
+#[non_exhaustive]
 pub enum AudioGenerationError {
     /// Http error (e.g.: connection error, timeout, etc.)
     #[error("HttpError: {0}")]
@@ -39,40 +46,7 @@ pub enum AudioGenerationError {
     ProviderResponse(provider_response::ProviderResponseError),
 }
 
-impl AudioGenerationError {
-    /// Returns the raw provider response body when available.
-    ///
-    /// This is available for:
-    /// - [`AudioGenerationError::ProviderResponse`] using its preserved body.
-    /// - [`AudioGenerationError::HttpError`] when it wraps an HTTP non-success response that carries a body.
-    pub fn provider_response_body(&self) -> Option<&str> {
-        match self {
-            Self::ProviderResponse(response) => Some(response.body.as_str()),
-            Self::HttpError(error) => error.non_success_body(),
-            _ => None,
-        }
-    }
-
-    /// Parses the provider response body as JSON.
-    ///
-    /// Returns:
-    /// - `Ok(Some(value))` when a body is present and valid JSON.
-    /// - `Ok(None)` when no provider response body is available.
-    /// - `Err(error)` when a body is present but isn't valid JSON.
-    pub fn provider_response_json(&self) -> Result<Option<serde_json::Value>, serde_json::Error> {
-        provider_response::json(self.provider_response_body())
-    }
-
-    /// Returns the HTTP status code when this error preserves one, either from a
-    /// non-success HTTP response or from a preserved provider response.
-    pub fn provider_response_status(&self) -> Option<http::StatusCode> {
-        match self {
-            Self::ProviderResponse(response) => response.status,
-            Self::HttpError(error) => error.non_success_status(),
-            _ => None,
-        }
-    }
-}
+crate::provider_response::impl_provider_response_helpers!(AudioGenerationError);
 
 pub trait AudioGeneration<M>
 where
--- a/crates/rig-core/src/audio_generation.rs
+++ b/crates/rig-core/src/audio_generation.rs
@@ -2,7 +2,7 @@
 //! Rig abstracts over a number of different providers using the [AudioGenerationModel] trait.
 use crate::markers::{Missing, Provided};
 use crate::{
-    http_client,
+    http_client, provider_response,
     wasm_compat::{WasmCompatSend, WasmCompatSync},
 };
 use serde_json::Value;
@@ -18,18 +18,52 @@ pub enum AudioGenerationError {
     #[error("JsonError: {0}")]
     JsonError(#[from] serde_json::Error),
 
-    /// Error building the transcription request
+    /// Error building the audio generation request
     #[error("RequestError: {0}")]
     RequestError(#[from] Box<dyn std::error::Error + Send + Sync + 'static>),
 
-    /// Error parsing the transcription response
+    /// Error parsing the audio generation response
     #[error("ResponseError: {0}")]
     ResponseError(String),
 
-    /// Error returned by the transcription model provider
+    /// Error returned by the audio generation model provider
     #[error("ProviderError: {0}")]
     ProviderError(String),
 }
+
+impl AudioGenerationError {
+    /// Returns the raw provider response body when available.
+    ///
+    /// This is available for:
+    /// - [`AudioGenerationError::ProviderError`] using its stored string.
+    /// - [`AudioGenerationError::HttpError`] when it wraps an HTTP non-success response that carries a body.
+    pub fn provider_response_body(&self) -> Option<&str> {
+        match self {
+            Self::ProviderError(body) => provider_response::body(Some(body.as_str()), None),
+            Self::HttpError(error) => provider_response::body(None, Some(error)),
+            _ => None,
+        }
+    }
+
+    /// Parses the provider response body as JSON.
+    ///
+    /// Returns:
+    /// - `Ok(Some(value))` when a body is present and valid JSON.
+    /// - `Ok(None)` when no provider response body is available.
+    /// - `Err(error)` when a body is present but isn't valid JSON.
+    pub fn provider_response_json(&self) -> Result<Option<serde_json::Value>, serde_json::Error> {
+        provider_response::json(self.provider_response_body())
+    }
+
+    /// Returns the HTTP status code when this error comes from a non-success HTTP response.
+    pub fn provider_response_status(&self) -> Option<http::StatusCode> {
+        match self {
+            Self::HttpError(error) => provider_response::status(Some(error)),
+            _ => None,
+        }
+    }
+}
+
 pub trait AudioGeneration<M>
 where
     M: AudioGenerationModel,
@@ -168,5 +202,53 @@ where
         let model = self.model.clone();
 
         model.audio_generation(self.build()).await
+    }
+}
+
+#[cfg(test)]
+mod provider_response_tests {
+    use super::*;
+    use http::StatusCode;
+
+    #[test]
+    fn audio_generation_error_provider_response_helpers_with_provider_json_body() {
+        let body = r#"{"error":{"message":"invalid voice"}}"#;
+        let error = AudioGenerationError::ProviderError(body.to_string());
+
+        assert_eq!(error.provider_response_body(), Some(body));
+        assert_eq!(error.provider_response_status(), None);
+        assert_eq!(
+            error.provider_response_json().expect("valid JSON"),
+            Some(serde_json::json!({ "error": { "message": "invalid voice" } }))
+        );
+    }
+
+    #[test]
+    fn audio_generation_error_provider_response_helpers_with_http_non_success() {
+        let body = r#"{"error":{"message":"bad request"}}"#;
+        let error =
+            AudioGenerationError::HttpError(http_client::Error::InvalidStatusCodeWithMessage(
+                StatusCode::BAD_REQUEST,
+                body.to_string(),
+            ));
+
+        assert_eq!(error.provider_response_body(), Some(body));
+        assert_eq!(
+            error.provider_response_status(),
+            Some(StatusCode::BAD_REQUEST)
+        );
+        assert_eq!(
+            error.provider_response_json().expect("valid JSON"),
+            Some(serde_json::json!({ "error": { "message": "bad request" } }))
+        );
+    }
+
+    #[test]
+    fn audio_generation_error_provider_response_helpers_with_unrelated_variant() {
+        let error = AudioGenerationError::ResponseError("parse failed".to_string());
+
+        assert_eq!(error.provider_response_body(), None);
+        assert_eq!(error.provider_response_status(), None);
+        assert_eq!(error.provider_response_json().expect("no body"), None);
     }
 }
--- a/crates/rig-core/src/audio_generation.rs
+++ b/crates/rig-core/src/audio_generation.rs
@@ -33,18 +33,22 @@ pub enum AudioGenerationError {
     /// Error returned by the audio generation model provider
     #[error("ProviderError: {0}")]
     ProviderError(String),
+
+    /// Raw error response preserved from the audio generation model provider
+    #[error("ProviderResponseError: {0}")]
+    ProviderResponse(provider_response::ProviderResponseError),
 }
 
 impl AudioGenerationError {
     /// Returns the raw provider response body when available.
     ///
     /// This is available for:
-    /// - [`AudioGenerationError::ProviderError`] using its stored string.
+    /// - [`AudioGenerationError::ProviderResponse`] using its preserved body.
     /// - [`AudioGenerationError::HttpError`] when it wraps an HTTP non-success response that carries a body.
     pub fn provider_response_body(&self) -> Option<&str> {
         match self {
-            Self::ProviderError(body) => provider_response::body(Some(body.as_str()), None),
-            Self::HttpError(error) => provider_response::body(None, Some(error)),
+            Self::ProviderResponse(response) => Some(response.body.as_str()),
+            Self::HttpError(error) => error.non_success_body(),
             _ => None,
         }
     }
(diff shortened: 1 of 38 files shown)


Shaurya-Sethi:
@gold-silver-copper please review when you find time - I have some free time this week in case you need changes or follow-up PRs for other providers

gold-silver-copper:
Thanks for putting this together — having a way to inspect provider error responses is definitely the right direction.

I think the main thing we need to tighten up before merging is the definition of what actually counts as a "provider response body".

Right now, `provider_response_body()` returns `Some(...)` for every `ProviderError(String)`:

https://github.com/0xPlaygrounds/rig/pull/1859/files#diff-00968c2f78c1de925876c5ea9778904d4f21816c307a82447587d8b0b178f7e3R100-R104

The problem is that `ProviderError` is not currently reserved for raw provider responses. It is also used for Rig-generated and internal diagnostics. For example, `from_stream_transport` converts non-HTTP stream transport errors into `ProviderError(error.to_string())`:

https://github.com/0xPlaygrounds/rig/pull/1859/files#diff-00968c2f78c1de925876c5ea9778904d4f21816c307a82447587d8b0b178f7e3R131-R136

The new tests then explicitly assert that this kind of internal transport message is returned by `provider_response_body()`:

https://github.com/0xPlaygrounds/rig/pull/1859/files#diff-46682d7c6a6129879cc3dd6ed8dca4cc31cdffef7b01124df94f0ec6de3c7901R917-R929

That means callers of `provider_response_body()` cannot reliably distinguish between:

* the actual response body returned by the provider, and
* a synthetic string generated by Rig while handling an error.
I think this needs a clearer boundary before it becomes public API.

My preference would be to store preserved provider responses separately from `ProviderError(String)`, for example:

```rust
ProviderResponseError {
    status: Option<StatusCode>,
    body: String,
}
```

`ProviderError` is already used quite broadly for provider-related failures, transport issues, and internal diagnostics. Reinterpreting it to always mean "raw provider payload" would require a fairly extensive audit across the codebase, and it would be easy for future changes to accidentally violate that invariant again.

A dedicated `ProviderResponseError`-style representation makes the distinction explicit in the type system and avoids relying on conventions that can drift over time.


There are also still several paths that flatten status and body together into a single `ProviderError`. For example, OpenAI image generation:

https://github.com/0xPlaygrounds/rig/pull/1859/files#diff-39f0be0b2de322c46f39dcbe067528439f70aaf872db68f9f1413d924d803d43R116-R122

Azure image generation:

https://github.com/0xPlaygrounds/rig/pull/1859/files#diff-565525947ff8a1a1f1328b12d5c2d28e3732b8e24f22fd1ea7f0640868d2371cbR987-R993

Azure audio generation:

https://github.com/0xPlaygrounds/rig/pull/1859/files#diff-565525947ff8a1a1f1328b12d5c2d28e3732b8e24f22fd1ea7f0640868d2371cbR1074-R1080

This preserves the body in a human-readable form, but it loses `provider_response_status()`, since status is currently only recoverable from `HttpError`:

https://github.com/0xPlaygrounds/rig/pull/1859/files#diff-00968c2f78c1de925876c5ea9778904d4f21816c307a82447587d8b0b178f7e3R118-R123

If we want this API to be reliable, those paths should preserve the status separately as structured data as well, probably via `HttpError::InvalidStatusCodeWithMessage` or an equivalent error variant.

I'd also prefer the tests to use more realistic non-2xx provider responses, e.g. `RecordingHttpClient::with_error(status, raw_json_body)`, since that's the primary scenario users are trying to debug.

Some of the new provider tests exercise success-status error envelopes instead:

https://github.com/0xPlaygrounds/rig/pull/1859/files#diff-82ff586f7cdb09b54c28b0b592d223261228f24fb9160bc57b244f49269bbe78R2254-R2286

That path is still worth covering, but it isn't the main real-world failure mode described in the issue.

Overall, I like the direction, but I think the contract should be made stricter before merge so users can trust that `provider_response_body()` actually represents the provider's response body.

Shaurya-Sethi:
Hi and thanks for the feedback - I've changed this to a draft PR. I'll work towards making the necessary changes and let you know when done :)

Shaurya-Sethi:
Thanks (again) for the detailed review. I'm reworking the PR as follows:

1. A dedicated `ProviderResponseError { status: Option<StatusCode>, body: String }` representation, stored in a new `ProviderResponse` variant on the capability error enums. `ProviderError(String)` will no longer be treated as a provider response body, so the helpers return `None` for it and Rig-generated diagnostics (like the stream transport strings) stop leaking through `provider_response_body()`.

2. The paths that flatten status and body into a single string (OpenAI image generation, Azure image and audio generation) will preserve them structurally via `HttpError(InvalidStatusCodeWithMessage(status, body))` instead, so `provider_response_status()` works there too.

3. Tests will primarily exercise realistic non-2xx responses through `RecordingHttpClient::with_error(status, raw_json_body)`, keeping the success-status error envelope coverage as a secondary case.

One note on the variant: `CompletionError`, `EmbeddingError` and `VerifyError` are not `#[non_exhaustive]`, so adding `ProviderResponse` is a breaking change for downstream exhaustive matches. I think that trade-off should be okay here?, since the stricter contract needs to live in the type system rather than in conventions, and I'll state the break explicitly in the PR description once it's ready. Flagging it now in case you would rather handle it differently. Let me know

Shaurya-Sethi:
Reopened after changes, and updated PR description too
Ready for your review @gold-silver-copper

gold-silver-copper:
A few more review points:


## 1. OpenAI Responses non-streaming errors still aren't inspectable (`providers/openai/responses_api/mod.rs`, ~L1531)

The non-success branch still does:

```rust
let text = http_client::text(response).await?;
Err(CompletionError::ProviderError(text))
```

So `provider_response_status()` / `provider_response_body()` / `provider_response_json()` all return nothing for failed `/responses` calls.

Since the default `openai::Client` completion model goes through the Responses API (the new tests use `CompletionsClient`, which is the chat-completions path), this means the most common OpenAI completion path doesn't benefit from the feature.

The fix should follow the convention used everywhere else in this PR: capture status before reading the body and return `HttpError(InvalidStatusCodeWithMessage(status, text))`.

Notably, the sibling Copilot Responses path was fixed, so I assume this one was just missed.

## 2. VerifyError helpers can't fire on real verify failures (`client/mod.rs`, ~L538–555)

`verify()` maps 500/529 to `VerifyError::ProviderError(text)` and other non-success statuses to `HttpError(Error::Instance(format!(...)))` — neither preserves an inspectable status/body (`non_success_status()` / `non_success_body()` return `None` for `Error::Instance`).

So the new `VerifyError::ProviderResponse` variant is never constructed by library code, and the helpers advertised in the `VerifyError` rustdoc return `None` for every actual verification failure.

The existing tests don't catch this because they hand-construct the error variants rather than driving `verify()` through a mock client.

Suggest returning `InvalidStatusCodeWithMessage(status, body)` from these arms and adding a test that exercises `verify()` end-to-end with `RecordingHttpClient`.

Related: `AudioGenerationError::ProviderResponse` is in the same situation — no provider path constructs it (Azure audio only emits `HttpError`). Fine if it's intentional forward-looking surface, but worth a note.

## Smaller points 

* **Responses streaming type: "error" events** (`responses_api/streaming.rs`, ~L534) extract only `error.message` and drop the rest of the payload (`code`, `type`, `request id`). Since the raw `data` JSON is in hand and this is a provider-authored error over a 2xx stream, `ProviderResponse(ProviderResponseError { status: None, body: data.to_owned() })` fits the taxonomy exactly — and would be the first real producer of `status: None`, which currently only appears in tests.
* The `let _ = err.message;` idiom (~15 sites) is confusing at every call site. `tracing::warn!` the parsed message before returning the preserved body so it stays useful.
* The three helper methods are copy-pasted verbatim across six error enums (~200 lines). A small `macro_rules!` in `provider_response.rs` would collapse this and prevent drift.
* Since this PR already breaks exhaustive matches on `CompletionError` / `EmbeddingError` / `VerifyError`, it's a natural moment to add `#[non_exhaustive]` (as `TranscriptionError` already has) so the next variant addition isn't breaking too. Also consider the `!` in the commit title given the documented breaking changes.
* If the Responses paths end up deferred to the follow-up list instead, the PR description and helper rustdocs should say so explicitly — right now the helpers read as if availability depends only on the error variant, which will generate "why is this always None" reports for untouched providers.

Shaurya-Sethi:
@gold-silver-copper Hey, I’ve addressed your review-points in this updated PR. Here's a quick summary (redundant with PR):

1. For **OpenAI Responses**, the non-success branch now captures status and returns `HttpError(InvalidStatusCodeWithMessage(status, text))`, matching the convention used everywhere else, so the default completion path benefits from the feature. I also added a test driving it via `RecordingHttpClient::with_error_response`.

2. For **`verify()`**, the 500/529 and generic non-success arms now return `InvalidStatusCodeWithMessage` so status and body stay inspectable; 401/403 still map to `InvalidAuthentication`. An end-to-end `verify()` test through a mock client was also added.

Worth flagging: genuine non-2xx already short-circuits to `HttpError(InvalidStatusCodeWithMessage)` inside `HttpClientExt::send` before these arms (and the Responses `else` branch) run, so for the default reqwest client the helpers were already functional on real 4xx/5xx. The arms only execute for clients that hand back the response instead of erroring. So these changes mostly harden that non-short-circuiting path and remove the misleading `ProviderError`.

I can adjust this if you'd rather handle the short-circuit case differently.

3. **Streaming error events** are now preserved as `ProviderResponse { status: None, body: <raw data> }`, keeping `code`/`type`/etc. This is the first real producer of `status: None`. The parsed `message` is logged before returning.

4. **`let _ = err.message`** is replaced across all sites with `tracing::warn!` of the parsed message. I went with `warn!` per your suggestion, but easy to drop to `debug!` if you'd prefer.

5. **Helper duplication** is collapsed into a single `macro_rules! impl_provider_response_helpers!` in `provider_response.rs` and applied across all six error enums.

6. I added `#[non_exhaustive]` to `CompletionError`, `EmbeddingError`, `VerifyError`, `ImageGenerationError`, and `AudioGenerationError` (Transcription already had it), since the PR was already breaking exhaustive matches.

7. I kept **AudioGenerationError::ProviderResponse** as forward-looking surface for symmetry and documented on the enum that no path constructs it yet (audio failures surface as `HttpError`, which the helpers read).

Small note on the `!` suffix in commits - I added this to `290a4da9` but there are more commits that include breaking changes - I didn't want to rewrite history - let me know if you'd like that done and I'll do it.

Ready for another look when you are. Thanks for your time btw.

gold-silver-copper:
This is much more solid, I'd normally merge as is but I am getting a bit stricter nowadays >:)

Hopefully final review:

1. **The "Responses streaming error events" claim doesn't hold for the live stream.** The description says streaming error events are preserved as `ProviderResponse { status: None, body }`, but the live loop (`stream_from_event_source_with_options`, what `chatgpt::stream()` actually uses) deserializes each SSE line into `StreamingCompletionChunk` — `#[serde(untagged)]` with only `Response`/`Delta`. A `{"type":"error",…}` event matches neither, so it falls into:

   ```rust
   let Ok(data) = data else { debug!(...); continue; };  // error event silently dropped
   ```

   and the stream ends as a partial success. The upgraded `ProviderResponse` handling lives in the buffered `raw_choices_from_sse_body`, which the live loop never calls — and `streaming_error_event_preserves_full_payload` tests that buffered parser directly, so it gives false confidence. The same pattern is systemic in the chat-completions streaming path: `normalize_chunk` maps an in-band `data:{"error":…}` envelope to `Ok(None) => continue` for every shared profile (groq/openai/deepseek/llamafile/xai/openrouter/copilot), and the `streaming_mid_stream_*` test injects a transport `Err`, not an in-band envelope. Net: a user inspecting a streamed provider error still gets `None`. This isn't a regression (the loop always dropped these), but please either wire the live loop to surface error events as `ProviderResponse` (and add a test that drives the live loop) or scope the description to the buffered/non-streaming path.

2. **Tests don't exercise the rewritten non-2xx branches.** The `*_http_non_success_*` tests for openai-completion, openai-embedding, and groq use `RecordingHttpClient::with_error`, which makes `send()` itself return `Err(...)`, short-circuiting at `self.client.send(req).await?` before the rewritten `else { … HttpError(InvalidStatusCodeWithMessage) }` arm runs — the assertions pass via the `#[from]` conversion, not the new code. (That arm is also only reachable through a custom `HttpClientExt`; under the bundled reqwest client `send()` already short-circuits.) `with_error_response` is the double that actually reaches it. Separately, ~13 providers got error-path rewrites with no direct test (deepseek, galadriel, llamafile, moonshot, perplexity, voyageai, mistral ×2, openrouter ×2, together ×2, huggingface, copilot embedding). Together is the one I'd most want pinned — it uses a distinct typed `ApiResponse::Error` storing `status: Some(status)` (`together/completion.rs:303-311`), structurally unlike the tested OpenAI/groq paths, so a broken mapping there ships undetected. A parametrized test per deserialization shape (or at least a Together case + `with_error_response` for one of the inline `else` branches) would close this.

---

- **`AudioGenerationError` doc overstates availability.** The note says real audio failures "surface as `Self::HttpError`, which the helpers already read," but only azure audio uses `HttpError`; openai/xai/openrouter/hyperbolic audio still emit `ProviderError(format!("{status}: {text}"))`, which the helpers map to `None` — so the helpers return `None` for 4 of 5 audio providers. (`AudioGenerationError::ProviderResponse` is also never constructed — fine as disclosed, but the surrounding claim is wrong.)
- **Transcription paths are untestable.** All `RecordingHttpClient` doubles return `NOT_IMPLEMENTED` for `send_multipart` (`test_utils/http.rs:161/209/275/334`), so the new transcription `ProviderResponse`/`HttpError` branches (OpenAI, Groq, Azure) can't be covered with the current doubles. Worth a multipart-capable double, even minimal.
- **Non-streaming Responses parser is inconsistent with its sibling.** `parse_sse_completion_body` (ChatGPT non-streaming completion) still flattens to `ProviderError(message)` while `raw_choices_from_sse_body` was upgraded — so even non-streaming Responses inspection is uneven.
- **`StructuredOutputError` wraps `PromptError`** but doesn't forward `provider_response_*`, so the forwarding chain stops one level short of `prompt_typed` callers.

---

- `image_generation.rs:46,48` trait-method docs still say "transcription request builder" / "at transcription time" (the PR fixed only the error-variant docs).
- **Weak assertion:** the 2xx-envelope tests assert `status == Some(StatusCode::OK)`, but the mock always returns 200 — they can't distinguish "preserved the real status" from "hardcoded OK."
- `provider_response_status()` returns `Some(200)` for a 2xx error envelope — truthful, but a consumer calling `is_client_error()` could be surprised; one-line doc callout would help. `PromptError::provider_response_status`'s doc also says "non-success HTTP response" when it can return `Some(2xx)`.
- `provider_response_json()` returns `Err` (not `Ok(None)`) on an empty preserved body, since several providers store `from_utf8_lossy(...)` unconditionally — an `if let Ok(Some(v))` caller silently misses empty bodies.
- **`warn!` field shape diverges:** most providers log `message = %err.message`; huggingface logs `error = %err`; together adds `code`. (Pre-existing: huggingface's `debug!("Huggingface completion error: …")` at `completion.rs:782` fires in the success branch.)
- **Dead `From<ApiErrorResponse>` impls** (e.g. `openai/embedding.rs:29-42`, `galadriel.rs:234-237`) still reduce to `err.message`, dropping status+body — they're bypassed by the new inline paths, but contradict the new contract; worth deleting or converting.
- `VerifyError::ProviderResponse` is constructed only in tests (dead in production) and, unlike `AudioGenerationError`, isn't documented as forward-looking.

Shaurya-Sethi:
@gold-silver-copper Hi. I've addressed your latest review on #1859 and pushed `d7f531e`.

Summary:
- Live Responses + chat-completions streaming now surface in-band provider errors as `ProviderResponse`
- `parse_sse_completion_body` preserves raw error payloads, with direct test coverage
- `StructuredOutputError` forwards `provider_response_*`
- Multipart-capable `RecordingHttpClient` + transcription non-success tests
- Audio/verify docs corrected; dead `From<ApiErrorResponse>` impls removed
- PR description updated to match actual scope/limits

Thanks!

gold-silver-copper:
LGTM! Thanks for working through this PR!

PR #1880, opened 2026-06-08
fix(streaming): record per-call token usage on chat generation spans

mateobelanger:
Streams driven inside a caller-owned tracing span were recording aggregate token usage onto that ambient span, which can drop undeclared gen_ai.usage fields. This records each completion call usage on the per-call chat_streaming span, while only writing aggregate usage to invoke_agent when Rig created that span. 

The change is provider-agnostic and adds MockCompletionModel tests with a span-capturing tracing layer for single-turn and multi-turn tool-use streams. Validated with `cargo fmt --check` and `cargo test -p rig-core stream_prompt_records --lib`.

## Justification

Cortex consumes rig's stream with its own `try_next()` loop inside an active span so LLM work nests under its request trace. That exact shape was the broken path: rig recorded only aggregate usage onto cortex's ambient span, which never declared the `gen_ai.usage.*` fields, so `tracing` dropped them. This means zero token costs for the run.

This fix emits per-call generation spans (parented under cortex's span) each carrying their own usage, so cortex gets native per-call cost/nesting and can delete its hand-rolled span workaround.

### DIFF
--- a/crates/rig-core/src/agent/prompt_request/streaming.rs
+++ b/crates/rig-core/src/agent/prompt_request/streaming.rs
@@ -320,15 +320,38 @@ fn record_completion_call_if_needed(
     completion_call_emitted: &mut bool,
     call_index: usize,
     current_call_usage: Option<crate::completion::Usage>,
+    chat_stream_span: &tracing::Span,
 ) -> Option<CompletionCall> {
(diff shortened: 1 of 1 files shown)


gold-silver-copper:
fantastic!

PR #1509, opened 2026-03-13
Redis vector store integration

daric93:
### Summary
Adds Redis vector store implementation to Rig framework using RediSearch for efficient KNN vector similarity search.

Fixes #1239 - Originally requested Valkey support, implemented as Redis for broader compatibility and ecosystem maturity. Redis provides the same vector capabilities through RediSearch while offering more established tooling, documentation, and cloud provider support.
### Vector Store Integration
Redis is an in-memory data structure store that can be used as a vector database through its RediSearch module. This integration leverages Redis's FT.SEARCH command with KNN queries for high-performance vector similarity search, making it suitable for applications requiring fast, scalable vector operations with the reliability and features of Redis.
### Key Capabilities

- High Performance: In-memory storage with sub-millisecond query times
- Scalability: Horizontal scaling with Redis Cluster support
- Persistence: Optional disk persistence for data durability
- Rich Filtering: Advanced metadata filtering alongside vector search
- Production Ready: Battle-tested in production environments worldwide

### Resources
Source Code: https://github.com/redis/redis
Rust SDK: https://github.com/redis-rs/redis-rs
RediSearch Module: https://github.com/RediSearch/RediSearch
Official Documentation: https://redis.io/docs/stack/search/
Redis Stack: https://redis.io/docs/stack/

Issue #1239, opened 2026-01-10
feat: Add support for Valkey vector store


daric93:
## Valkey Vector Store Integration Request
Adding Valkey as a possible data store would be quite beneficial for the library, as it’s a community-driven, Redis-protocol-compatible project that’s seeing growing adoption as a vendor-neutral alternative. Supporting Valkey would expand Rig’s compatibility with existing Redis-based infrastructure and managed services, increasing the number of potential use cases for caching, metadata storage, and lightweight retrieval workflows.

I’m happy to implement this if it makes sense for the project — the scope looks fairly clear (adding a new crate, implementing the relevant datastore traits, and wiring it into the release list). I wanted to reach out first to check whether Valkey support would be welcome before starting work.

### Resources
[Valkey](https://valkey.io/)
[Valkey-Glide](https://github.com/valkey-io/valkey-glide)

joshua-mo-143:
Hi @daric93, thanks for opening an issue!

I think it would be very beneficial to see examples of companies that are actually using Valkey in production for this specific use case. I'm happy to accept this but it's difficult to justify adding if nobody's actually going to use it

daric93:
That makes sense. Since Valkey adoption is still early, I’m happy to switch the proposal to Redis instead. Redis is widely used in production for embeddings, metadata, and lightweight vector similarity, which should make the justification much clearer. Given Valkey’s Redis protocol compatibility, this would also keep the door open for Valkey support later with minimal additional work or full compatibility.

daric93:
Hi @joshua-mo-143 
Went ahead and implemented this as a Redis vector store in PR #1509 . Used Redis instead of Valkey given the more established production adoption. Happy to adjust based on your feedback!

PR #1885, opened 2026-06-09
fix(rig-redis): address review comments from #1509 and harden edge cases

daric93:
## Summary

Follow-up to the merged [#1509](https://github.com/0xPlaygrounds/rig/pull/1509) (Redis vector store integration) addressing the [review comments](https://github.com/0xPlaygrounds/rig/pull/1509#issuecomment-4654557758) left by the maintainer. Also relates to the original request in [#1239](https://github.com/0xPlaygrounds/rig/issues/1239).

### P1 — Metadata filtering not supported for inserted documents

**Comment:** `InsertDocuments` only writes `document`, `embedded_text`, and the vector field. Filter queries reference fields that never exist in the hash.

**Fix:** Implemented `with_metadata_fields()` builder method. When configured, specified top-level scalar fields (string/number/bool) are extracted from the serialized document JSON during insertion and written as separate hash fields. This enables RediSearch filter queries on those fields. Reserved field names (`document`, `embedded_text`, vector field) are guarded against to prevent data corruption. Null/array/object values are silently skipped with a `tracing::warn`.

---

### P1 — Key prefix mismatch in docs/examples

**Comment:** README creates index with `PREFIX 1 doc:` but the example doesn't use `.with_key_prefix("doc:")`, so inserted documents are invisible to FT.SEARCH.

**Fix:** Added `.with_key_prefix("doc:".to_string())` to the example and README. Added doc comments emphasizing the prefix must match the index configuration.

---

### P1 — No escaping in filter values

**Comment:** Values with special characters (spaces, `|`, braces, `@`) are interpolated directly into RediSearch syntax, risking injection/parse errors.

**Fix:** Added `escape_tag_value()` and `escape_text_value()` functions that escape RediSearch special characters per [official Redis documentation](https://redis.io/docs/latest/develop/ai/search-and-query/advanced-concepts/tags/). Applied escaping in all filter construction methods. Added `Filter::raw()` for pre-escaped queries with security warning. Added regression tests for special characters.

---

### P2 — Range operators on non-numeric values silently degrade

**Comment:** `gt("status", "active")` becomes `@status:{active}` (tag filter) instead of erroring.

**Fix:** The `SearchFilter` trait returns `Self` (not `Result`), so errors cannot be returned. Added `tracing::warn!` when non-numeric values are used with `gt`/`lt`/`gte`/`lte`, and documented the limitation clearly.

---

### P2 — LIMIT not always applied

**Comment:** Without a threshold, RediSearch defaults to `LIMIT 0 10`, so `samples > 10` is silently capped at 10 results.

**Fix:** Now always appends `LIMIT 0 req.samples()` to override RediSearch's default.

---

### P2 — Integration tests panic without Docker

**Fix:** Marked all integration tests with `#[ignore]`. They only run with `cargo test -- --ignored`.

---

### P2 — Tests mutate local Redis with FT.DROPINDEX

**Fix:** Removed auto-detection of local Redis. Tests always use testcontainers unless `REDIS_URL` is explicitly set. Added `verify_redisearch()` capability probe.

---

### P3 — README shows `new()` as sync and links to wrong path

**Fix:** Added `.await?` to README example, fixed link to `crates/rig-redis/examples`.

---

## Additional Hardening

Based on additional review:

- **Empty document JSON handling:** `parse_search_response` now skips documents with empty/unparseable JSON (with `tracing::warn`) instead of aborting the entire result set
- **Empty `tag_in` guard:** `tag_in(field, vec![])` returns match-all `*` instead of invalid `@field:{}`
- **Missing `__vector_score` warning:** emits `tracing::warn` when score field is absent
- **`text_phrase()` method:** added for exact phrase matching (`@field:"phrase"`)
- **`text_contains()` documented:** clarified it performs token-AND, not phrase search
- **`Filter::raw()` security warning:** explicitly warns against unsanitized user input
- **Dynamic API subset documented:** notes that `VectorStoreIndexDyn` only supports Eq/Gt/Lt/And/Or
- **`From<u64>` precision loss:** documented that values above 2⁵³ lose precision

## Example Rewrite

Rewrote `examples/vector_search_redis.rs` from a minimal word-definition demo to a product catalog demonstrating:
- Basic vector similarity search (no filter)
- TAG field filtering (`category = Electronics`)
- NUMERIC range filtering (`price < 100`)
- Combined AND filters (`Electronics AND price > 100`)
- Proper document round-tripping (all fields preserved)
- `with_metadata_fields` and `with_key_prefix` usage

## Test Results

- 34 filter unit tests ✓
- 15 integration tests ✓ (TAG, NUMERIC, boolean, AND/OR, no-match, missing fields, backward compat, top_n_ids)

## Changes

- `crates/rig-redis/src/lib.rs` — metadata fields, LIMIT fix, graceful error handling, docs
- `crates/rig-redis/src/filter.rs` — escaping, `text_phrase`, empty `tag_in` guard, docs
- `crates/rig-redis/tests/integration_tests.rs` — 10 new metadata tests, #[ignore], testcontainers
- `crates/rig-redis/tests/filter_tests.rs` — escaping + edge case tests
- `crates/rig-redis/examples/vector_search_redis.rs` — full rewrite with filtering demos
- `crates/rig-redis/README.md` — metadata filtering docs, fixed examples

### DIFF
--- a/crates/rig-redis/README.md
+++ b/crates/rig-redis/README.md
@@ -6,9 +6,16 @@ Vector store index integration for [Redis](https://redis.io/) using RediSearch v
 
 - Vector similarity search using Redis's RediSearch module
 - Support for KNN (k-nearest neighbors) queries
-- Metadata filtering with Redis query syntax
 - Document insertion with automatic embedding storage
 - Compatible with Redis 7.2+ or Redis Stack
+
+## Filtering
+
+Filters apply to fields that exist in your RediSearch index schema **and** are
+present in the stored hash keys. The `InsertDocuments` implementation only writes
+`document`, `embedded_text`, and the vector field. If you need filterable metadata
+fields, write them to the hash separately or use a custom insertion approach that
+includes the filterable fields in the hash.
 
 ## Prerequisites
 
@@ -38,25 +45,28 @@ Replace `1536` with your embedding model's dimensionality.
 
 ## Usage Example
 
-```rust
+```rust,no_run
 use rig::providers::openai;
 use rig::vector_store::{InsertDocuments, VectorStoreIndex};
 use rig_redis::RedisVectorStore;
 
+# async fn run() -> Result<(), Box<dyn std::error::Error>> {
 // Create embedding model
-let openai_client = openai::Client::from_env();
+let openai_client = openai::Client::from_env()?;
 let model = openai_client.embedding_model(openai::TEXT_EMBEDDING_3_SMALL);
 
 // Create Redis client
 let redis_client = redis::Client::open("redis://127.0.0.1:6379")?;
 
-// Create vector store
+// Create vector store — key prefix must match the index PREFIX configuration
 let vector_store = RedisVectorStore::new(
     model,
     redis_client,
     "word_idx".to_string(),      // index name
     "embedding".to_string(),      // vector field name
-);
+)
+.await?
+.with_key_prefix("doc:".to_string());
 
 // Insert documents
 vector_store.insert_documents(documents).await?;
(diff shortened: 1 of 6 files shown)


daric93:
Closing this PR — I see that the maintainer already pushed #1884 which rewrites the filter API (adding `Result` returns, `RedisNumber` with NaN/Inf validation, field name validation, etc.). My changes conflict significantly with that rewrite.

The metadata filtering feature (`with_metadata_fields`) from this PR is still not in upstream, so I may open a smaller follow-up PR rebased on the new filter API if there's interest. Happy to coordinate.

PR #1894, opened 2026-06-10
fix(openai-responses): reject malformed output tags

gold-silver-copper:
This is stacked on top of #1855 and preserves the original commit attribution from that PR.

#1855 preserves raw payloads for unknown OpenAI Responses output items. This follow-up keeps that behavior, but restores strict deserialization for malformed output tags:

- unknown string `type` values are still preserved as `Output::Unknown(Value)`
- missing `type` now errors instead of becoming `Unknown`
- non-string `type` now errors instead of becoming `Unknown`
- known malformed variants still error

Validation:

```sh
cargo fmt
cargo test -p rig-core providers::openai::responses_api::tests --lib
```

Both passed locally.


ForeverAngry:
LGTM!

PR #1902, opened 2026-06-12
fix(tool)!: lock tool-pipeline behavior with cassettes and fix five issues surfaced while recording

gold-silver-copper:
## Summary

This PR locks in the externally observable behavior of rig's tool pipeline with cassette tests, then fixes five issues those tests surfaced. The motivation is the planned migration of the handrolled tool plumbing onto rmcp: with the wire traffic frozen, that migration can be verified against recorded reality instead of re-derived expectations.

Read the commits in order: (1) tests pinning current behavior, (2) the fixes — each re-recording exactly the cassettes that pinned the old behavior, so the YAML diffs show each fix's precise wire-level consequence, (3–5) hardening from further review: definition-assembly dedup, advertised-tools accuracy carried in serialized machine state, single MCP schema conversion path, ordered multimodal parts, multi-turn completability for forced tool choice, and two test-reliability fixes.

> **Note on tool choice:** in multi-turn agent runs, `ToolChoice::Required`/`Specific` relax to `Auto` on the wire after the first turn that produces tool calls, so a forced run can finish with a text answer instead of exhausting `max_turns`. The configured choice is never mutated — it is preserved in snapshots/diagnostics and still bounds rig-side validation every turn, so a `Specific` allowlist keeps rejecting out-of-list calls run-long. The relaxation is tracked by a flag in serialized state, so a resumed run does not re-force. This is a behavior change from main (see **Breaking changes**); a more explicit, opt-in alternative — a tool-shaped exit — is noted under **Deliberately not fixed here** and **Follow-ups**.

## The test suite (26 new Gemini cassette tests)

- **agent_tools_e2e** — real `ToolSet` execution through both built-in drivers: chained multi-turn runs with usage aggregation, parallel calls landing in one tool-result message (also under `with_tool_concurrency(1)`), null-args normalization, string-verbatim vs JSON-serialized outputs.
- **tool_errors** — tool `Err` values, argument deserialization failures, and execution-time missing tools round-trip to the model as tool results; recovery is recorded end to end.
- **tool_hooks** — `on_tool_call` Skip/Terminate semantics; call/result hooks observing exact names, args, and raw outputs.
- **tool_server** — `ToolServerHandle` add/remove between turns changes the next request's advertised definitions; one handle backing multiple agents (the surface `McpClientHandler` drives on tool-list-changed).
- **tool_definitions** — rich JSON-schema conversion to Gemini's schema type, `#[rig_tool]` macro output on the wire, duplicate-name registration semantics.
- **dynamic_tools** — `ToolEmbedding` toolsets retrieved per prompt from an in-memory vector store over recorded Gemini embeddings, merged with static tools, sample cap enforced.
- **rmcp_parity** — an in-process MCP server over a duplex transport behind `McpTool`. The keystone pair replays one shared cassette with the same tool registered natively and via rmcp, proving the two registration paths produce identical wire requests. Also covers `is_error` round-trips, image content, live tool-list swaps, and audio content.
- **tool_choice** — `Specific` allowlist enforcement (raw streaming and non-streaming: only the named tool is callable, others are rejected) and `None` (no tool calls emitted) for both drivers.

Re-record any of them with `RIG_PROVIDER_TEST_MODE=record GEMINI_API_KEY=… cargo test --all-features --test gemini <filter> -- --test-threads=1`, same as the `agent_run` suites from #1901.

## The fixes (pydantic-ai cited as prior art throughout)

1. **Duplicate tool names sent duplicate function declarations and the provider 400'd the request** (`"Duplicate function declaration found: …"`). Registration is now last-wins with a warning across every path, and definition assembly in `get_tool_defs` also dedupes a dynamically retrieved name overlapping a static one (lossless: one shared toolset backs both lists, so repeats are the same tool). pydantic-ai raises an eager `UserError` at registration instead — happy to follow up with that stricter, breaking API if preferred.
2. **Tool ordering was nondeterministic across processes** (`ToolSet` HashMap iteration), which broke replay and silently defeats provider prompt caching whenever the `tools` array reorders. `ToolSet` now tracks registration order and iterates it everywhere, matching pydantic-ai's insertion-ordered dicts. Replaced names keep their position.
3. **The rmcp adapter flattened MCP image content into a base64 data-URL string**, so models received images as text. `McpTool::call` now emits the JSON shapes `ToolResultContent::from_tool_output` converts into real image parts, and multi-part results preserve their original text/image interleaving via an ordered `parts` array (the parser learned text entries; the grammar is documented on `from_tool_output`). Text-only results stay verbatim. Mirrors pydantic-ai's `ImageContent` → `BinaryImage` mapping. `McpTool::definition` also now delegates to the single `schema_as_json_value` conversion path.
4. **Tool failures leaked internal error-type names into model-visible text** (`ToolCallError: …`, `ToolNotFoundError: …`, `MCP tool error:`). Failures now read `Tool 'x' failed: <root message>` / `Invalid arguments for tool 'x': <serde message>` (with a `Fix the errors and try again.` suffix, as pydantic-ai appends) and `Unknown tool name: 'x'. Available tools: …` — where the available list names exactly the executable tools advertised on the turn that produced the call (including dynamically retrieved ones), sourced from `AgentRun` serialized state so resumed runs report the same set. Tracing still logs the full error chain.
5. **`PromptError::MaxTurnsError` displayed as `MaxTurnError: (reached max turn limit: n)`** — fixed to match the type name.

## Strict-review hardening

A close review against pydantic-ai surfaced more issues, fixed here (each design was adversarially checked against simpler alternatives that turned out to be defective):

- **MCP tools could silently evict a user's tool.** An MCP tool sharing a name with a directly-registered tool overwrote it (last-wins) and a later `tools/list_changed` refresh then deleted it wholesale — silently (the refresh-loop warning is dead code, since `remove_tool` is infallible). Now tracked by ownership: `ToolServerHandle::add_tool_owned`/`remove_tool_owned` + an `OwnerId` tag mean a refresh removes only the handler's own tools and never a name a user has taken over. A handler-local skip list was rejected — it breaks reconnect and the reverse registration order, and races. Adds pydantic-ai-style namespacing (`McpClientHandler::with_tool_prefix`): the prefix is advertised to the model while the MCP server is still called with the original name.
- **The rmcp adapter's remaining content gaps.** `is_error` results now surface their text even when mixed with non-text parts (was: collapsed to "No message returned"); audio content degrades to a text placeholder instead of failing the whole call and discarding co-returned content; resources use a labeled `resource {uri} ({mime}):\n{body}` format instead of an unprincipled concatenation.
- **`McpTool::definition`** delegates to the single `schema_as_json_value` conversion path (dropping an `unwrap_or_default` that could silently emit an empty schema).

## Deliberately not fixed here

- **No way to keep forcing a tool call on *every* turn.** The wire relaxation above (Required/Specific → Auto after the first tool-call turn) is what lets a forced multi-turn run terminate, but it is unconditional: a caller who genuinely wants the forced choice re-sent on every turn (e.g. a pipeline that must always call a tool) cannot express that today. The relaxation is also implicit — it silently changes the wire choice from what the caller configured (validation still uses the configured choice, so the `Specific` allowlist is unaffected). The more principled fix follows pydantic-ai: a tool-shaped exit (a `final_result`/output tool that is always available, so the model can finish while forcing stays on every turn), to be added separately. Until then, drive turns manually via `AgentRun`, or pair `Required` with your own terminal tool. `Specific` still enforces its allowlist every turn; `None`/`Auto` are unchanged.

## Breaking changes

Marked with `!` since these behaviors observable by callers and by models change:

1. **Tool-failure text sent to the model changed format** (see fix 4). Code string-matching the old wrapper text must update.
2. **Registering two tools under one name is now last-wins** (direct registrations) instead of failing the request at the provider; **MCP-vs-direct name collisions are skipped** by `McpClientHandler` (use `with_tool_prefix`).
3. **MCP resource tool results use a new `resource {uri} ({mime}):\n{body}` format**, and **MCP audio content becomes a text placeholder** instead of erroring.
4. **`ToolChoice::Required`/`Specific` no longer force a tool call on every turn of a multi-turn run.** After the first turn that produces tool calls, the wire choice relaxes to `Auto` so the run can terminate with text (see the **Note on tool choice**). The configured choice is unchanged for validation (the `Specific` allowlist still applies every turn) and for snapshots/diagnostics, but the request sent to the provider differs from main, which re-sent the forced choice every turn. Callers depending on forced-every-turn behavior must drive turns manually via `AgentRun` or add their own terminal tool.

## Test-reliability fixes included

Two pre-existing flakes surfaced while validating, both root-caused rather than retried away:

- `deepseek::permission_control` tests raced each other on a shared `./test.txt` (one test's cleanup deleted the other's fixture mid-run). Each test now uses a unique temp path; wire strings are unchanged, so the cassettes are untouched.
- The `stream_prompt_records_*_on_chat_spans_*` tests intermittently captured zero spans. Cause (confirmed by instrumenting the span creation site): `tracing` computes a callsite's cached interest on FIRST registration from the registering thread's dispatcher, so a parallel test without a subscriber can permanently cache `Interest::never` for the production callsites these tests assert on. Fixed by serializing scoped-subscriber tests behind a shared guard and having the harness warm the driver's callsites from its own thread plus rebuild the interest cache before asserting. 15 consecutive full-suite runs pass where roughly one in three failed before.

## Follow-ups (intentionally not in this PR)

- **A tool-shaped exit for forced tool choice.** See *Deliberately not fixed here* — the pydantic-ai-style `final_result`/output tool that lets `Required`/`Specific` runs finish *without* the implicit wire relaxation, so forcing can stay on every turn. Intended to supersede the current relax-to-`Auto` behavior.
- **Per-tool failure retry budget.** The `Fix the errors and try again.` suffix encourages retrying a failing tool, but nothing bounds the loop except `max_turns` (pydantic-ai pairs the same suffix with a per-tool `max_retries`). The fix needs a breaking `PromptError` variant and a newly recorded cassette, so it ships separately as `feat(agent)!: per-tool failure budget`.
- **`ToolResultContent::Audio`.** A core type addition so MCP audio can become a real audio part instead of a placeholder (pydantic-ai maps it to `BinaryContent`).
- **Structured validation-error payloads.** pydantic-ai sends a JSON array of error details; rig pins serde's one-line message. Richer payloads would be a separate, breaking text change.
- **Eager-error tool registration.** pydantic-ai raises a `UserError` on duplicate names at registration; rig chose last-wins-with-warning. Open for maintainer preference.

## Test plan

- `cargo test --test gemini --features "image audio rmcp derive"` — 116/116 replay green without provider keys, including the native-vs-rmcp shared-cassette parity pair.
- `cargo test -p rig-core --lib --features rmcp` — 900 passed, including coverage for ordering/dedup/merge semantics, ownership-tagged registration (collision skip, reverse-order protection, reconnect reclaim, prefix), failure-text formats, advertised-tool-name accuracy surviving serialization, forced-choice wire relaxation with validation and the relaxation flag pinned across serialization, and MCP image/hybrid/interleaving/error/resource/audio mapping.
- `cargo clippy --workspace --all-targets` and `cargo clippy --all-features --all-targets` — clean.

The commits are cleanly separable (tests first, then fixes, then the strict-review hardening) — glad to split this into smaller PRs if that's easier to review.


canardleteer:
I like the cassette suite and the rmcp hardening — especially ordered multimodal parts, the `ToolResultContent`-shaped image path, and the resource/audio degradation fixes in `McpTool::call`. That closes a lot of the “lossy flattening” pain we were hitting on MCP tools.

One follow-up I’d like to align on **before** opening a separate PR/issue (or contributing here if you prefer):

**`structuredContent` + pluggable adaptation**

MCP `CallToolResult` can carry `structuredContent` (typed JSON matching `outputSchema`) alongside `content[]`. From what I can see in this PR, the adapter still targets the string/`parts` path the agent loop ingests via `ToolDyn::call → String` — which is fine for backward compatibility — but **`structuredContent` still looks unmapped**.

I'm interested in a small, additive **`McpResultAdapter`** (name bikesheddable):

- **Default:** reproduces post-#1902 behavior byte-for-byte (so cassettes stay valid).
- **Optional:** e.g. `StructuredPreferredAdapter` / host-selected adapter at `rmcp_tool` registration.
- **Not** changing `ToolDyn` globally — adapter output still serializes to the same `String` boundary unless/until the loop grows a richer tool-result path.

Motivation: different providers want different shapes (plain text vs structured JSON vs preserved parts), and later MCP async (tasks, MRTR continuations) must not assume every outcome is a plain `CallToolResult` text blob.

**Questions for you @gold-silver-copper:**

1. Is this in scope for the rmcp migration follow-up you mentioned, or should I file a standalone `feat:` issue?
2. Would you want adapter selection wired now (per-`McpTool` builder knob), or wait until host-visible tool metadata exists (native vs MCP, `outputSchema`, `taskSupport`)?
3. Any objection to preserving `structuredContent` as JSON in the default string when present (even before a full adapter trait)?

Happy to draft the issue or a minimal follow-up PR on top of this branch once you point at preferred scope. Not planning to duplicate the cassette work here.

gold-silver-copper:
@canardleteer No objections to preserving StructuredContent, on the surface it sounds great. I would prefer it as an issue first with no PR, the tool server and general tools implementation in Rig is going to be seeing a fairly substantial refactor soon, so I'd prefer to not add to much onto it before that happens. But an Issue for posterity would be much appreciated. Closing this PR atm since it was just a draft, the first half of it was merged as #1913 and #1915

PR #1903, opened 2026-06-12
Avoid double slash in provider URLs

eriktews:
Only append a slash to base_urls of api providers when they don't already end with a slash.


gold-silver-copper:
lgtm

PR #1912, opened 2026-06-15
feat: Migrate to package-based examples to avoid dev-dependency build failures and slow builds

### LINKED ISSUE #1911: Proposal: Migrate to package-based examples to avoid dev-dependency build failures and slow builds
# Proposal: Migrate to package-based examples to avoid dev-dependency build failures and slow builds

## Problem Statement
Currently, running a simple example like `cargo run --example agent` triggers the compilation of all `[dev-dependencies]` in the `rig` workspace.

This introduces two main issues:
1. **Compilation Failures due to External Tools:** Some dev-dependencies require external system tools (such as `protoc` for certain integration crates). If a user does not have these tools installed, the compilation fails, preventing them from running even basic, unrelated examples.
2. **Slow Build Times:** Users must compile a large set of unrelated dev-dependencies just to run a basic example, which makes the initial build slow.

## Proposed Solution: Package-Based Workspace Examples
We can migrate the examples into individual packages (workspace member crates) under the `examples/` directory (e.g., `examples/agent`). This isolates each example's dependency tree, avoiding build failures from unrelated dev-dependencies and significantly reducing compilation times.

### Expected Behavior Phases:

* **Before Migration (Current State):**
  ```bash
  cargo run --example agent
  ```
  *(Compiles all workspace dev-dependencies; slow to build and fails if system tools are missing).*

* **During Migration (Intermediate State):**
  Both standard examples and migrated package examples will work as we transition them one by one:
  ```bash
  # For non-migrated examples
  cargo run --example some-other-example

  # For migrated examples
  cargo run -p example-agent
  ```

* **After Migration (Final State):**
  All examples run as individual packages:
  ```bash
  cargo run -p example-agent
  ```

## Current Progress
I have already started working on this migration and will open a Pull Request shortly to transition the first few examples. Please let me know if you have any initial feedback or suggestions on this direction.

Atliac:
Closes #1911

## Description

This PR reorganizes the repository structure by moving the main facade package to a dedicated directory and introduces a faster, package-based approach for running examples to improve developer experience.

No breaking changes made.

## Key Changes

1. **Workspace Reorganization**:
   - Moved the `rig` facade package from the root directory to `crates/rig`. This keeps the repository root clean and consistent with typical multi-crate workspace layouts.

2. **Standalone Example Crate**:
   - Created a new `example-agent` package under the workspace (copied from `crates/rig/examples/agent.rs`).
   - Running `cargo run -p example-agent` provides faster compilation and execution times during local development compared to the standard `cargo run --example agent` command.

3. **Documentation Updates**:
   - Added guidance in `CONTRIBUTING.md` explaining how to create and run new example packages:
     ```bash
     # To create a new example package:
     cargo new --name example-agent examples/agent

     # To run the example package:
     cargo run -p example-agent
     ```

## To Verify

- Run `cargo run --example agent`
- Run `cargo run -p example-agent`


gold-silver-copper:
Within the example packages Cargo.toml, publish= false should be set. Similar to how ratatui does it: https://github.com/ratatui/ratatui/blob/main/examples/apps/canvas/Cargo.toml

otherwise LGTM

Atliac:
Given that we use `release-plz`, we can set the default publishing behavior to false and selectively enable it only for the crates under the `crates/` directory. This is safer and avoids future maintenance overhead. I'll implement this change.

gold-silver-copper:
Your recent merge broke a few of the gemini tests. Also  I think the rig -> crates/rig migration should be a separate PR. And on further thought I'm not sure how neccesary it really is even, the current repo structure is inspired by bevy and it works well for them. I don't see much of a gain in the migration, so I'd push back on it unless there's a very good reason since there's already been a lot of churn in rig structure lately.

As for the examples migration, yeah I think that's very good. I'd prefer not touching release-plz though because its somewhat finnicky, and I'd prefer just explicitly setting publish=false for now. Again tho, should probably be a separate PR from the ->crates/rig migration. 

P.S. I also generally prefer these things to be opened as issues first so that this discussion can be had prior to you starting implementation. Also with architectural changes that require a lot of migration, such as the examples migration, I prefer doing them myself due to the review overhead required. But the issue and thought is much appreciated.

Atliac:
Thanks for the feedback.

Regarding the Gemini test failures, I believe they might have been introduced by a recent commit on `upstream/main` that was merged into this branch, rather than the migration changes themselves. 

For the `release-plz` configuration, my goal was to change the default behavior so that no crates are published unless explicitly allowed, while specifically enabling publishing only for the packages in the `crates` folder. I believe this is a safer and cleaner approach than manually setting `publish = false` on every new example, and it shouldn't cause any issues with the release process.

That said, I agree that the `rig -> crates/rig` migration is not necessary. In fact, since the entire migration isn't needed right now, I will go ahead and close this PR and the corresponding issue. 

(As an aside, the original intent of the PR was to try and optimize the structure for subsequent development, but I completely understand the preference to avoid unnecessary churn at this stage.)

Finally, thank you again for the guidance.

gold-silver-copper:
Just to be clear, I really do like the idea of separate packages for each example, I think that could speed up development and testing a lot. Feel free to open a new issue/pr for that specifically, I will soon if you don't :)

I think in a smaller PR the release-plz change is acceptable, but I would prefer having publish=false set per package anyway, just to be extra explicit. Yeah it's not the cleanest but better safe than sorry.

Thanks again

Atliac:
Just a heads-up: this PR is functional. However, I am currently focusing on other tasks and won't be reopening this PR, as this refactor isn't strictly necessary.

If you'd still like to proceed with the migration yourself, you can try commit 6541b3403d15b048711e910fcafdeff0cd6d2c7d. It leaves the release-plz configuration untouched (though it doesn't have the latest upstream/main commits merged). Adding publish = false to the new examples should be all that is needed.

PR #1929, opened 2026-06-20
feat(agent): add OutputMode to compose structured output with tools (#1928)

### LINKED ISSUE #1928: bug: output_schema is applied to every agent turn, suppressing tool calls in tool + structured-output agents

## Bug Report

When an agent has **both** `output_schema` and tools, rig applies the output schema (the provider's structured-output / `format` / `response_format` constraint) to **every** turn of the multi-turn agent loop — including the intermediate, would-be tool-call turns. Because providers force the model to emit schema-conforming output on those turns, the model is biased to **answer immediately instead of calling its tools** (and on some providers it is a hard error — see the pydantic-ai note below). The agent silently skips its tools and produces an often-hallucinated structured answer on turn 1.

### Where it happens

`prepare_completion_request` attaches the schema to the per-turn request unconditionally:

```rust
// crates/rig-core/src/agent/completion.rs ~149
.output_schema_opt(output_schema.cloned())
// ...and tools on the same request, ~214/224
.tools(tooldefs)
```

This builder runs on **every** loop iteration, so each request carries both the schema and the tools. The Ollama provider then maps the schema straight to the request `format` field (`crates/rig-core/src/providers/ollama.rs` ~537: `format: req.output_schema`); other providers map it to their own structured-output constraint (`response_format` for OpenAI/Azure, `response_mime_type` for Gemini, etc.).

## Reproduction

A deterministic cassette lands in #1927 (`tests/providers/ollama/cassette/agentic.rs`, scenario `agentic/structured_output_with_tools`):

```rust
let agent = client
    .agent("qwen3:4b")
    .preamble("You are a weather assistant. Use the get_weather tool to look up weather, then answer.")
    .tool(WeatherTool::new(call_count.clone()))     // tool is available
    .output_schema_raw(schema)                       // {city, summary}
    .additional_params(json!({ "think": true }))
    .default_max_turns(5)
    .build();

let response = agent
    .prompt("What is the current weather in Tokyo? Use the get_weather tool, then return the city and a one-sentence summary.")
    .await?;
```

Recorded live against a local Ollama (`qwen3:4b`):

- `get_weather` was invoked **0 times** — the model emitted the `{city, summary}` JSON immediately, with a fabricated summary, instead of calling the tool.
- The recorded request carries both `"format"` and `"tools"`, confirming rig sent the schema on the (only) turn.

This is **not** a parsing bug — rig constructs and parses the combined request correctly. The schema constraint on tool-eligible turns defeats the tools.

## Expected behavior

An agent configured with tools **and** `output_schema` should be able to call its tools across the intermediate turns and be constrained to the schema only when it produces its final answer.

---

## Analysis: this is the native-structured-output-vs-tools conflict

[pydantic-ai](https://ai.pydantic.dev/output/) frames the same problem with three output modes, which map directly onto rig:

| pydantic-ai mode | What it does | Composes with tools? | rig today |
|---|---|---|---|
| **`ToolOutput`** (their **default**) | the output type is registered as an *output tool* the model **calls**; it uses regular tools, then calls the output tool to finalize | ✅ yes | only in `Extractor` (see below) |
| **`NativeOutput`** | provider-native structured output — *"the model is forced to only output text matching the JSON schema"* | ❌ pydantic-ai warns: *"Gemini cannot use tools at the same time as structured output… will result in an error"*; other models silently skip tools | this is the agent's `output_schema` path — **the bug** |
| **`PromptedOutput`** | schema described in the prompt, model returns text, framework parses | ✅ yes (model chooses) | not available |

#1928 is a textbook instance of the documented native-output-vs-tools conflict: rig's agent only has `NativeOutput`, and it applies it alongside tools.

## rig already has the `ToolOutput` building blocks — in the Extractor

`crates/rig-core/src/extractor.rs` is *already* pydantic-ai's `ToolOutput` pattern:

```rust
const SUBMIT_TOOL_NAME: &str = "submit";              // :47
.tool(SubmitTool::<T> { _t: PhantomData })            // :334  — the output schema AS a tool
.tool_choice(ToolChoice::Required)                    // :335
// preamble: "Use the `submit` function to submit the structured data… ALWAYS CALL `submit`"
// then it filters the assistant turn for the `submit` call, extracts its args,
// deserializes, and retries if it wasn't called (:247–298).
```

The synthetic schema-shaped tool, call detection, argument extraction, and retry-on-failure exist and are battle-tested — they are simply not wired into the `Agent`'s `output_schema`/`output_schema_raw` path, which reaches for native `format` instead.

## Proposed fix: a configurable `OutputMode` (pydantic-ai parity)

Expose an `OutputMode` knob on the agent (and extractor) builders, mirroring pydantic-ai's three modes, defaulting to **tool-based output when the agent has function tools**:

```rust
#[derive(Clone, Debug, Default)]
pub enum OutputMode {
    /// Resolve at build time: `Tool` if the agent has function tools, else
    /// `Native`. Preserves current behavior for tool-less agents and fixes
    /// this issue for tool + schema agents.
    #[default]
    Auto,
    /// Output schema as a synthetic "output tool" the model calls
    /// (pydantic-ai ToolOutput). No native format constraint is sent.
    Tool,
    /// Provider-native structured output (`format`/`response_format`).
    /// Current behavior; may suppress tools on some providers.
    Native,
    /// Schema injected into the system prompt; the model returns text that
    /// rig parses (pydantic-ai PromptedOutput). Fallback for weak/local models.
    Prompted,
}
```

```rust
let agent = client.agent(model)
    .tool(read_file).tool(list_directory)
    .output_schema_raw(schema)            // Auto -> Tool because tools exist: works
    .build();

client.agent(model).tool(t).output_schema::<Out>().output_mode(OutputMode::Native);  // opt into native
client.agent(model).output_schema::<Out>().output_mode(OutputMode::Prompted);        // tiny-model fallback
```

### Behavior per mode

Resolve `Auto` → `Tool` if the toolset is non-empty, else `Native`, at request-preparation time (`agent/completion.rs::prepare_completion_request`), then:

- **`Tool`** — do **not** set `request.output_schema`. Instead inject a synthetic output tool whose `parameters` = the schema (generalize `SubmitTool` to a non-generic variant that takes a raw `schemars::Schema`, since `output_schema_raw` is schema- not type-based; its `call()` is a no-op sink — the args *are* the answer). Use `tool_choice = Auto` (not `Required`, so the model can call real tools first) and append a short instruction: *"When you have gathered enough information, call the output tool with the final structured result."* When the model calls the output tool, the loop treats it as terminal (`agent/run/mod.rs`, `agent/prompt_request/mod.rs`): validate its JSON args against the schema, on failure feed an error tool-result back and retry within `default_max_turns`, on success emit `Done(structured_string)`. The output tool is never executed.
- **`Native`** — current behavior: set `request.output_schema` (the existing per-provider mapping). The only mode that touches providers.
- **`Prompted`** — set neither: append the schema + instruction to the preamble (*"Respond with ONLY a JSON object matching this schema: …"*); tools remain available; parse the final text turn as JSON (tolerating ```` ```json ```` fences), retrying on parse failure.

### Notes

- `Tool` and `Prompted` are handled entirely in the agent layer and are **provider-agnostic** (no per-provider changes); only `Native` uses the per-provider mappings, which already exist. `Tool` mode also sidesteps the Gemini "tools + structured output = error" case.
- **No behavior change for tool-less agents/extractors**: `Auto` resolves to `Native` when there are no function tools (the extractor stays tool-based). The only change is the previously-broken tool + schema case, which now works.
- `output_mode` is additive — `output_schema` / `output_schema_raw` keep working.
- Reimplement `Extractor` on top of the same mechanism (it becomes "an agent with `OutputMode::Tool`"), removing the duplicated tool-output logic and letting extractor users pick `Native`/`Prompted` too.
- Default the parallel-output strategy to **early** (finalize as soon as the output tool is called; ignore concurrent tool calls); leave pydantic-ai's graceful/exhaustive strategies and multiple/union output types as follow-ups.

### Acceptance

- Unit: `Auto` resolves correctly; `Tool` adds the output tool and omits `request.output_schema`; `Native` sets it; `Prompted` augments the preamble only.
- Cassette regression: re-record `tests/providers/ollama/cassette/agentic.rs::structured_output_with_tools` under the default and assert the tool **is** invoked *and* the structured JSON is returned (a `Native`-mode variant can document the suppression).
- Existing extractor tests/cassettes still pass after unification; `cargo test`, `clippy --all-targets`, `fmt --check`, the `wasm32-unknown-unknown` check, and `cassette_safety` all stay green.

## Additional context

- Real-world impact: [repo-tagger](https://git.milliways.info/milliways-infosec-tools/repo-tagger) drives Ollama with `output_schema_raw` + two tools (`list_directory`, `read_file`) + `think:true` to explore a repository and emit schema-conforming tags. Because the schema is applied on every turn, the model frequently skips the exploration tools and answers from nothing — the app works around it by retrying the whole agent run up to 10×. Reported by Erik Tews.
- Provider-general (any provider mapping `output_schema` to a structured-output constraint), easiest to observe on Ollama. Surfaced while adding Ollama cassette coverage in #1927.
- References: pydantic-ai output modes — https://ai.pydantic.dev/output/ ; rig precedent — `crates/rig-core/src/extractor.rs` (`SubmitTool`, `tool_choice(Required)`, submit-call extraction/retry).

**Environment:** rig-core `main` (v0.39.0); Ollama provider; `qwen3:4b`; non-streaming agent loop with `default_max_turns > 1`.

gold-silver-copper:
## Summary

Fixes #1928. rig applies `output_schema` as the provider's **native** structured-output constraint (`format`/`response_format`) on **every** turn of the agent loop — including tool-call turns. That suppresses tool calls: the model emits schema JSON immediately instead of using its tools, so a tool + structured-output agent never calls its tools.

This adds an `OutputMode` knob on the agent builder (modeled on [pydantic-ai's output modes](https://ai.pydantic.dev/output/#structured-output)) so the structured output can be produced as a **tool call the model makes after using its tools**, which composes correctly:

```rust
let agent = client.agent(MODEL)
    .preamble("…")
    .tool(WeatherTool)
    .output_schema_raw(schema)
    // .output_mode(OutputMode::Native)   // opt back into the old behavior
    .build();
```

### `OutputMode`

| Mode | Behavior |
|------|----------|
| `Auto` *(default)* | Resolves to `Tool` when there's an output schema **and** ≥1 tool **and** the tool choice permits the output-tool call; otherwise `Native`. Only changes behavior for the broken tool + schema case. |
| `Tool` | Registers the schema as a synthetic, collision-safe output tool (`final_result`) the model calls to finalize. **No** native constraint, so the model can call its real tools first. The agent loop intercepts the call and finalizes with its arguments. (pydantic-ai `ToolOutput`.) |
| `Native` | Legacy native structured output — the escape hatch. (pydantic-ai `NativeOutput`.) |
| `Prompted` | Injects the schema into the preamble and parses the model's text — for weak/local models. (pydantic-ai `PromptedOutput`.) |

### Implementation notes

- Wired through **both** the streaming and non-streaming drivers via the shared `AgentRun` intercept (single decision point).
- Typed prompts (`prompt_typed::<T>()`) and the `Extractor` pin `Native`, so their existing behavior is unchanged (the `Extractor` already implements its own tool-based output via `SubmitTool`).
- The finalizing turn is persisted as the assistant's final **text** (keeping reasoning, dropping the output-tool call) so saved/returned history has no unanswered `tool_use` that providers reject when the conversation is replayed on a later turn.
- Tool mode **degrades to `Native`** when `tool_choice` forbids the output-tool call (`None`/`Specific`), so structured output is still enforced rather than silently dropped.
- Once committed, Tool mode is **pinned for the run**, so a per-turn tool-set change (e.g. RAG retrieval) can't flip back to the native constraint mid-run.

### Tests

- Unit tests for mode resolution, tool-choice degradation, output-tool name collision handling, and the run intercept (including history validity — no dangling `tool_use`).
- Live-recorded Ollama cassette tests (replay by default) for **Tool** mode (streaming + non-streaming, with a real tool actually invoked), **Native**, and **Prompted** modes.

All 896 rig-core lib tests pass; `clippy --all-features --all-targets`, `fmt`, and `cargo doc -D warnings` are clean; all Ollama (16) and Anthropic (72) cassettes replay green.

Closes #1928


gold-silver-copper:
## Update — addressed review feedback (commit `367e788`)

A detailed framework-comparison review (vs pydantic-ai / LangChain / Instructor) prompted a second round of changes. All are covered by unit tests + the existing live-recorded cassettes; `clippy --all-features --all-targets`, `fmt`, `cargo doc -D warnings`, and the Ollama (16) / Anthropic (72) / OpenAI (84) cassette suites are all green.

**Provider-aware `Auto`** — new `CompletionModel::composes_native_output_with_tools()` (default `false`; `true` for OpenAI chat-completions/responses + Anthropic). `Auto` now routes a tool + schema agent to `Tool` **only** on providers whose native constraint would suppress tool calls (e.g. Ollama), and keeps guaranteed **`Native`** structured output on providers that compose the two. So the fix no longer downgrades OpenAI/Anthropic from a hard native guarantee to best-effort.

**Stronger `Tool`-mode contract** — when the model finalizes without calling the output tool, or calls it with arguments missing required schema fields, the run **re-prompts** a bounded number of times (default 1) before finalizing best-effort. A plain-text turn that is already valid structured output is **accepted** instead of wasting a re-prompt. The retry budget resets on real-tool progress (per finalization attempt, not per run).

**Typed unification** — `prompt_typed::<T>()` now flows through `OutputMode` like the untyped API (it was pinned to `Native`); the typed `send()` deserializes tolerantly (strips prose / markdown fences).

**Robustness** — Tool mode is pinned for the run via the persisted `output_tool_name`, so the pin survives serialize/resume and a tool-less first turn can still resolve to `Tool` once tools appear (only `Tool` is pinned; `Native`/`Prompted` re-resolve). A mid-run real-tool name collision with the synthetic output tool is warned.

**⚠️ Behavior change to note for reviewers:** `prompt_typed::<T>()` **with tools** on a provider that does *not* compose native structured output (e.g. Ollama) now uses `Tool` mode (best-effort + bounded re-prompt) instead of the previously force-pinned `Native`. On OpenAI/Anthropic, typed + tools stays `Native`. Callers who relied on the native constraint on a non-composing provider can opt back in with `.output_mode(OutputMode::Native)`.

Deferred as follow-ups (documented, not blocking): a `jsonschema`-backed full validation pass (the current check is dependency-free, required-field-based), and a structural (non-name) marker for the output tool.

gold-silver-copper:
Correction to the note above (commit `167a006`): the **typed unification (#5) is reverted** — `prompt_typed::<T>()` stays pinned to `Native`. Routing the typed path through `OutputMode` changed typed + tools from the native constraint to a synthetic output tool, which broke the typed + tools cassettes of every OpenAI-compatible provider (recorded under the native path; not re-recordable here). Unifying the typed path is left as a follow-up.

Everything else stands: provider-aware `Auto`, the Tool-mode re-prompt/validation backstop, mode pinning via `output_tool_name`, and the doc/robustness fixes. The untyped `output_schema`/`output_mode` API composes structured output with tools.

PR #1952, opened 2026-06-25
feat(rig-core): wire tool call extensions through the agent loop + polish (stacked on #1537)

gold-silver-copper:
## Stacked on #1537

This PR is **stacked on top of #1537** (`feat: add ToolCallContext for per-call runtime context in tool dispatch`). Its base branch `pr-1537-base` is a mirror of #1537's head (`94401fe`), so the diff here shows **only the follow-up changes**. Review/merge #1537 first; this should be retargeted onto `main` once #1537 lands.

It addresses the review findings on #1537: the doctest was CI-breaking, the feature couldn't actually be reached from an agent run, and a few `http::Extensions`-parity/naming items.

### 1. The feature now actually reaches tools through `Agent::prompt()` (the main gap)
#1537 plumbed `call_*_with_context` through `ToolServerHandle → ToolSet → ToolType → ToolDyn → Tool`, but **both agent dispatch sites still called the no-context `call_tool`**, which injects an empty context — and there was no public API to attach one. So the motivating use cases (A2A `context_id`/`task_id`, per-call auth/session) were unreachable through `Agent::prompt()` / `stream_prompt()`.

- Added a `tool_extensions` field + `.with_tool_extensions(..)` builder on **both** `PromptRequest` and `StreamingPromptRequest` (per-call injection, mirroring pydantic-ai's per-run `deps=`; nothing is exposed to the model).
- Routed both dispatch sites to `call_tool_with_extensions`. In the concurrent blocking loop the extensions are shared via `Arc`, so they're not deep-cloned per tool call.
- Added **end-to-end tests** proving an injected value reaches a tool through the agent loop, and that the default path dispatches through `call_with_extensions` with an empty map.

```rust
#[derive(Clone)]
struct SessionId(String);

let mut extensions = ToolCallExtensions::new();
extensions.insert(SessionId("session-123".to_string()));

let answer = agent.prompt("do the thing").with_tool_extensions(extensions).await?;
```

### 2. Fix the CI-breaking doctest
The doctest used `use rig::tool::ToolCallContext;`, but the crate is `rig_core`, so `cargo test -p rig-core --doc` failed with `E0433: unresolved crate 'rig'`. Fixed to `rig_core::tool::ToolCallExtensions`.

### 3. Rename `ToolCallContext` → `ToolCallExtensions`
The old name collided with the dependency `rmcp::handler::server::tool::ToolCallContext` (a different concept) and with rig's existing `InvalidToolCallContext` / `Tool::Context`. The new name matches the `Extensions` vocabulary the maintainers cited as the target direction. Methods renamed to `call_with_extensions` / `call_tool_with_extensions`. Module `tool::context` → `tool::extensions`. Nothing is released yet, so this is not a breaking change to any published surface.

### 4. `http::Extensions` parity + cleanup
- `insert` now returns the displaced value (`Option<T>`), like `http::Extensions::insert`.
- `TypeId` keys are hashed with a no-op `IdHasher` instead of SipHash, like `http::Extensions` (with a panic-free byte fallback so it stays correct regardless of how `TypeId` hashes).
- Added `len` / `is_empty`.
- Deleted the dead `ToolType::call` shim (was silenced with `#[allow(dead_code)]`).
- Documented the `call` vs `call_with_extensions` override contract on `Tool`.

## Test plan
- [x] `cargo test -p rig-core --lib` — 900 passed, 0 failed (incl. new e2e + type-map tests)
- [x] `cargo test -p rig-core --doc` — 98 passed, 0 failed (the previously-failing doctest now passes)
- [x] `cargo clippy -p rig-core --lib --tests -- -D warnings` — clean
- [x] `cargo fmt -p rig-core -- --check` — clean
- [x] `cargo check -p rig-core --features rmcp` — clean
- [x] `cargo check -p rig-core --target wasm32-unknown-unknown --no-default-features --features wasm` — clean (exercises the `cfg(target_family = "wasm")` branch)

## Notes / possible follow-ups
- Per-call injection only (matches pydantic-ai). An agent-level default (set once on the builder, overridable per call) could be layered on later if wanted.
- A wasm build in CI would lock in the `target_family = "wasm"` branch (the crate currently has two wasm predicates: `target_family = "wasm"` here vs `all(feature = "wasm", target_arch = "wasm32")` for `WasmCompat*` — this code matches its immediate neighbors `ToolError`/`WasmBoxedFuture`).


gold-silver-copper:
Superseded by #1953, which rebases this work onto current `main` (where #1945 unified tool dispatch into `run_single_tool`) as a single, non-stacked PR. Closing in favor of #1953.

PR #1958, opened 2026-06-25
fix(openai-compat): normalize evicted tool-call string arguments to an object

wey-gu:
## Summary

The OpenAI Chat Completions-compatible streaming aggregator can emit a tool call
whose `function.arguments` is a bare JSON **string** instead of an object. When a
downstream serializer expects an object — e.g. the Anthropic protocol's
`tool_use.input` — it forwards the string verbatim, and strict providers reject
the request with:

```
tool_use.input: Input should be a valid dictionary
```

This breaks any multi-turn agent loop the moment a model emits more than one tool
call in a single assistant turn.

## Root cause

`send_compatible_streaming_request` accumulates streamed tool-call argument
fragments as a `serde_json::Value::String` (see `append_tool_call_arguments`).
There are two finalizers:

- `finalize_pending_tool_call` (end-of-stream / finish-reason path) already
  parses a `Value::String` into the underlying object.
- `finalize_completed_streaming_tool_call` (the **eviction** path — taken when a
  distinct, differently-named tool call starts in the same turn and evicts the
  previous one) only normalized `Value::Null` -> `{}` and passed everything else
  through verbatim.

So an evicted tool call leaks its accumulated arguments **string** straight into
`ToolCall.function.arguments`, and the object-typed downstream serializer emits a
string.

## Repro

A streaming turn with two distinct tool calls (the second evicts the first):

```
tool_call[0]: id=call_aaa name=search args="{\"query\":\"one\"}"   <- evicted
tool_call[1]: id=call_bbb name=search args="{\"query\":\"two\"}"
```

The evicted `call_aaa` is finalized via `finalize_completed_streaming_tool_call`
and emerges with `arguments = String("{\"query\":\"one\"}")` rather than the
object `{"query":"one"}`. Forwarding that to an Anthropic-style endpoint yields
the `Input should be a valid dictionary` rejection.

## Fix

Make the eviction finalizer mirror `finalize_pending_tool_call`: parse a
`Value::String` payload into the underlying object, normalizing empty,
unparseable, or valid-but-non-object payloads to `{}` so the emitted
`function.arguments` is always a dictionary. `Value::Null` -> `{}` behavior is
preserved.

Single-function change in
`crates/rig-core/src/providers/internal/openai_chat_completions_compatible.rs`.

## Tests

- `eviction_finalizer_parses_string_arguments_into_object`
- `eviction_finalizer_normalizes_empty_and_unparseable_strings_to_object`
- `evicted_tool_call_emits_object_input_end_to_end`

`cargo test -p rig-core` passes (17/17 in the affected module).

### DIFF
--- a/crates/rig-core/src/providers/internal/openai_chat_completions_compatible.rs
+++ b/crates/rig-core/src/providers/internal/openai_chat_completions_compatible.rs
@@ -474,8 +474,33 @@ fn append_tool_call_arguments(tool_call: &mut RawStreamingToolCall, chunk: &str)
 pub(crate) fn finalize_completed_streaming_tool_call(
     mut tool_call: RawStreamingToolCall,
 ) -> RawStreamingToolCall {
-    if tool_call.arguments.is_null() {
-        tool_call.arguments = serde_json::Value::Object(serde_json::Map::new());
+    // The eviction path (distinct-named tool calls within one assistant turn)
+    // previously only normalized a `null` arguments value to `{}` and otherwise
+    // passed the value through verbatim. Streamed OpenAI-compatible tool-call
+    // arguments accumulate as a JSON *string* (`Value::String`), so an evicted
+    // tool call leaked a bare string into `ToolCall.function.arguments`. A
+    // downstream serializer that expects an object (e.g. the Anthropic protocol's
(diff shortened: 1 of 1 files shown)


gold-silver-copper:
Awesome PR, thanks!

PR #1964, opened 2026-06-26
fix(anthropic): coerce tool_use.input to an object at the send boundary

wey-gu:
fix(anthropic): coerce tool_use.input to an object at the send boundary

The Anthropic API requires `tool_use.input` to be a JSON object and rejects
anything else with a deterministic 400:
`messages.N.content.M.tool_use.input: Input should be a valid dictionary`.

Both assistant-content conversion sites (`Message::try_from` and
`anthropic_content_from_assistant_content`) forwarded
`ToolCall.function.arguments` verbatim into `Content::ToolUse { input, .. }`.
`arguments` is provider-shaped `serde_json::Value`: depending on how the
turn was built (replayed history, an evicted streamed tool call, a provider
that accumulates arguments as a JSON string), it can arrive as a
`Value::String`, `Value::Null`, or another non-object. Anthropic then
rejects the whole multi-turn request, which breaks every tool conversation
on Anthropic-shaped endpoints (including third-party gateways that speak the
Anthropic protocol).

This is the Anthropic-serialization counterpart of the openai-compat
eviction fix (#1958): #1958 stops a string leaking out of the eviction
source; this normalizes at the Anthropic send boundary so the contract
holds no matter how `arguments` was produced upstream.

`coerce_tool_input`:
- object -> passed through unchanged
- JSON string encoding an object -> parsed into that object
- non-JSON string / non-object JSON string / null / array / number / bool
  -> `{}` (the only shape the API accepts)

Adds unit coverage for each branch.

## Verification

- New unit test `coerce_tool_input_normalizes_non_object_arguments` pins each
  branch (object passthrough, JSON-string-of-object parse, and
  string/null/array/number/bool -> `{}`).
- `cargo test -p rig-core --lib coerce_tool_input` -> 1 passed.
- Both `Content::ToolUse` construction sites
  (`Message::try_from` and `anthropic_content_from_assistant_content`) route
  through `coerce_tool_input`, so the object invariant holds for replayed
  history and freshly built turns alike.

### DIFF
--- a/crates/rig-core/src/providers/anthropic/completion.rs
+++ b/crates/rig-core/src/providers/anthropic/completion.rs
@@ -1062,7 +1062,7 @@ impl TryFrom<message::AssistantContent> for Content {
                 Ok(Content::ToolUse {
                     id,
                     name: function.name,
-                    input: function.arguments,
+                    input: coerce_tool_input(function.arguments),
                 })
             }
             message::AssistantContent::Reasoning(reasoning) => Ok(Content::Thinking {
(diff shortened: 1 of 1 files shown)


gold-silver-copper:
Good PR!

PR #1969, opened 2026-06-26
feat(agent): inject messages into a running agent run

gold-silver-copper:
## Summary

Reimplements mid-run **message injection** (originally #1858) from scratch on the new `AgentRunner`/`AgentRun`/`AgentHook` architecture, hardened across two rounds of adversarial review.

```rust
let mut runner = agent.runner("Investigate and report.").max_turns(10);
let injector = runner.message_injector();           // cloneable, Send
let run = tokio::spawn(async move { runner.run().await });
injector.inject("Heads up: the deploy just failed, factor that in.")?;
let response = run.await??;
```

## How it works

- `message_injector()` on `AgentRunner` (and `PromptRequest`/`StreamingPromptRequest`) returns a cloneable `MessageInjector`. `inject(impl Into<Message>)` pushes onto a `futures::mpsc` queue the driver drains before each model call and folds into the sans-IO `AgentRun` via `inject_message()` — identical on the blocking and streaming drivers.
- **Delivered as its own user turn** (never merged into a tool-result message — OpenAI's converter drops that mixed shape).
- **Best-effort / in-flight**: folds in before the next model call; if the run finishes first it's dropped (logged) and a later `inject` returns `MessageInjectError::RunFinished`. No keep-alive — injection does not resurrect a finished run. Only user turns are delivered; injections never displace an output-mode reprompt or invalid-tool-retry feedback.

## Cross-provider coalescing

Delivering a separate user turn can place it next to a prior user turn (a tool-result turn, RAG documents, or a hoisted-system gap). To keep that valid on every provider, a new shared `providers::coalesce` pass merges adjacent same-role **user** turns on each provider's converted message list:

- Applied across in-core providers (anthropic, gemini, openai chat + responses, cohere, mistral, ollama, deepseek, azure, groq, together, llamafile, xai; openai-compatible providers inherit it) **and rig-bedrock**.
- **Required** by AWS Bedrock's strictly-alternating Converse API; a no-op for already-alternating histories elsewhere.
- **Parallel tool-result messages are never coalesced** (they carry distinct `tool_call_id`s) — each provider's impl matches only user turns.
- Side benefit: normalizes RAG-documents+prompt and hoisted-system gaps into single user turns.

## Design / prior art

Delivery follows **pydantic-ai's `AgentRun.enqueue`**; the fold-between-steps point matches **Vercel AI SDK's `prepareStep`** and **LangGraph's `Command(resume=…)`** — without a checkpointer (in-process for the run's lifetime).

This was developed with two rounds of multi-agent adversarial review, which caught the OpenAI-drop (v1 merge), the keep-alive budget bug, and the Gemini/Bedrock consecutive-user gap.

## Testing

AgentRun unit (push-separate delivery, fold order, ExecutingTools serde, reprompt/retry non-derailment, non-user drop); driver tests (inject-before, mid-tool-loop at the request level, multi-producer, RunFinished); an OpenAI-converter regression pinning that the separate-turn shape survives where a merged tool-result+text turn drops the text; per-provider coalesce + parallel-tool-results-preserved tests; the `agent_with_message_injection` example. `cargo test -p rig-core` (1073) + `rig-bedrock` (86) pass; clippy `--all-targets` + fmt clean.

Closes #1858.


lazytanuki:
Hi !
Will this feature be implemented in another PR or is it just unwanted?
Thanks

gold-silver-copper:
> Hi ! Will this feature be implemented in another PR or is it just unwanted? Thanks

I think this feature is wanted, this was just me playing around with adding it. I will open another PR eventually, unless you want to redo yours

lazytanuki:
No problem go ahead !

PR #1982, opened 2026-06-30
feat(ollama): Extend `think` options with `max`

m-dreiling:
Add the `max` option to the `think` parameter as described in the Ollama capabilities: https://docs.ollama.com/capabilities/thinking


gold-silver-copper:
AWESOME

PR #1985, opened 2026-06-30
refactor(agent): unify the streaming and non-streaming drivers over one engine

gold-silver-copper:
## Summary

The non-streaming (`AgentRunner::run`) and streaming (`AgentRunner::stream`) agent
drivers each hand-rolled their own ~drive loop around the shared sans-IO
`AgentRun` machine — two near-identical loops that had to be kept in lock-step by
hand (the recurring source of run/stream drift). This collapses them onto **one
shared engine**, `drive_agent`, leaving only the genuinely irreducible per-medium
piece — the model transport — behind a small `TurnSource` trait.

After this change both surfaces share the entire loop: outer `next_step`
dispatch, the `CompletionCall` hook + request preparation, invalid tool-call
recovery wiring, tool execution, conversation memory, and finalization. `run()`
is now a fold over the engine; `stream()` forwards it.

This continues the driver-unification arc started by the `AgentRun` extraction
(#1899): that landed the shared *decision* machine; this lands the shared *driver*.

## How it's structured (reviewable commit-by-commit)

1. **Extract shared helpers; pin blocking span topology.** `acquire_agent_span`,
   `resolve_completion_call`, `append_run_messages` are shared by both drivers.
   Adds a span safety-net test pinning `run()`'s span shape (name `chat`,
   `invoke_agent` creation, the `follows_from` chain, and `created_agent_span`-gated
   usage) so the later cutover can't silently drift it — the streaming side was
   already pinned by `assert_stream_usage_recorded_on_chat_spans`.

2. **Introduce `drive_agent` + `TurnSource`; route streaming through it.** The
   medium-independent outer loop lives in `drive_agent`; `StreamingTurnSource`
   holds the assembler-driven turn, per-delta hot-path gating, mid-stream
   invalid-tool recovery, and the `chat_streaming` span. `stream()` becomes a thin
   shell.

3. **Drive `run()` as a fold over the engine.** `UnaryTurnSource` keeps the unary
   `model.completion()` transport, the `chat` span + `follows_from` chain, and
   run-level recording. `run()` folds `drive_agent` to its `Done` item.

4. **Unify tool execution into `drive_tool_calls`.** One implementation for both
   surfaces (the span-chaining difference is a closure parameter); the
   blocking-only `execute_tools_buffered` is deleted.

## What is deliberately *not* unified

The `CallModel` transport stays two-bodied behind `TurnSource`: streaming advances
the machine incrementally (`record_streamed_completion_call` / `streamed_turn` /
`resolve_streamed_invalid_tool_call`) while yielding deltas mid-stream, whereas
the unary path feeds a whole `ModelTurn` to `model_response`. That divergence is
intrinsic to the medium and is the one thing the trait isolates rather than
shares. Transport is kept decoupled from surface on purpose — `run()` still issues
a unary provider call (no "completion = drain the stream"), so providers whose
streaming and non-streaming endpoints differ see byte-identical behavior.

## Behavior changes (all intentional, called out for review)

- **Outer-span usage guard.** `run()` previously recorded run-level
  usage/completion onto a *caller-supplied* outer span unconditionally; it now
  guards on `created_agent_span`, matching `stream()`. A caller-owned span is no
  longer polluted. (No test pinned the old behavior; a new test pins the new one.)
- **Setup errors are surfaced.** A provider-stream / request-build error inside
  the streaming generator was silently swallowed by `?` (whose `Err` ends an
  `async_stream!` without yielding); these now surface as an error item, which
  the blocking fold also needs.
- **Cosmetic logs.** `run()`'s per-turn log now reads "Current conversation
  Turns: n/m" (was "…depth…"); the blocking "Depth reached" and streaming "Agent
  multi-turn stream finished" run-completion markers are unified into one shared
  "Agent run finished" event.
- **Blocking tool execution at `tool_concurrency > 1`** now uses
  `buffer_unordered` + call-order reindex (was `buffered`). Persisted history
  stays call-ordered; only internal completion ordering differs (unobservable to
  `run()`).
- **Multi-tool terminate → run-all-then-decide (streaming change only).** When a
  turn emits several tool calls and one tool's hook returns `Flow::Terminate`,
  both surfaces now run **all** of the turn's tools (each independently
  hook-gated, firing its hooks) before surfacing the lowest-call-index terminate
  error. `run()` is unchanged — this restores its pre-refactor `buffered(1)` +
  `collect`-all semantics; the change is that `stream()` no longer stops early at
  `tool_concurrency(1)`. This makes the two lock-step at **any** concurrency and
  matches the dominant agent-framework behavior (semantic-kernel, LangGraph,
  Vercel AI SDK, LangChain, pydantic-ai all run-all for a deliberate abort; only
  openai-agents short-circuits). Streaming's *surfaced items* are unchanged (post-
  error items are suppressed exactly as before); only a sibling's tool body +
  hooks now run. Pinned by a regression test.

## Invariants preserved

- Public API unchanged: `AgentRunner::run`/`stream`, `MultiTurnStreamItem`,
  `FinalResponse`, `PromptResponse`, the hook surface, and the sans-IO `AgentRun`
  surface keep their signatures. The new engine types are `pub(crate)`.
- The sans-IO serialize/resume keystone is untouched — `AgentRun` remains the
  sole owner of serializable state; `TurnSource`/sources hold only per-call
  presentation ephemera.
- R1 span identity: per-turn usage stays on the chat span and run-level recording
  on the agent span, each per-medium; `stream()` instruments the engine with the
  agent span, `run()` runs under the ambient span (agent span detached + linear
  `follows_from` chain). `test_span_context_isolation` semantics preserved.
- WASM: the new trait and `DriveStream` alias use the existing `WasmCompatSend`
  markers and cfg'd `Send` bounds; no `tokio`, no spawn.

## Testing

- `cargo test -p rig-core` — 1052 passed, 0 failed (incl. the full run/stream
  parity suite, the streaming usage-span tests, and the new blocking span test).
- `cargo clippy -p rig-core --all-targets` — clean (workspace's strict lint set).
- `cargo fmt` — clean.
- `cargo test -p rig --all-features --test gemini agent_run` (cassette replay) —
  18 passed, 0 failed (covers `agent_run_streamed::*`, `agent_run_resume::*`
  serialized-state resume, recovery, and hand-driven stepping — both streaming
  and non-streaming through real Gemini response shapes).

## Follow-ups enabled (not in this PR)

Because transport and surface are now independent type parameters of one engine,
two new capabilities fall out of a one-line type swap and could be added later as
additive API: streaming agent-level events over a *non-streaming* model
(`CompletionCall` + `FinalResponse`, no token deltas), and collecting a final
`PromptResponse` over a *streaming-only* provider. Left out here to keep the PR
focused on the refactor and free of new public surface.

This is a large diff for a refactor; happy to split it (e.g. land commits 1–2,
then 3–4) if the maintainers prefer smaller PRs.

### DIFF
--- a/crates/rig-core/src/agent/prompt_request/streaming.rs
+++ b/crates/rig-core/src/agent/prompt_request/streaming.rs
@@ -633,6 +633,204 @@ where
     }
 }
 
+/// Execute a turn's tool calls, shared by both surfaces.
+///
+/// Emits a `ToolCall` item before each tool runs and a `ToolResult` item as each
+/// finishes, feeding the (call-ordered) results back into the machine. The tool
+/// items carry no raw provider response, so this is generic over the surface's
+/// `R`. `chain_tool_span` lets the blocking surface chain spans into its linear
+/// `follows_from` sequence (the streaming surface returns the span unchanged).
+pub(crate) fn drive_tool_calls<'a, M, R, F>(
+    runner: &'a AgentRunner<M>,
+    run: &'a mut AgentRun,
+    calls: Vec<PendingToolCall>,
+    chain_tool_span: F,
+) -> DriveStream<'a, R>
+where
+    M: CompletionModel,
+    R: WasmCompatSend + 'a,
+    F: Fn(tracing::Span) -> tracing::Span + WasmCompatSend + 'a,
+{
+    Box::pin(async_stream::stream! {
+        let full_history_for_errors = run.full_history();
+
+        if runner.concurrency <= 1 {
+            // Sequential default: one ToolCall/ToolResult pair per tool, strictly
+            // interleaved. Byte-identical to the pre-concurrency behavior.
+            let mut results: Vec<UserContent> = Vec::with_capacity(calls.len());
+
+            for pending in calls {
+                let tool_call = pending.tool_call;
+                if let Some(result) = pending.preresolved_result {
+                    results.push(result);
+                    continue;
+                }
+                let internal_call_id =
+                    pending.internal_call_id.unwrap_or_else(crate::id::generate);
+
+                let tool_span = chain_tool_span(new_execute_tool_span());
+
+                yield Ok(MultiTurnStreamItem::stream_item(
+                    StreamedAssistantContent::ToolCall {
+                        tool_call: tool_call.clone(),
+                        internal_call_id: internal_call_id.clone(),
+                    },
+                ));
+
+                let result = run_single_tool(
+                    &runner.hooks,
+                    &runner.tool_server_handle,
+                    &runner.tool_extensions,
+                    &tool_call,
+                    &internal_call_id,
+                    &full_history_for_errors,
+                )
+                .instrument(tool_span)
+                .await;
+
+                match result {
+                    Ok(content) => {
+                        results.push(content.clone());
+                        if let UserContent::ToolResult(tool_result) = content {
+                            yield Ok(MultiTurnStreamItem::StreamUserItem(
+                                StreamedUserContent::ToolResult {
+                                    tool_result,
+                                    internal_call_id,
+                                },
+                            ));
+                        }
+                    }
+                    Err(err) => {
+                        yield Err(StreamingError::Prompt(Box::new(err)));
+                        return;
+                    }
+                }
+            }
+
+            if let Err(err) = run.tool_results(results) {
+                yield Err(Box::new(err).into());
+                return;
+            }
+        } else {
+            // Concurrent path: emit every ToolCall stream item eagerly in call
+            // order, then run tools concurrently and surface each ToolResult as
+            // soon as its tool finishes. Results are still stored by original
+            // call index, so persisted history remains deterministic.
+            let call_count = calls.len();
+            let mut indexed_calls: Vec<(usize, PendingToolCall, tracing::Span, Option<String>)> =
+                Vec::with_capacity(call_count);
+
+            for (index, mut pending) in calls.into_iter().enumerate() {
+                let tool_span = chain_tool_span(new_execute_tool_span());
+
+                if pending.preresolved_result.is_some() {
+                    indexed_calls.push((index, pending, tool_span, None));
+                    continue;
+                }
+
+                let internal_call_id = pending
+                    .internal_call_id
+                    .clone()
+                    .unwrap_or_else(crate::id::generate);
+                pending.internal_call_id = Some(internal_call_id.clone());
+
+                yield Ok(MultiTurnStreamItem::stream_item(
+                    StreamedAssistantContent::ToolCall {
+                        tool_call: pending.tool_call.clone(),
+                        internal_call_id: internal_call_id.clone(),
+                    },
+                ));
+
+                indexed_calls.push((index, pending, tool_span, Some(internal_call_id)));
+            }
+
+            let unordered = stream::iter(indexed_calls)
+                .map(|(index, pending, tool_span, yield_id)| {
+                    let hooks = &runner.hooks;
+                    let tool_server_handle = &runner.tool_server_handle;
+                    let tool_extensions = &runner.tool_extensions;
+                    let full_history_for_errors = &full_history_for_errors;
+                    async move {
+                        if let Some(result) = pending.preresolved_result {
+                            return (index, yield_id, Ok(result));
+                        }
+
+                        let internal_call_id =
+                            pending.internal_call_id.unwrap_or_else(crate::id::generate);
+                        let result = run_single_tool(
+                            hooks,
+                            tool_server_handle,
+                            tool_extensions,
+                            &pending.tool_call,
+                            &internal_call_id,
+                            full_history_for_errors,
+                        )
+                        .await;
+                        (index, yield_id, result)
+                    }
+                    .instrument(tool_span)
+                })
+                .buffer_unordered(runner.concurrency);
+            futures::pin_mut!(unordered);
+
+            let mut results: Vec<Option<UserContent>> = Vec::with_capacity(call_count);
+            results.resize_with(call_count, || None);
+            let mut first_error: Option<PromptError> = None;
+            while let Some((index, yield_id, result)) = unordered.next().await {
+                if first_error.is_some() {
+                    continue;
+                }
+                match result {
+                    Ok(content) => {
+                        let Some(slot) = results.get_mut(index) else {
+                            first_error = Some(PromptError::CompletionError(
+                                CompletionError::ResponseError(
+                                    "tool execution returned an invalid result index".to_string(),
+                                ),
+                            ));
+                            continue;
+                        };
+                        *slot = Some(content.clone());
+                        if let Some(internal_call_id) = yield_id
+                            && let UserContent::ToolResult(tool_result) = content
+                        {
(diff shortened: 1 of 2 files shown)


gold-silver-copper:
## Self-review (adversarial)

Ran a multi-dimension adversarial review of the diff (behavior-equivalence vs the pre-refactor drivers, error paths, tool concurrency, span/telemetry fidelity, the `final_item` restructure, and lifetimes/Send/resume), with each candidate finding independently verified against the actual code + the `main` baseline, plus a dedicated deep pass on `drive_tool_calls`.

**Verified correct** (no change needed): the concurrent tool reassembly/index ordering, the lowest-index error tie-break + surfacing gate, the preresolved `Span::none()` path, internal-call-id propagation, and the absence of panics/unwraps/OOB on fallible paths. The serialize/resume keystone holds (`AgentRun` remains the sole owner of serializable state).

**Three confirmed divergences, all addressed** (`c1e1a4e7`):

1. **(medium) Multi-tool terminate at the default `tool_concurrency(1)`.** When an early tool's hook terminates a multi-tool turn, blocking `run()` no longer drains the remaining tools (the old `buffered(1)` + `collect` ran them all and fired their hooks first). This was a real, untested change — but it makes `run()`/`stream()` *actually* lock-step at concurrency 1, which `tool_concurrency`'s docs already promised and the pre-refactor code violated for exactly this case. Kept the unified behavior, **declared it** in the PR description, and added `default_concurrency_terminate_stops_remaining_tools_on_both_drivers` to pin it. (At `concurrency > 1`, in-flight siblings still drain.)

2. **(low) Run-completion log dropped** from the unified Done arm — restored as one shared "Agent run finished" event (unifying the old per-surface "Depth reached" / "multi-turn stream finished" markers).

3. **(nit) `agent_name` field dropped** from the empty-streamed-turn warning — restored (the source now carries the resolved agent name).

No blocker/high findings. 1052 rig-core lib tests + clippy + fmt green; Gemini `agent_run` cassettes re-verified.

gold-silver-copper:
## Tool-terminate semantics: researched → run-all-then-decide

Following up on the multi-tool-terminate behavior change, I researched how mature agent frameworks handle "a tool batch is short-circuited by a deliberate abort on one tool":

| Framework | Tool-batch exec | Sibling fate on a deliberate abort |
|---|---|---|
| semantic-kernel | unbounded `gather` | run-all (`terminate` checked *after* gather) |
| LangGraph | parallel `gather`/`executor.map` | run-all (siblings run; results discarded) |
| LangChain v1 | per-tool `Send` fan-out | run-all (reject is local to that call) |
| Vercel AI SDK | `Promise.all` | run-all (denied tool filtered pre-dispatch; siblings run) |
| pydantic-ai | per-segment parallel | run-all-then-decide (cancels only on a genuine exception) |
| openai-agents | bounded, `FIRST_COMPLETED` drain | short-circuit: skip-new + **cancel** in-flight (bespoke phase-aware machinery) |

**5 of 6 run all of the turn's tools, then honor the abort.** The lone short-circuiter (openai-agents) cancels in-flight work — which doesn't map to rig's non-cancellable tool futures (cancelling mid-execution risks partial side effects), and it carefully *drains* anything past the tool body anyway.

**Decision: run-all-then-decide for both surfaces** (`b2ea0a35`). This is the dominant pattern, it **restores `run()`'s pre-refactor `buffered(1)`+`collect`-all behavior** (so `run()` is now behavior-preserving), and it makes `run()`/`stream()` lock-step at *any* concurrency. Each tool stays independently hook-gated (a guardrail blocks any subset via per-call `Skip`/`Terminate`); a `Terminate` records the abort but doesn't pre-empt siblings' own hooks. Streaming's *surfaced items* are unchanged (post-error items suppressed as before); only a sibling's tool body + hooks now run — matching blocking. The only net change vs the previous push is that `stream()` no longer stops early at `tool_concurrency(1)`.

Verified: all three orderings ([A err,B ok] / [A ok,B err] / [A err,B err]) produce the exact original blocking result; streaming surfaced items byte-identical. 1052 lib tests + clippy + fmt green; cassettes re-verifying.

gold-silver-copper:
## Review pass 3

A third review pass (doc/claim consistency, accumulated round-2 + run-all interactions, test quality, completeness) — all findings **low/nit, no correctness bugs**. The unified-driver behavior, the "Agent run finished" log (`run.turn()` is valid at `Done`), and the `agent_name` field all verified clean; the "behave identically"/"lock-step" doc claims verified *accurate* (the run-all change made the two surfaces more aligned, not less).

Fixed (`de859c36`):
- **Stale `buffered` references.** The refactor removed the `buffered` combinator, but ~5 doc comments / test strings still named it. `run()`'s concurrent tool path now uses `buffer_unordered` + call-index-slot reindexing. Updated all sites (incl. the `tool_concurrency(0)` clamp rationale, which was doubly stale — `concurrency <= 1` routes to the sequential loop, so `buffer_unordered(0)` is unreachable).
- **Test strengthened.** `default_concurrency_terminate_runs_remaining_tools_on_both_drivers`'s streaming half asserted only the sibling call-count; it now also asserts the terminate error surfaces (with its reason), mirroring the blocking half — so a "runs siblings but swallows the terminate" regression fails.

Everything else the pass flagged was a false-positive (verified) or already declared in the PR body (the per-turn "depth" -> "Turns" log wording). Doc/test-only diff; 1052 lib tests + clippy + fmt green.

gold-silver-copper:
## Review pass 4 (performance/allocation + holistic)

No correctness bugs, but this pass found real efficiency regressions and — notably — a **test tautology** that had been hiding the outer-span guard. Fixed in `a1264d7f`:

**Perf regressions (blocking `run()` paid for work it discards):**
- The `Done` arm cloned the **full message history on every run**, even with no memory backend (the common path) — the baseline gated that clone inside the memory check. `append_run_messages` now borrows and clones only when appending.
- `drive_tool_calls` now takes `forward_items`: the blocking fold passes `false` to skip building the per-tool `ToolCall`/`ToolResult` items it discards (and the `tool_call`/result clones they need — potentially large for multimodal tool outputs); results move into the vec instead of cloning. Same "blocking opts out" pattern as the existing `final_item → None`. Tool execution + hooks are byte-identical (verified by the run/stream parity suite).

**Test tautology (the important one):** both outer-span guard tests (blocking + streaming) instrumented under a fieldless `info_span!("outer")`. Recording onto an *undeclared* field is a silent no-op invisible to the capture layer, so the "must not record onto the caller's span" assertions passed **regardless of the guard**. The outer spans now declare the fields, so the tests genuinely fail if the guard is removed (verified by bypassing it). This means the declared outer-span guard (change a) is now actually pinned on both surfaces.

**Consistency:** streaming now gates its per-turn `gen_ai.completion` on `created_agent_span` too (was ungated) — the guard my PR added to blocking now applies uniformly, so neither surface pollutes a caller/parent span that declares that field (nested-agent case).

**Doc:** the `TurnSource` trait comment listed invalid-tool resolution as living in `drive_agent`; it lives in each source's `run_model_turn`. Corrected.

The holistic pass otherwise confirmed convergence: fold/`filter_map` completeness, `StreamingError↔PromptError` round-trip, drop/borrow safety, and generic-over-both-sources correctness all check out. 1052 lib tests + clippy + fmt green; cassettes re-verifying.

PR #1990, opened 2026-07-01
fix(ollama): omit `think` when unset so the model default applies

### LINKED ISSUE #1970: bug(ollama): `think` is always serialized (defaults to `false`), so a model's default thinking can't be used
Hi, and thanks for rig! 🙏 I ran into this while building a local Ollama agent and wanted to report it clearly along with a suggestion. Happy to send a PR if the approach sounds good to you.

## Summary

The Ollama provider always serializes a `think` field and defaults it to `false` when the caller doesn't set one. There is no way to **omit** `think`. Because Ollama treats *absent* `think` as "use the model's default", rig can never express that neutral state — it always forces thinking on or off.

## Background: Ollama's `think` is tri-state

`/api/chat` distinguishes three states:

| `think` | behavior |
|---|---|
| **absent** | use the model's default (thinking-capable models think) |
| `false` | suppress thinking |
| `true` / `"low"`… | force thinking (errors if the model isn't flagged thinking-capable) |

The **absent** state is the safe, neutral choice — it respects whatever the model/server decided.

## Where it happens

`rig-core/src/providers/ollama.rs` (as of `rig-core` 0.39.0): `OllamaCompletionRequest.think` is a non-optional field with no `skip_serializing_if`, defaulted to `Think::Bool(false)`:

```rust
// field — always serialized
think: Think,

// in TryFrom<(&str, CompletionRequest)>
let mut think = Think::Bool(false);
```

So a built request always carries `think: false` or `think: true|level` — never absent. This is also asserted by the existing test `test_completion_request_with_think_false_default`, which expects the serialized request to contain `"think": false` when the caller passed nothing. Both the non-streaming and streaming paths go through this same `TryFrom`, and `OllamaCompletionRequest` is `pub(super)`, so downstream code can't override the serialization either.

## Why this is a problem

Defaulting to `false` is not neutral — it actively turns OFF thinking the model would do by default. This bites in a common real-world case:

Some models think by default but are **not** advertised as thinking-capable in `/api/show` `capabilities` (frequently true for GGUF models imported via `hf.co/...`). For such a model, Ollama behaves like this (verified on Ollama 0.30.11):

- `think` **absent** → returns a populated `thinking` field (it thinks),
- `think: false` → suppresses thinking,
- `think: true` → **errors**: `"<model> does not support thinking"`.

Because rig forces `false` and `true` is rejected, **there is no value a rig caller can pass to get this model's default reasoning output.** The one state that works (absent) is the one rig removed.

Note this is independent of Ollama's own capability-detection quirk: even against a perfectly-behaving Ollama, defaulting an optional upstream parameter to a concrete value (rather than omitting it) silently changes model behavior the caller never asked to change.

### Real-world example

I hit this in my own project; the workaround/notes live here for reference: kanghouchao/ExpertBase#7 (the in-repo code only passes `think:true` when the model reports the capability and otherwise omits it — but rig re-adds `false`, so it's a no-op without a fix here).

## Expected behavior

When the caller does not put `think` in `additional_params`, rig should omit the field entirely so Ollama applies the model's default. Explicit `think: true|false|level` should keep working exactly as today.

## Suggestion

Make `think` optional and skip serialization when unset (default `None`) — roughly:

```rust
think: Option<Think>,
// ...
let mut think: Option<Think> = None;
// ...
if let Some(think_val) = obj.remove("think") {
    think = Some(match think_val { /* unchanged */ });
}
```

and update `test_completion_request_with_think_false_default` to assert `think` is absent when unset. This is fully backward compatible for callers that pass an explicit value.

I'd be glad to open a PR with this if you're open to it. Thanks again!

## Related

- #961 (unable to disable thinking) — its fix appears to have introduced the unconditional `false` default; this report is the inverse (can't *omit* to get the model default).
- #505 (support ollama thinking).

## Environment

- rig-core 0.39.0
- ollama 0.30.11

SarthakB11:
# fix(ollama): omit `think` when unset so the model default applies

## Description

The Ollama provider always serialized a `think` field, defaulting it to `false` when the
caller did not set one. Ollama's `/api/chat` treats `think` as tri-state: an absent field
means "use the model's default", `false` suppresses thinking, and `true`/level forces it.
By always sending `false`, rig turned off thinking that a model would otherwise do by default,
and there was no way to express the neutral absent state.

This makes `OllamaCompletionRequest.think` optional and skips serialization when it is unset,
so a request built without an explicit `think` omits the field entirely. Explicit
`think: true | "low" | "medium" | "high"` requests are unaffected.

Fixes #1970

## Type of change

- [x] Bug fix

## Testing

- [x] `cargo test -p rig-core --lib providers::ollama::tests::`: 28 passed, 0 failed.
- [x] Verified the no-`think` case omits the field, and the explicit `true`/`low`/`medium`/`high` cases still serialize the expected top-level value.

## AI assistance

This PR was written with the assistance of AI tools (Claude Code).

## Notes

PR #1982 also edits `crates/rig-core/src/providers/ollama.rs` (adding a `max` think level). The scope here is disjoint (default/omission only), but a small rebase conflict near the `Think` handling is possible if both land; happy to rebase on whichever merges first.

### DIFF
--- a/crates/rig-core/src/providers/ollama.rs
+++ b/crates/rig-core/src/providers/ollama.rs
@@ -410,7 +410,8 @@ pub(super) struct OllamaCompletionRequest {
     #[serde(skip_serializing_if = "Vec::is_empty")]
     tools: Vec<ToolDefinition>,
     pub stream: bool,
-    think: Think,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    think: Option<Think>,
     #[serde(skip_serializing_if = "Option::is_none")]
     max_tokens: Option<u64>,
     #[serde(skip_serializing_if = "Option::is_none")]
@@ -450,7 +451,7 @@ impl TryFrom<(&str, CompletionRequest)> for OllamaCompletionRequest {
                 .collect::<Vec<_>>(),
         );
 
-        let mut think = Think::Bool(false);
+        let mut think: Option<Think> = None;
         let mut keep_alive: Option<String> = None;
 
         let options = if let Some(mut extra) = req.additional_params {
@@ -458,7 +459,7 @@ impl TryFrom<(&str, CompletionRequest)> for OllamaCompletionRequest {
             if let Some(obj) = extra.as_object_mut() {
                 // Extract `think` parameter
                 if let Some(think_val) = obj.remove("think") {
-                    think = match think_val {
+                    think = Some(match think_val {
                         Value::Bool(think) => Think::Bool(think),
                         Value::String(think) => Think::Level(match think.to_lowercase().as_str() {
                             "low" => Level::Low,
@@ -475,7 +476,7 @@ impl TryFrom<(&str, CompletionRequest)> for OllamaCompletionRequest {
                                 "`think` must be a 'low', 'medium', 'high', or bool".into(),
                             ));
                         }
-                    };
+                    });
                 }
 
                 // Extract `keep_alive` parameter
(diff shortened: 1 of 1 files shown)


gold-silver-copper:
AWESOME

PR #2025, opened 2026-07-07
feat(anthropic): support ANTHROPIC_BASE_URL in from_env()

### LINKED ISSUE #2024: feat(anthropic): support ANTHROPIC_BASE_URL in from_env()

## Feature Request

Support `ANTHROPIC_BASE_URL` environment variable in `Client::from_env()`.

### Motivation

Currently `Client::from_env()` only reads `ANTHROPIC_API_KEY`. When using third-party Anthropic-compatible APIs (e.g., GLM, MiniMax), users have to manually build the client with `Client::builder().base_url(...)`.

The OpenAI client already supports `OPENAI_BASE_URL` in `from_env()`. This would make the behavior consistent across providers.

### Proposal

Modify `from_env()` in `rig-core/src/providers/anthropic/client.rs` to read `ANTHROPIC_BASE_URL` from the environment and pass it to the builder.

```rust
fn from_env() -> Result<Self, Self::Error>
where
    Self: Sized,
{
    let key = std::env::var("ANTHROPIC_API_KEY")
        .expect("ANTHROPIC_API_KEY not set");
    let mut builder = Self::builder().api_key(&key);

    if let Ok(base_url) = std::env::var("ANTHROPIC_BASE_URL") {
        builder = builder.base_url(&base_url);
    }

    builder.build().map_err(Into::into)
}
```

No breaking changes. The function signature and error behavior remain exactly the same as the current implementation.

### Alternatives

Users can already do this manually with `Client::builder().base_url(...).build()`. This change just makes `from_env()` more convenient and consistent with OpenAI.

TropicalDog17:
## Summary
Closes #2024 
- Read `ANTHROPIC_BASE_URL` in `anthropic::Client::from_env()` and pass it to the builder when set
- Matches the existing `OPENAI_BASE_URL` behavior for OpenAI clients
- No breaking changes — when the env var is unset, the default Anthropic API URL is used

Fixes #2024

## Test plan

- [x] `cargo fmt`
- [x] `cargo clippy -p rig-core --all-targets --all-features`
- [x] `cargo test -p rig-core anthropic::client::tests`

### DIFF
--- a/crates/rig-core/src/providers/anthropic/client.rs
+++ b/crates/rig-core/src/providers/anthropic/client.rs
@@ -112,9 +112,16 @@ impl ProviderClient for Client {
     where
         Self: Sized,
     {
+        let base_url = crate::client::optional_env_var("ANTHROPIC_BASE_URL")?;
         let key = crate::client::required_env_var("ANTHROPIC_API_KEY")?;
 
-        Self::builder().api_key(key).build().map_err(Into::into)
+        let mut builder = Self::builder().api_key(key);
+
+        if let Some(base) = base_url {
+            builder = builder.base_url(&base);
+        }
+
+        builder.build().map_err(Into::into)
     }
 
     fn from_val(input: Self::Input) -> Result<Self, Self::Error>


gold-silver-copper:
Merged as #2051, with your contributions preserved

PR #2027, opened 2026-07-07
feat(vertexai): round-trip Gemini thoughtSignature in tool-calling loop

### LINKED ISSUE #2026: rig-vertexai drops Gemini thoughtSignature in tool loop → INVALID_ARGUMENT on multi-tool requests
### Summary

`rig-vertexai` drops Gemini's opaque **`thoughtSignature`** in the tool-calling loop. Gemini 2.5/3.x "thinking" models attach a `thoughtSignature` to `functionCall` parts, and it **must be echoed back verbatim** on every subsequent turn of a multi-tool conversation. Because `rig-vertexai` neither reads it from the response nor writes it back into the request `contents`, the second turn of any multi-tool request is rejected by Vertex with:

```
INVALID_ARGUMENT — Function call is missing a thought_signature in functionCall parts.
This is required for tools to work correctly.
https://ai.google.dev/gemini-api/docs/thought-signatures
```

### Where the gap is

Both sides of the round-trip are missing in `crates/rig-vertexai/src/types/`:

- **Read side** — `completion_response.rs`: the conversion from `VertexGenerateContentOutput` to `CompletionResponse` never reads `part.thought_signature`, so it is lost when building `ToolCall`.
- **Write side** — `message.rs`: the conversion from `RigMessage` to `vertexai::model::Content` never calls `Part::set_thought_signature(...)`, so it is never sent back.

The carrier fields already exist end-to-end: `rig_core::message::ToolCall.signature: Option<String>` and the Vertex SDK's `Part.thought_signature` (protobuf `Bytes`). Only the wiring is missing.

### Prior art

The **direct Gemini provider** in `rig-core` already round-trips this via `ToolCall.signature` / `Reasoning::new_with_signature` (see `crates/rig-core/src/providers/gemini/completion.rs`), landed in #1396 ("cross-provider reasoning trace roundtrip"). Vertex AI was simply never in scope for that refactor — it uses a different SDK (`google-cloud-aiplatform-v1`, protobuf `Bytes`) rather than the direct provider's JSON `String`, so it needs its own small wiring.

No existing issue/PR covers this specific Vertex gap (searched `thought_signature` / `thoughtSignature vertex`). #1185 ("Add Gemini 3.0 streaming support for Vertex AI") touches `thoughtSignature` but only for a new streaming module — adjacent, not a duplicate.

### Proposed fix

Mirror the direct-provider pattern, scoped to the two Vertex files:

- Read side: carry `part.thought_signature` into `ToolCall.signature` (base64-encoded, since it is raw bytes).
- Write side: base64-decode `ToolCall.signature` back into `Part::set_thought_signature(...)`.
- Robustness: a malformed signature is dropped with a warning rather than failing the whole turn; a signature-bearing part that is neither a function call nor text is logged before being dropped.

PR opening shortly.

### Environment

- `rig-vertexai` 0.38.2 / 0.39.0 / current `main` (gap present in all)
- Provider: Vertex AI, `gemini-2.5-flash` / `gemini-3.x` thinking models

pdesainteagathe:
Closes #2026.

## Problem

Gemini 2.5/3.x "thinking" models attach an opaque **`thoughtSignature`** to `functionCall` parts, which must be echoed back verbatim on every subsequent turn of a multi-tool conversation. `rig-vertexai` drops it on **both** sides of the tool loop, so turn 2+ of any multi-tool request is rejected by Vertex with:

```
INVALID_ARGUMENT — Function call is missing a thought_signature in functionCall parts.
```

- Read side (`completion_response.rs`) never read `part.thought_signature`.
- Write side (`message.rs`) never called `Part::set_thought_signature(...)`.

## Fix

Mirrors the pattern already present in the direct Gemini provider (`rig-core/src/providers/gemini/completion.rs`, from #1396), scoped to the two Vertex files. The carrier fields already exist end-to-end (`ToolCall.signature: Option<String>` and `Part.thought_signature: Bytes`), so this is only wiring:

- **`completion_response.rs`** — carry `part.thought_signature` into `ToolCall.signature` (base64, since it is raw bytes).
- **`message.rs`** — base64-decode `ToolCall.signature` back into `set_thought_signature(...)`.
- **`Cargo.toml`** — add `base64` (workspace dependency).

### Robustness

- A malformed signature is dropped **with a warning** rather than failing the whole `Content` conversion — one bad byte must not kill every other tool call in the same turn.
- A signature-bearing part that is neither a function call nor text (a standalone "thinking" part with no rig-core carrier) is logged before being dropped, so a later `INVALID_ARGUMENT` can be traced back rather than being silent.

## Tests

Added 4 unit tests (all green):
- read side captures `thought_signature` into `ToolCall.signature` (base64)
- read side leaves `signature` `None` when absent
- write side echoes the signature back into the `Part`
- write side drops a malformed signature without failing the conversion

`cargo test -p rig-vertexai` and `cargo clippy -p rig-vertexai --all-targets` pass.

### DIFF
--- a/crates/rig-vertexai/Cargo.toml
+++ b/crates/rig-vertexai/Cargo.toml
@@ -10,6 +10,7 @@ description = "VertexAI model provider for Rig integration."
 workspace = true
 
 [dependencies]
+base64 = { workspace = true }
 google-cloud-aiplatform-v1 = { workspace = true }
 google-cloud-auth = { workspace = true }
 rig-core = { path = "../rig-core", version = "0.39.0", default-features = false }
--- a/crates/rig-vertexai/src/types/completion_response.rs
+++ b/crates/rig-vertexai/src/types/completion_response.rs
@@ -1,3 +1,5 @@
+use base64::Engine as _;
+use base64::engine::general_purpose::STANDARD as BASE64;
 use google_cloud_aiplatform_v1 as vertexai;
 use rig_core::OneOrMany;
 use rig_core::completion::{CompletionError, CompletionResponse, Usage};
@@ -27,6 +29,13 @@ impl TryFrom<VertexGenerateContentOutput> for CompletionResponse<VertexGenerateC
         // vertexai internally uses a wkt::Struct (serde_json::Map<String, serde_json::Value>) in
         // function calling args. We need to convert that to serde_json::Value for rig_core::completion type matching
         for part in content.parts.iter() {
+            // Gemini "thinking" models attach an opaque `thoughtSignature` to (usually) the
+            // functionCall part. It must be echoed back verbatim on subsequent turns or Vertex
+            // rejects the request with INVALID_ARGUMENT ("missing a thought_signature"). We carry
+            // it through rig-core's `ToolCall.signature` (base64, since it is raw bytes).
+            let signature = (!part.thought_signature.is_empty())
+                .then(|| BASE64.encode(&part.thought_signature));
+
             if let Some(function_call) = part.function_call() {
                 let args_json = function_call
                     .args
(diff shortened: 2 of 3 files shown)


gold-silver-copper:
This has been merged as stacked PR #2052, thank you!

PR #2040, opened 2026-07-08
refactor(providers)!: collapse remaining providers onto GenericCompletionModel<Ext> (#2035 phases 2–4)

### LINKED ISSUE #2035: refactor(providers): collapse 13 hand-rolled OpenAI-compatible providers onto GenericCompletionModel<Ext>
# Summary

rig-core has two shared-reuse mechanisms in the provider layer, at very different levels of adoption:

1. **Streaming** is centralized in `providers/internal/openai_chat_completions_compatible.rs` (`send_compatible_streaming_request` + `CompatibleStreamProfile`). **15 providers already delegate to it.** This layer is done — nothing to do here.
2. **The completion-model layer one level up** — request building, message conversion, tool plumbing, error handling — is centralized in `GenericCompletionModel<Ext, H>` (`providers/openai/completion/mod.rs:1141`) plus the `Provider`/`Capabilities`/`Capable` machinery (`client/mod.rs:241-289`). **Only 3 providers have adopted it**: `minimax`, `zai`, and `xiaomimimo`.

The result is that **13 OpenAI-chat-compatible providers still hand-roll their own `struct CompletionModel`, request/response types, and `TryFrom<message::Message>` conversions**, re-deriving what the generic model already does. Several of them (deepseek, hyperbolic, galadriel, mira) clone the OpenAI wire types wholesale. Rough estimate: **~7,000 lines of near-duplicate code collapse to ~150–350-line Ext + client shells.**

All file/line references below are against `4c25682e`.

# The template already works

The three migrated providers prove the generic path is complete. Each is a ~350–400-line file (`minimax.rs`: 358, `zai.rs`: 357, `xiaomimimo.rs`: 406) that supplies **nothing but an `Ext` marker and a client builder** and gets completion, streaming, **and tools** for free — zero tool-plumbing lines of their own:

```rust
// minimax.rs:94
type Completion = Capable<super::openai::completion::GenericCompletionModel<MiniMaxExt, H>>;
// minimax.rs:107 — same pattern works for Anthropic-shaped APIs too
type Completion = Capable<super::anthropic::completion::GenericCompletionModel<MiniMaxAnthropicExt, H>>;
```

The `Ext` contract is tiny (`client/mod.rs:241-266`): a `VERIFY_PATH` const, an optional `build_uri` override, and an optional `with_custom` request hook. Notably, the **default** `build_uri` already handles the blank-base-URL case that exists for Azure's sake.

The generic request path (`openai/completion/mod.rs:1213-1352`) already covers the quirks providers hand-rolled workarounds for — e.g. the "don't send `response_format` alongside tools on the first turn" llama.cpp quirk is handled generically at `openai/completion/mod.rs:~1291`, yet `llamafile.rs:291` still carries its own copy.

# Current state

**Fully migrated (3):** minimax, zai, xiaomimimo.

**Half-migrated (1):** `moonshot` — its Anthropic-endpoint side already uses `GenericCompletionModel<MoonshotAnthropicExt>` (`moonshot.rs:140-142`), but its OpenAI side still wires a hand-rolled `struct CompletionModel` (`moonshot.rs:471`) into `Capabilities<H> for MoonshotExt` (`moonshot.rs:128-129`) — even though `MoonshotExt` already exists (`moonshot.rs:56`) and implements `Provider`. One provider, both patterns.

**Hand-rolled but OpenAI-chat-compatible (13):**

| provider | file | lines | notes |
|---|---|---|---|
| llamafile | `providers/llamafile.rs` | 862 | already uses `openai::ToolDefinition`; its one quirk is handled generically |
| moonshot (OpenAI side) | `providers/moonshot.rs` | 805 | `MoonshotExt` already exists; uses `openai::Message`/`openai::CompletionResponse` |
| together | `providers/together/completion.rs` | 440 | small; streaming already delegates to OpenAI profile |
| perplexity | `providers/perplexity.rs` | 565 | no tool calling (upstream limitation); reversed conversion direction (`:279,330`) |
| galadriel | `providers/galadriel.rs` | 685 | clones OpenAI types (`:225,286,294,419`) |
| mira | `providers/mira.rs` | 835 | ad-hoc `impl From<Message> for serde_json::Value` (`:587`), multi-shape `enum CompletionResponse` (`:128`) |
| hyperbolic | `providers/hyperbolic.rs` | 957 | clones OpenAI types (`:168,250`) |
| groq | `providers/groq.rs` | 1056 | custom `CompatibleStreamProfile` already (`:770`) |
| azure | `providers/azure.rs` | 1377 | uses `openai::Message` + `type Response = openai::CompletionResponse` (`:570,675,686`) |
| deepseek | `providers/deepseek.rs` | 1451 | clones OpenAI types; `reasoning_content` + prefix-message fields (`:201`) |
| huggingface | `providers/huggingface/completion.rs` | 1338 | streaming already delegates (`huggingface/streaming.rs:74`) |
| mistral | `providers/mistral/completion.rs` | 1317 | prefix assistant messages, different `tool_choice` shape |
| openrouter | `providers/openrouter/completion.rs` | 4248 (~2046 non-test) | special case, see below |

(Not in scope: gemini, anthropic, cohere, ollama, copilot, xai — genuinely different wire formats or already on a different shared path.)

# Per-provider quirks (what the Ext hooks must absorb)

| provider | quirk | mechanism |
|---|---|---|
| azure | blank base_url + `api-version` query param + deployment-path routing | `Provider::build_uri` / `with_custom` — both exist for exactly this |
| deepseek | `reasoning_content` response field; beta prefix-message request fields | extend shared types with optional fields, or Ext-level request hook |
| mistral | prefix assistant messages; different `tool_choice` shape | same |
| perplexity | citations; no tool calling | optional response fields; `Capabilities` already models missing capabilities via `Nothing` |
| mira | multiple upstream response shapes | untagged enum on the response side or normalize in `with_custom` |
| llamafile | `response_format`-after-tools ordering | **already handled generically** (`openai/completion/mod.rs:~1291`) — delete the local copy |
| openrouter | `ProviderPreferences` routing config (~530 lines, `:46-576`) + `reasoning_details` + video/file content | preferences are genuinely unique — keep; see below |

**openrouter** deserves its own treatment: ~2,200 of its 4,248 lines are tests, ~530 are legitimately unique routing config, but **~900 lines (`:784-1710`) are a forked parallel message model** (`Message`, `UserContent`, `ImageUrl`, `FileContent`, `VideoUrlContent`) plus bidirectional `TryFrom` converters to/from the OpenAI types — forked only to add `reasoning_details` (`:1398`) and extra content variants. Extending the shared OpenAI message types with those optional fields would cut the fork to ~300 lines. This can be a follow-up sub-task rather than a blocker for the rest.

# Proposed sequencing

**Phase 1 — prove the pattern (smallest blockers):**
1. `llamafile` — near-zero blockers; mostly deletion.
2. `moonshot` — finish the half-done migration; `MoonshotExt` already implements `Provider`.
3. `together` — smallest hand-rolled file.

**Phase 2 — straightforward but bigger:**
4. `azure` (build_uri/with_custom hooks), `groq`, `huggingface`, `perplexity`.

**Phase 3 — the type-cloners (need optional-field extensions to shared types first):**
5. `deepseek`, `hyperbolic`, `galadriel`, `mira`, `mistral`.

**Phase 4 (follow-up):** de-fork openrouter's message model onto extended shared types.

Each provider should be its own PR — behavior-preserving, reviewable against its cassette/unit suite.

# Testing & safety

- Cassette regression suites already exist for **deepseek, groq, mistral, openrouter** (`tests/cassettes/*`, added in #2006–#2011) — these guard the riskiest migrations.
- **llamafile, moonshot, together, azure, hyperbolic, galadriel, mira, perplexity, huggingface have no cassette coverage.** Recording a minimal cassette suite (completion + streaming + one tool round-trip) *before* migrating each one should be part of the definition of done — otherwise we're refactoring wire-format code blind.
- The per-provider serialization unit tests that live in the hand-rolled modules should be preserved/ported where they cover provider-specific wire shapes.

# Side benefits that fall out automatically

- **Error handling unification**: the hand-rolled providers use the legacy `enum ApiResponse<T> { Ok, Err }` flatten in their non-streaming paths, which discards the HTTP status code; the generic model uses the structured `from_http_response` path that preserves status + body (same machinery the shared streaming module already uses). Migrating fixes the streaming/non-streaming error asymmetry per provider for free.
- **Message-conversion consistency**: the ad-hoc outliers (mira's `From<Message> for serde_json::Value`, perplexity's reversed conversion direction, galadriel's split `TryFrom` pair) all disappear in favor of the one shared `TryFrom<message::Message>` (`openai/completion/mod.rs:690`).

# Acceptance criteria

- [ ] Each listed provider's completion path goes through `GenericCompletionModel<Ext>` (or a documented decision is recorded for why it can't).
- [ ] No provider in the list retains its own `struct CompletionModel`, request-param struct, or `TryFrom<message::Message>` impl duplicating the shared one.
- [ ] Cassette coverage exists for each provider before its migration PR lands.
- [ ] Non-streaming error paths preserve HTTP status via the structured error path (no `ApiResponse<T>` flatten remaining in migrated providers).
- [ ] `CONTRIBUTING.md` / `providers/mod.rs` checklist (#1997) updated to state that new OpenAI-compatible providers MUST use `GenericCompletionModel<Ext>` (the minimax/zai/xiaomimimo template).

# Out of scope

- The shared streaming module (`internal/openai_chat_completions_compatible.rs`) and the `ProviderClient` dyn registration layer — both already consistent and well-tested.
- Non-OpenAI-shaped providers (gemini, anthropic, cohere, ollama, copilot, xai/responses).

gold-silver-copper:
Completes #2035 in one pass: phases 2 and 3 (all 11 remaining hand-rolled OpenAI-compatible providers, including the two phase-1 leftovers) plus phase 4 (the OpenRouter message-model de-fork).

**Net: −5,661 lines** (48 files, +2,022/−7,683).

### Every provider on the list now rides `GenericCompletionModel<Ext>`

**groq, deepseek, mistral, together, moonshot (OpenAI side), perplexity, hyperbolic, galadriel, mira, azure, huggingface** lose their hand-rolled `CompletionModel` structs, request-param structs, `TryFrom<message::Message>` conversions, and per-provider streaming profiles. Each becomes an Ext + client shell; `CompletionModel` in each module is a type alias for the generic model. Non-completion capabilities (groq transcription, deepseek model listing, mistral/together embeddings, azure embeddings/transcription/audio, hyperbolic image/audio, huggingface transcription/image, mira `list_models`) stay as they were.

### The shared layer grew exactly the hooks the quirk table in the issue called for

All on `OpenAICompatibleProvider` (the trait added in #2038):

- **`type Response`** — providers keep provider-specific response payloads: deepseek's cache-token usage, mistral's cached-token fallback logic, mira's untagged `Structured | Simple(String)` shape. (Acceptance criteria ban hand-rolled *request* machinery; response structs are data, not duplication.)
- **`completion_path(&self, model)`** — expresses Azure's `{endpoint}/openai/deployments/{model}/chat/completions?api-version=…` routing **behavior-preservingly** (the issue guessed `build_uri` would do; it can't — it never sees the model), and lets mistral/together/hyperbolic/mira/huggingface reach their `/v1` completion routes without moving `BASE_URL` out from under their sibling endpoints.
- **`prepare_request`** (typed, pre-serialization) — groq's `compound_custom.enabled_tools` native-tool folding, moonshot's `required`→`auto`+steering-message coercion, huggingface's Fireworks model-id mapping, perplexity/mira tool stripping.
- **`finalize_request_body`** (serialized body, post-stream-merge) — deepseek's string-flattened content + `tool_calls[].index` + thinking-gated tool-choice suppression, mistral's `"any"` tool choice + `prefix` field + reasoning stripping, mira's raw-message flattening.
- **`SUPPORTS_RESPONSE_FORMAT`** / **`STREAM_INCLUDE_USAGE`** consts for providers that reject `json_schema` response formats or `stream_options`.

Plus shared-path fixes that fell out: `openai::ToolChoice` gains the `Function` variant OpenAI's real API supports (so `ToolChoice::Specific` stops erroring everywhere), the shared `ToolResult` conversion prefers `call_id` over `id` (latent bug — Mistral's live API rejected the old behavior), the streaming delta accepts Groq's `reasoning` alias, `function_call` finish reasons map to tool calls, and 2xx error envelopes of any JSON shape now preserve the raw body through `provider_response_*`.

### Phase 4: OpenRouter de-fork

The ~930-line parallel message model (`Message`, `UserContent`, `ImageUrl`, `FileContent`, `VideoUrlContent`, 5 bidirectional converters) is deleted. Shared openai types gain OpenRouter's skip-when-absent extensions: `UserContent::Video`, `ImageUrl.detail: Option<ImageDetail>` (OpenAI still emits `"detail":"auto"` byte-identically), assistant `reasoning_details`/`images` fields, a `role:"model"` deserialize alias, and the audio part tag corrected to `input_audio`. `openrouter::{Message, UserContent, …}` are re-exports. OpenRouter keeps its request/response/streaming layer (`ProviderPreferences`, cost accounting, reasoning-details grouping, generated-image extraction) plus a small `messages_from_rig_message` conversion for reasoning-details replay — the documented exception the issue anticipated.

## Verification

- **All 13 cassette suites replay green — 872 tests, 0 failures**: openai 110, chatgpt 53, copilot 58, anthropic 96, gemini 143, ollama 47, xai 64, **openrouter 78 (zero cassette edits — the de-fork is wire-invisible)**, deepseek 67, groq 41, mistral 40, mistralrs 39, llamafile 36.
- **Cassette-guarded migrations validated against live APIs**: the groq suite was re-recorded live against the new wire (the only diff was groq's vestigial `"stream":false`); 4 deepseek + 1 mistral scenarios re-recorded for *intended* fixes (`max_tokens` now forwarded; `call_id` in tool results — Mistral's API rejected the old wire, proving the fix). Every other deepseek/mistral cassette replays **byte-exact** against its original recording.
- `cargo test -p rig-core --lib`: 1,157 passed. Doc tests: 86 passed. `cargo check --workspace --all-targets`, clippy (default and `image,audio` features), and fmt clean.

## Breaking changes

- Custom out-of-tree `OpenAICompatibleProvider` impls must add `type Response = openai::CompletionResponse;` (other items are defaulted).
- Deleted per-provider public request/message/tool-choice types (e.g. `deepseek::ToolChoice`, mistral/galadriel/hyperbolic/perplexity request+response forks, groq's `GroqToolChoice`); per-provider `StreamingCompletionResponse` types are aliases of the shared one.
- Behavior notes (full list in CHANGELOG): `max_tokens` now forwarded by deepseek/together/hyperbolic/galadriel/azure; together streams via standard `stream`/`stream_options`; perplexity's non-streaming endpoint drops its stray `/v1`; galadriel response tool-call ids use the real id; mira's preamble is a `system` message; `output_schema`→`response_format` is deferred while tools are pending a result.

## Acceptance criteria from #2035

- [x] Each listed provider's completion path goes through `GenericCompletionModel<Ext>` (OpenRouter's response/streaming layer documented as the exception, per the issue's own phase-4 scoping).
- [x] No provider retains its own `struct CompletionModel`, request-param struct, or duplicating `TryFrom<message::Message>`.
- [x] Cassette coverage guards the riskiest migrations (deepseek/groq/mistral/openrouter suites all green; llamafile from #2038). azure/together/moonshot/perplexity/hyperbolic/galadriel/mira/huggingface have no recordable local endpoint or key available — their live `#[ignore]` suites compile against the new path and their wire-shape unit tests were ported.
- [x] Non-streaming error paths preserve HTTP status/body via the structured path — no `ApiResponse<T>` flatten remains in any migrated completion path (the per-provider error-envelope regression tests were ported and pass).
- [x] `CONTRIBUTING.md` + `providers/mod.rs` checklist now state new OpenAI-compatible providers MUST use `GenericCompletionModel<Ext>`.

Closes #2035.


gold-silver-copper:
Addressed the review findings in a01cc2e6 — verification of each:

1. **Hyperbolic (Blocker) — confirmed true, fixed.** The old `TryFrom` warned and dropped `output_schema`, `tools`, and `tool_choice`; the migrated path forwarded all three. Restored: `SUPPORTS_RESPONSE_FORMAT = false` plus a `prepare_request` that clears tools/tool_choice with the original warnings (unit-tested). `max_tokens` forwarding is kept as the documented improvement — it's universally supported.

2. **Galadriel (Blocker) — confirmed true, fixed.** Old code warned and dropped `output_schema` (it always sent tools). Added `SUPPORTS_RESPONSE_FORMAT = false`.

3. **Perplexity (High) — confirmed true, fixed.** Old code sent every message as `{role, content: String}`. Added a `finalize_request_body` that flattens text-only content-part arrays back to newline-joined strings (unit-tested); arrays containing non-text parts pass through for sonar's documented multimodal handling rather than erroring as before.

4. **Llamafile (High) — partially true, fixed + verified empirically.** The content-array change was the documented intent of #2038, but two sub-points were real: (a) `output_schema` → old code warned and dropped, the generic path sent `response_format` — restored with `SUPPORTS_RESPONSE_FORMAT = false` (older llamafile builds predate `json_schema` support); (b) fair point that Ollama's OpenAI layer is its own Go implementation, not llama.cpp's server. I installed `llama.cpp` (`brew`), started a real `llama-server --jinja` on a local GGUF, and **re-recorded the four chat cassettes against it** — completion, streaming, non-streaming tool round-trip, and streaming tool round-trip all pass against genuine llama.cpp, including content-part arrays and `tool_calls`/`tool_result` echoes. Cassette provenance is documented in `cassette_support.rs`.

5. **Mistral streaming usage (Medium) — confirmed true, fixed structurally.** `StreamingCompletionResponse` is now generic over the provider usage payload (`StreamingCompletionResponse<U = openai::Usage>`), selected via a new `OpenAICompatibleProvider::StreamingUsage` associated type. Mistral streams parse into `mistral::Usage` again (`num_cached_tokens` fallback + `prompt_token_details` alias), and DeepSeek gets its cache hit/miss counters back too. Both suites' streaming cassettes replay green against the restored types.

Full battery after the fixes: all 13 cassette suites green (872 tests), 1,159 lib tests, workspace check / clippy / fmt clean.

gold-silver-copper:
Ran a fresh multi-angle self-review of this branch (8 finder passes: line-by-line, removed-behavior audit, cross-file trace, reuse/simplification/efficiency/altitude/conventions) and pushed fixes as 92ec0917.

**Correctness fixes:**
1. **Tool-id asymmetry (shared)** — the tool-result side preferred `call_id` (fixed earlier on this branch) but the assistant tool-call echo still serialized plain `id`; Responses-API history replayed via chat completions would send mismatched ids. Both sides now use `call_id.unwrap_or(id)`.
2. **Llamafile structured outputs** — the `SUPPORTS_RESPONSE_FORMAT = false` I added in the previous review round was an over-correction: current main (post-#2038) already mapped `output_schema` → `response_format` for llamafile, and llama.cpp supports `json_schema`. Reverted.
3. **Mira/Perplexity streaming** — both now set `STREAM_INCLUDE_USAGE = false`; their pre-migration requests never carried `stream_options` and Mira's gateway is strict about unknown params.
4. **Mistral reasoning-model streaming** — the shared delta parser only accepted string `content`; Mistral's magistral models stream content-part arrays, which were dropping whole chunks. The shared deserializer now flattens text parts (restores the old Mistral tolerance).
5. **Moonshot forced tool choice** — the new `ToolChoice::Function` variant sailed past the `Required` coercion to an API that only accepts auto/none; it's rejected client-side again with a descriptive error (pre-migration behavior).
6. **OpenRouter reasoning key** — plain assistant reasoning now hits the wire as `reasoning` (OpenRouter's documented field) instead of the shared `reasoning_content`, renamed at the `final_request_body` choke point.
7. **Azure telemetry** — `gen_ai.provider.name` restored to `azure.openai`.
8. **Mira usage overflow** — `total - prompt` on usize now saturates (a gateway reporting total < prompt would panic in debug builds).
9. **Perplexity history hygiene** — role:`tool` messages and assistant `tool_calls`/`reasoning_content` remnants are stripped from shared histories (mirroring the Mira hook).

**Cleanup:** extracted shared `strip_unsupported_tools` + `flatten_text_content_parts` helpers (the 4 hand-rolled JSON flatteners and 3 tool-strippers now share them, exact per-provider separators preserved); deleted dead `galadriel::Usage` / `hyperbolic::Usage`+`EmbeddingData`; dropped a per-stream model clone.

**Reviewed and deliberately not changed** (documented behavior/design): the response_format tools-pending deferral now applying to groq/mistral/azure; groq's `x_groq.usage` fallback removal; the `Video`/`reasoning_details` extensions on shared openai types (explicitly sanctioned by #2035 phase 4); the double-serialization on no-op finalize hooks (a skip-const would be a silent-drift footgun); the shared conversion dropping user text mixed with tool results in one message (pre-existing shared-path behavior, follow-up candidate).

Post-fix: all 13 cassette suites green (872 tests, no cassette edits), 1,161 lib tests, workspace check/clippy/fmt clean under default and `image,audio` features.

gold-silver-copper:
Second self-review round (8 fresh finder angles focused on the previous round's fixes and thinly-covered areas) — fixes pushed as ca83a99d.

**Bugs found and fixed — notably three in my own round-one fixes:**
1. **CI-breaking doc link** — `[OpenAICompatibleProvider::completion_path]` in azure.rs doesn't resolve (type not in scope); the doc CI job runs with `-D warnings` and would have hard-failed. Fixed with a fully-qualified link.
2. **Round-one mistral fix silently no-opped** — the scripted swap of Mistral's inline content-flattener for the shared helper missed its anchor after rustfmt reflowed the code, leaving the duplicate the extraction was supposed to remove. Now actually uses the helper (and this time verified).
3. **Round-one openrouter id asymmetry** — I fixed the shared assistant tool-call echo to prefer `call_id`, but OpenRouter's own tool-result conversion still sent the plain id, recreating the mismatch inside OpenRouter. Both sides now agree.
4. **Round-one perplexity strip was incomplete** — removing `tool_calls` left content-less `{"role":"assistant"}` messages, and removing `tool` messages left consecutive assistant turns; Perplexity rejects both (strict alternation). Replaced with a shared `sanitize_plain_text_history` helper (also adopted by mira and hyperbolic) that drops tool-call-only assistant turns and merges consecutive assistant messages — unit- and per-provider-tested.
5. **Azure deployment URL followed per-request model overrides** — old code pinned the URL to the configured deployment and only overrode the body model; the generic path was routing the override into `/openai/deployments/{model}/` (404 for non-deployment names). `completion_path` now receives the handle's configured model, restoring the old semantics.
6. **Streaming spans lost `gen_ai.system_instructions`** — every pre-migration provider recorded it on streaming spans and the shared non-streaming span still does; restored on the shared streaming span.
7. **Cassette privacy leak** — the llamafile cassettes I recorded against `llama-server` embedded my local GGUF blob path (with username) as the response model string. Re-recorded with `--alias`; all four cassettes are clean.

**Coverage restored** (flagged by the test-integrity audit): groq reasoning-param flattening, deepseek per-role join separators, and the new sanitization behavior all have unit tests again.

**Cleanup:** deleted OpenRouter's dead pre-migration `ToolChoice` trio and Mistral's dead `Message` constructors; the streaming delta parser and Mistral's response parser now share one text-part joiner; `flatten_text_content_parts` builds its string without intermediate Vecs; documented the responses-API `ToolChoice::Mode` interaction with the new `Function` variant.

**Reviewed, deliberately unchanged** (CHANGELOG notes added): moonshot no longer preserves reasoning-only assistant history turns (shared conversion drops empty turns — emitting them would risk 400s on stricter providers); empty assistant content now errors for hyperbolic/perplexity/huggingface (aligns with the shared path's long-standing semantics); together's `stream_options` modernization remains unverified without a key; openrouter multi-name `Specific` tool choice now errors client-side (the old fork serialized a non-standard array).

Post-fix: all 13 cassette suites green (872 tests), 1,166 lib tests, workspace check / clippy / fmt / **doc** all clean.

gold-silver-copper:
Third self-review round — this pass aimed the finder angles at the round-two fixes themselves, the build matrix, cassette internals, telemetry parity, and docs-vs-code accuracy. Fixes pushed as cc6565f9.

**Bugs found & fixed (several in the previous round's own fixes):**
1. **Sanitize helper only half-fixed alternation** — it merged consecutive *assistant* turns but its own removals also create *user/user* adjacency (tool exchange with no surviving assistant turn), empirically probed to yield `["user","user"]` — still a Perplexity 400. The merge now covers consecutive same-role turns, and `refusal` parts count as text when flattening (a refusal-bearing assistant array previously escaped both flatten and merge).
2. **Stale release notes from my own round-two revert** — the CHANGELOG still claimed llamafile drops `output_schema`, contradicting the code after the revert; also fixed the `images` "skip-when-empty" wording (it's inbound-only) and a reference to the deleted `together::ToolChoice`.
3. **Streaming telemetry recorded fiction** — `gen_ai.input.messages` was serialized *before* the finalize hooks, so spans showed tool-role messages and content arrays the wire never carried. Now captured post-finalize.
4. **Shallow stream-params merge clobbered caller `stream_options`** (streaming only; non-streaming forwarded them) — now respected when present.
5. **OpenRouter encrypted-reasoning id correlation** — the recovery path only fell back to `call_id`, dropping or null-ing the Encrypted detail when the provider id lived in `id`; it now correlates with the actual wire tool-call id.
6. **OpenRouter non-streaming usage underflow** — `total - prompt` on usize replaced with the reported `completion_tokens` (matching its own streaming path).
7. **Video data-URI round-trip was lossy-then-fatal** — data URIs with MIMEs outside the `VideoMediaType` set (e.g. `video/quicktime`) decomposed into Base64 videos that error on re-serialization; they now stay as URLs and round-trip byte-identically.
8. **Telemetry unification documented + completed** — the `chat_streaming`→`chat` span change and new `gen_ai.input.messages` recording for migrated providers is now in the CHANGELOG (it's also a PII-surface note); openrouter's streaming span joins the fleet convention and gains the TRACE wire log its non-streaming path had. `gen_ai.request.model` now reports per-request overrides.

**API-surface hygiene:** documented all previously-omitted public removals (raw response types, together/moonshot ToolChoice enums, groq/deepseek `send_compatible_streaming_request`, openrouter content-builder helpers, the multi-name `Specific` arity error); added `CompletionResponse` compat aliases to perplexity/hyperbolic/galadriel/huggingface; un-hid `GenericCompletionModel` (it's the advertised extension point) and documented the hook execution order on `OpenAICompatibleProvider` for out-of-tree implementors.

**Verified clean this round:** full feature matrix (`--all-features`, no-default-features, doc job with `-D warnings`), all re-recorded cassettes re-audited for secrets/org-ids/local paths (none), scenario-registry checks, and replay of all 13 suites.

**Skipped with rationale:** the shared `TryFrom<Message> → rig message` conversion drops `reasoning_details` (openrouter's own response path preserves them; mapping them in the shared conversion would change all providers — follow-up), and openrouter's reasoning-details ordering/index-synthesis quirks (pre-existing, low-confidence, untouched by this PR).

Post-fix: 13/13 cassette suites green (872 tests), 1,168 lib tests, workspace check / clippy (all features) / fmt / docs clean.

gold-silver-copper:
Fourth self-review round — finders aimed at the round-three fixes, the cumulative final state, agent-loop integration, and a live llama.cpp exercise pass. Fixes pushed as 1f4b0f62.

**The most severe find of all four rounds, caught by tracing the agent loop rather than the diff:**
1. **Agents silently lost structured output on 8 providers.** `GenericCompletionModel` hardcodes `composes_native_output_with_tools() = true`, but the pre-migration hand-rolled models defaulted to `false` — so agents with tools + `output_schema` used tool-mode schema enforcement. After the migration, `SUPPORTS_RESPONSE_FORMAT=false` providers (deepseek, together, moonshot, huggingface, galadriel, hyperbolic, perplexity, mira) resolved output mode to Native and then warn-dropped the schema: **no output tool, no response_format, no enforcement at all**. The flag now follows `SUPPORTS_RESPONSE_FORMAT` — schema-less providers fall back to tool mode as before; groq/mistral/azure compose natively like openai.

**Three round-three fixes needed correcting (the pattern continues):**
2. The `stream_options` clobber fix **inverted** the bug — caller-supplied `stream_options` now caused `include_usage` to be skipped entirely, silently zeroing streamed usage. It's now deep-inserted so both the caller's keys and the usage chunk survive.
3. The `completion_tokens` switch ignored that OpenRouter gateways omit the field (`serde(default)` → 0), zeroing usage where the old subtraction worked. Now falls back to saturating `total − prompt`, with unit tests for both divergent and omitted accounting.
4. The same-role merge was baked into the shared sanitize helper unconditionally, fusing legitimate consecutive user turns (and their `name` attribution) on hyperbolic/mira, which never needed alternation. Merging is now opt-in, enabled only for perplexity.

**New `SUPPORTS_TOOLS` const** (perplexity/hyperbolic/mira): tools/tool_choice are dropped with a warning during conversion, *before* tool-choice validation — a multi-name `ToolChoice::Specific` no longer errors client-side on providers that warn-ignored it pre-migration. This replaces the `strip_unsupported_tools` prepare hooks at the right layer.

**Streaming robustness:** a delta carrying both `reasoning_content` and `reasoning` is no longer a serde duplicate-field error that dropped the whole chunk; `StreamingToolCall.index` defaults to 0 when omitted (Mistral marks it optional — the deleted mistral profile had this default); shared `CompletionResponse` defaults `object`/`created` so lax gateways (HF router sub-providers) aren't misclassified as error envelopes.

**API surface (while the major bump is free):** `openai::ToolChoice` is `#[non_exhaustive]` (OpenAI already defines more modes); `GenericCompletionModel`'s flag fields are private with the builder methods as the supported path, and the redundant `with_model` duplicate of `new` is removed; openrouter streaming spans now actually record `gen_ai.input.messages` (completing the round-three claim).

**Live verification:** ran the full `#[ignore]` llamafile live suite against a real llama.cpp `llama-server` — 23/29 pass including **all structured-output tests** (the `json_schema` response_format is sent and enforced by the real server); the 6 failures were proven model/server-side by reproducing one with a raw curl containing no rig-generated payload.

**Refuted (again):** two finders re-claimed moonshot previously sent single-name forced tools; `origin/main`'s `openai::ToolChoice::try_from` errored on every `Specific`, so the fail-fast is parity-correct — the comment now cites the exact pre-migration error to stop the recurring flag.

**New regression tests (7):** azure deployment-URL pinning under a per-request model override (recording client), openrouter divergent/omitted usage, openrouter encrypted-detail wire-id fallbacks ×2, video MIME round-trips ×2, perplexity multi-name tool-choice tolerance.

Post-fix: 13/13 cassette suites green (872 tests), 1,176 lib tests, 86 doctests, workspace check / clippy `--all-features` / fmt / docs clean.

PR #2088, opened 2026-07-10
Add SuperGrok OAuth to xAI provider

wey-gu:
## Summary

- add a separate SuperGrok OAuth client to the existing xAI provider while preserving the API-key client and its public types
- consume and safely refresh LiteLLM-compatible `auth.json` caches with cross-process locking, token rotation, atomic private writes, and bounded auth requests
- restrict OAuth bearer traffic to `https://api.x.ai`, disable refresh redirects, and avoid exposing OAuth response bodies

The default cache is `~/.config/litellm/xai_oauth/auth.json`. Callers can select another cache with `XAI_OAUTH_AUTH_FILE`, `XAI_OAUTH_TOKEN_DIR`, or `Client::builder().oauth().token_dir(...)`.

Rig consumes existing credentials; it does not add an interactive login flow. `XAI_API_BASE` remains API-key only. xAI's API-key verification endpoint does not accept subscription OAuth, so generic `.verify()` is explicitly unsupported for OAuth clients.

## Testing

- `cargo test -p rig-core providers::xai --lib --all-features`
- `cargo check -p rig-core --all-features --locked`
- `cargo check -p rig-core --no-default-features --features wasm --target wasm32-unknown-unknown --locked`
- `cargo check -p rig-core --target x86_64-pc-windows-gnu --locked`
- `cargo clippy -p rig-core --all-targets --all-features -- -D warnings`
- `cargo test -p rig-core --doc --all-features --locked`
- `cargo fmt --all --check`

The full `rig-core --all-features` suite passes 1,250 tests and has one unrelated Gemini image-generation failure that reproduces unchanged on `upstream/main`.


gold-silver-copper:
I'm for this feature, but there's a number of blockers before merge, please address them 🙏 

 ### Blockers

 - Refresh-token disclosure: token refresh follows 307/308 redirects after validating only the original URL, so its form body can reach an arbitrary host.
   auth/native.rs:119-138.
 - Bearer-token disclosure: OAuth fallback now honors arbitrary XAI_API_BASE; a configured non-xAI base receives the cached OAuth bearer token. client.rs:121,
   162-169.
 - Broken advertised LiteLLM interoperability: default lookup omits LiteLLM’s litellm/ directory, so normal existing caches are not found. client.rs:177.
 - Undocumented breaking public API: public XAiExt/XAiExtBuilder lose Copy/Default behavior and the client-builder API-key type changes, while changelogs call the
   release additive. client.rs:44-56.

 ### Decisions needing approval

 - OAuth verification: generic .verify() bypasses dynamic OAuth authorization. Either make it OAuth-aware, or explicitly document/test it as unsupported if
   /v1/api-key accepts only API keys.
 - Shared-cache concurrency: current lock is process-local; support for multiple processes sharing a rotating-token cache needs a per-cache cross-process lock.
   auth/mod.rs:62-66.

 ### Fixes worth doing now

 - Disable refresh redirects (or validate every hop); never retain raw OAuth error bodies. auth/native.rs:129-145.
 - Restrict OAuth requests to trusted xAI API origins, or do not apply XAI_API_BASE in OAuth mode.
 - Correct LiteLLM cache/env resolution and add default-path/preference tests.
 - Make Windows replacement preserve the prior cache on failure. auth/native.rs:270-299.
 - Add bounded connect/total auth-request timeouts.
 - Add rustdoc for public XAiAuth.
 - Add focused regressions for refresh rotation/failure, verify decision, and OAuth request paths.

wey-gu:
Addressed the review blockers in 05b4a8fa:

- disabled auth redirects, bounded auth requests, and removed raw OAuth response bodies from errors
- restricted OAuth bearer requests to the canonical xAI API origin
- corrected LiteLLM cache resolution and added cross-process cache locking
- preserved the existing API-key public types and builder contract
- documented and tested generic verify as unsupported for subscription OAuth
- preserved the prior Windows cache if replacement fails
- added focused regression coverage for redirects, origins, cache precedence, token rotation, locking, and compatibility

Native, wasm, Windows, clippy, doctest, docs, and the full upstream CI suite are green.

PR #2089, opened 2026-07-10
feat(vertexai): honor additional_params generation_config

### LINKED ISSUE #2067: feat(vertexai): map additional_params generation config (incl. thinking_config) into GenerationConfig
## Problem

`rig-vertexai` silently drops `CompletionRequest.additional_params`: `VertexCompletionRequest::generation_config()` (crates/rig-vertexai/src/types/completion_request.rs) maps only `temperature`, `max_tokens`, and `candidate_count` into `google_cloud_aiplatform_v1::model::GenerationConfig`, and nothing else in the request-conversion path reads `additional_params`.

Concretely, this means there is no way to set Gemini's thinking configuration (`thinking_budget` for Gemini 2.5, `thinking_level` for Gemini 3 / 3.5) through the Vertex AI provider — even though:

- the underlying `google-cloud-aiplatform-v1` crate fully supports it (`GenerationConfig::set_thinking_config`, `generation_config::ThinkingConfig`), and
- the `rig-core` Gemini provider already established a pattern for exactly this after #1520: it deserializes `additional_params` into a typed `AdditionalParameters` whose `GenerationConfig` carries `thinking_config { thinking_budget | thinking_level, include_thoughts }`.

So the same agent code gets thinking control on `providers::gemini` but loses it silently when moved to `rig-vertexai` — a surprising inconsistency between the two Gemini-family providers, and the silent drop makes it hard to notice (the request succeeds, just without the intended config).

## Proposed fix

Mirror the `rig-core` Gemini provider's contract in `rig-vertexai`:

- Deserialize `additional_params.generation_config` (same shape/field names as the gemini provider's `AdditionalParameters`, camelCase) in `VertexCompletionRequest::generation_config()`.
- Map `thinking_config` through to `google_cloud_aiplatform_v1`'s `GenerationConfig::set_thinking_config` (and pass through other recognized generation-config fields where the Vertex type supports them).
- Explicit precedence: typed request fields (`temperature`, `max_tokens`) win over `additional_params` duplicates, matching the principle that the typed surface is authoritative.
- Unit tests on the conversion (additional_params → GenerationConfig), including the thinking_budget/thinking_level mutual-exclusion cases documented on the gemini provider's `ThinkingConfig`.

Happy to submit a PR for this — it's a contained change to the request-conversion path with no public-API break (additional_params is already accepted and currently ignored).

domenic-donato:
## Problem

`VertexCompletionRequest::generation_config()` builds the Vertex `GenerationConfig` from only the typed `CompletionRequest` fields (`temperature`, `max_tokens`) and silently drops anything supplied via `additional_params` — so there is no way to set `thinking_config` (`thinking_budget`, `thinking_level`, `include_thoughts`) or `candidate_count` on a Vertex request, even though the Gemini provider accepts exactly those through `additional_params.generation_config`.

Fixes #2067

## Change

- `generation_config()` now honors `additional_params.generation_config`, accepting the same shape the Gemini provider already accepts (`rig_core::providers::gemini::completion::gemini_api_types::AdditionalParameters`), then applies the typed request fields **after** the `additional_params`-derived config so typed fields win over duplicates — mirroring the Gemini provider's own `create_request_body` precedence.
- It returns `Result<Option<GenerationConfig>, CompletionError>`: a malformed `additional_params.generation_config` now surfaces as a deserialization error instead of being silently ignored.
- Scope: `crates/rig-vertexai` only (2 files, +242/−6). No new markdown files; no public API surface removed.

## Tests

13 new unit tests in `types/completion_request.rs` covering: `thinking_config` and `candidate_count` pass-through, typed-field precedence over `additional_params` duplicates, empty/absent `additional_params` (byte-identical behavior to before), and the malformed-config error path. Cassette tests: N/A — this change is request-construction only (no live-response shapes changed); the existing suite covers the request path.

Ran locally at this head: `cargo fmt -- --check`, `cargo clippy -p rig-vertexai --all-features --all-targets` (clean, no new suppressions), `cargo test -p rig-vertexai` (all green).

## AI assistance disclosure

This patch was developed with substantial AI assistance (Claude, via Attuned Intelligence's engineering fleet), under human direction and review by @domenic-donato, who takes responsibility for the contribution. The issue analysis, implementation, and tests were AI-authored; the maintainer reviewed the change before and after submission.

### DIFF
--- a/crates/rig-vertexai/src/completion.rs
+++ b/crates/rig-vertexai/src/completion.rs
@@ -90,7 +90,7 @@ impl CompletionModelTrait for CompletionModel {
         let vertex_request = VertexCompletionRequest(request);
 
         let contents = vertex_request.contents()?;
-        let generation_config = vertex_request.generation_config();
+        let generation_config = vertex_request.generation_config()?;
         let system_instruction = vertex_request.system_instruction();
         let tools = vertex_request.tools();
         let tool_config = vertex_request.tool_config();
--- a/crates/rig-vertexai/src/types/completion_request.rs
+++ b/crates/rig-vertexai/src/types/completion_request.rs
@@ -1,6 +1,10 @@
 use crate::types::message::RigMessage;
 use google_cloud_aiplatform_v1 as vertexai;
 use rig_core::completion::CompletionError;
+use rig_core::providers::gemini::completion::gemini_api_types::{
+    AdditionalParameters, ThinkingConfig as GeminiThinkingConfig,
+    ThinkingLevel as GeminiThinkingLevel,
+};
 
 pub struct VertexCompletionRequest(pub rig_core::completion::CompletionRequest);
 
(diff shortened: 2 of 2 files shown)


domenic-donato:
Closing this — #2086 (merged yesterday) already solves this the right way, and I should have re-checked main before opening. Nice implementation there: the typed-field precedence with the range-validation bypass is cleaner than what this PR did, and the `set_candidate_count(1)` rationale (never request candidates the normalized response would silently discard) is a design decision this PR would have gotten wrong. Thanks — glad to see #2067 landed.

(Disclosure, consistent with the PR: authored with AI assistance under my direction and review.)

PR #2125, opened 2026-07-13
chore: rm unsused deps and folder ref

scffs:
I found that some stuff is outdated in the toml file. I ran only build and it works

FYI looks like some docstrings also outdated as well as the docs, I'll probably create issue in the future.

For example –> https://rig.rs/docs/concepts/tools#a-complete-example
There is no `definition` function anymore as I see error in my IDE.


scffs:
@gold-silver-copper, can you review, please?

PR #2162, opened 2026-07-16
fix(openai): Accept also null values for strict for tool calls

eriktews:
The OpenAI API allows the field strict for tool calls to be set to true, false or null. False is the default. Currently, rig only accepts true and false, but not null. This PR makes rig accept null as well and maps it to false.


gold-silver-copper:
Merged as stacked PR #2178 which preserves your attributions. Thanks for the find :)

Issue #1429, opened 2026-02-24
bug: `UserContent::document_url(url, media_type)` makes OpenAI throw 400 bad request


danechitoaie:
Using `UserContent::document_url(url, media_type)` with a PDF file throws the error bellow:

```
{
    "error": {
        "message": "Mutually exclusive parameters: 'input[1].content[0]'. Ensure you are only providing one of: 'file_id' or 'filename'.",
        "type": "invalid_request_error",
        "param": "input[1].content[0]",
        "code": "mutually_exclusive_parameters"
    }
}
```

Seems to be that the JSON sent to OpenAI is:
```
...
"content": [
                {
                    "file_url": "https://....../abc.pdf",
                    "filename": "document.pdf",
                    "type": "input_file"
                }
            ],
...
```

Notice the extra `"filename": "document.pdf",`.

If I send the same request from Postman to OpenAI without the `"filename": "document.pdf",` it works.

```
rig-core = { version = "0.31.0", features = ["derive"] }
```

Also Anthropic doesn't seem to suport URLs at all for PDFs, they only want data in base64.

joshua-mo-143:
Hey, thanks for the issue! Fixing this rn

edit: Anthropic does support URLs for PDFs. You can see this [very clearly in their docs.](https://platform.claude.com/docs/en/api/messages#url_pdf_source) Maybe we just need to change our side slightly to accommodate for their API shape?

joshua-mo-143:
Closing as this has now been fixed with the merge of #1431 #1432, please let us know if you have any other issues!

PR #2166, opened 2026-07-16
fix(openai): omit filename for URL-backed PDFs in Responses API requests

### LINKED ISSUE #2164: bug: URL-backed PDFs still get 400 mutually_exclusive_parameters from the OpenAI Responses API

## Bug Report

This is the same failure reported in #1429, which was closed as fixed by #1432 — but the fix landed on a parallel conversion path, not the one requests actually go through.

There are two conversions in `crates/rig-core/src/providers/openai/responses_api/mod.rs` that turn a rig message containing a `UserContent::Document` into Responses API types:

1. `impl TryFrom<crate::completion::Message> for Vec<InputItem>` — used by `impl TryFrom<(String, crate::completion::CompletionRequest)> for CompletionRequest`, i.e. the path `CompletionModel::completion()` requests are built through. Its PDF arm sets `filename: Some("document.pdf")` **unconditionally**, including for `DocumentSourceKind::Url` where it also sets `file_url`.
2. `impl TryFrom<message::Message> for Vec<Message>` — the conversion #1432 fixed (its URL arm correctly yields `filename: None`).

The Responses API treats `file_url` and `filename` as mutually exclusive on `input_file` content parts (see the [file inputs documentation](https://platform.openai.com/docs/guides/pdf-files)), so every completion request carrying a URL-backed PDF is rejected.

## Reproduction

```rust
let message = Message::User {
    content: OneOrMany::one(UserContent::document_url(
        "https://bitcoin.org/bitcoin.pdf",
        Some(DocumentMediaType::PDF),
    )),
};
// any completion through CompletionModel::completion() / an agent prompt
```

The serialized request contains:

```json
{"content":[{"file_url":"https://bitcoin.org/bitcoin.pdf","filename":"document.pdf","type":"input_file"}],"role":"user","type":"message"}
```

and the API responds:

```json
{"error":{"message":"Mutually exclusive parameters: 'input[0].content[0]'. Ensure you are only providing one of: 'file_id' or 'filename'.","type":"invalid_request_error","param":"input[0].content[0]","code":"mutually_exclusive_parameters"}}
```

Verified against the live API; the same request without `filename` is accepted.

## Expected behavior

URL-backed PDFs serialize with `file_url` only, as the path fixed by #1432 already does.

## Additional context

Affects at least v0.37.0 through v0.40.0 and current `main`. Both conversion paths already existed when #1432 merged, so this was missed rather than regressed. I have a fix ready (mirror #1432's change in the `Vec<InputItem>` PDF arm) and will open a PR.

dgrijalva:
Fixes #2164

## Problem

Two conversions in `responses_api/mod.rs` produce `input_file` content parts. #1432 (closing #1429) fixed `TryFrom<message::Message> for Vec<Message>`, but requests built by `CompletionModel::completion()` go through `TryFrom<crate::completion::Message> for Vec<InputItem>`, whose PDF arm still set `filename: Some("document.pdf")` unconditionally — including alongside `file_url`, which the API rejects with 400 `mutually_exclusive_parameters` ([file inputs docs](https://platform.openai.com/docs/guides/pdf-files)).

## Implementation

`filename` is now produced by the same `match` as `file_data`/`file_url`: `Some("document.pdf")` for base64 documents, `None` for URLs — mirroring what #1432 did in the parallel conversion. No public API changes.

## Testing

- Unit tests pin the serialized `input_file` JSON for URL PDFs on all three paths (`Vec<InputItem>`, the full wire request via `TryFrom<(String, CompletionRequest)>`, and `Vec<Message>`), plus base64 PDFs keeping their filename.
- New cassette test `url_pdf_document_prompt` **recorded** against the live API (recording fails with the 400 before this fix); full OpenAI cassette suite replays green.
- `cargo clippy --all-features --all-targets` and `cargo fmt -- --check` clean.

Related: #1429, #1432

### DIFF
--- a/crates/rig-core/src/providers/openai/responses_api/mod.rs
+++ b/crates/rig-core/src/providers/openai/responses_api/mod.rs
@@ -487,11 +487,13 @@ impl TryFrom<crate::completion::Message> for Vec<InputItem> {
                             media_type: Some(DocumentMediaType::PDF),
                             ..
                         }) => {
-                            let (file_data, file_url) = match data {
-                                DocumentSourceKind::Base64(data) => {
-                                    (Some(format!("data:application/pdf;base64,{data}")), None)
-                                }
-                                DocumentSourceKind::Url(url) => (None, Some(url)),
+                            let (file_data, file_url, filename) = match data {
+                                DocumentSourceKind::Base64(data) => (
+                                    Some(format!("data:application/pdf;base64,{data}")),
+                                    None,
+                                    Some("document.pdf".to_string()),
+                                ),
+                                DocumentSourceKind::Url(url) => (None, Some(url), None),
                                 DocumentSourceKind::Raw(_) => {
                                     return Err(CompletionError::RequestError(
                                         "Raw file data not supported, encode as base64 first"
@@ -512,7 +514,7 @@ impl TryFrom<crate::completion::Message> for Vec<InputItem> {
                                         file_id: None,
                                         file_data,
                                         file_url,
-                                        filename: Some("document.pdf".to_string()),
+                                        filename,
                                     }),
                                     name: None,
                                 }),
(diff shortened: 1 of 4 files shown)


gold-silver-copper:
Awesome. work, thanks 😎

PR #2167, opened 2026-07-16
fix(anthropic): support URL-backed PDF documents in requests

### LINKED ISSUE #2165: bug: URL-backed PDF documents are rejected by the Anthropic request conversion

## Bug Report

The Anthropic half of #1429. #1431 added the `DocumentSource::Url` wire variant and response-side parsing, but the request-side conversion (`crates/rig-core/src/providers/anthropic/completion.rs`, the `UserContent::Document` arm of `TryFrom<message::Message> for Message`) was not updated: it still rejects `DocumentSourceKind::Url` for PDFs with `"Only base64 encoded data is supported for PDF documents"` — even though the Anthropic Messages API supports URL-sourced PDFs via `"source": {"type": "url", ...}` ([PDF support documentation](https://docs.anthropic.com/en/docs/build-with-claude/pdf-support)).

## Reproduction

```rust
let message = Message::User {
    content: OneOrMany::one(UserContent::document_url(
        "https://bitcoin.org/bitcoin.pdf",
        Some(DocumentMediaType::PDF),
    )),
};
// any completion request → ConversionError before anything is sent
```

## Expected behavior

The document converts to `{"type": "document", "source": {"type": "url", "url": "..."}}`. The `DocumentSource::Url` variant added by #1431 already serializes to exactly this shape; it just isn't reachable from the request side. Verified against the live Messages API that this shape works.

## Additional context

Affects at least v0.37.0 through v0.40.0 and current `main`. I have a fix ready (map `DocumentSourceKind::Url` to `DocumentSource::Url` in the PDF arm) and will open a PR.

dgrijalva:
Fixes #2165

## Problem

#1431 (closing the Anthropic half of #1429) added `DocumentSource::Url` and response-side parsing, but the request-side `UserContent::Document` conversion still rejected URL-backed PDFs with `"Only base64 encoded data is supported for PDF documents"`. The Messages API supports URL-sourced PDFs ([docs](https://docs.anthropic.com/en/docs/build-with-claude/pdf-support)).

## Implementation

The PDF arm of the conversion now maps `DocumentSourceKind::Url` to `DocumentSource::Url { url }`, which serializes to `"source": {"type": "url", "url": ...}`. Base64/string handling is unchanged; the rejection error remains for other source kinds and now reads "Only base64 encoded data or URLs are supported for PDF documents". No public API changes.

## Testing

- Unit test pins the converted message's `source` JSON to `{"type": "url", "url": ...}`.
- New cassette test `url_pdf_document_prompt` **recorded** against the live API (recording fails with a ConversionError before this fix); full Anthropic cassette suite replays green.
- `cargo clippy --all-features --all-targets` and `cargo fmt -- --check` clean.

Related: #1429, #1431

### DIFF
--- a/crates/rig-core/src/providers/anthropic/completion.rs
+++ b/crates/rig-core/src/providers/anthropic/completion.rs
@@ -1274,21 +1274,19 @@ impl TryFrom<message::Message> for Message {
                         ))?;
 
                         let source = match media_type {
-                            DocumentMediaType::PDF => {
-                                let data = match data {
-                                    DocumentSourceKind::Base64(data)
-                                    | DocumentSourceKind::String(data) => data,
-                                    _ => {
-                                        return Err(MessageError::ConversionError(
-                                            "Only base64 encoded data is supported for PDF documents".into(),
-                                        ));
-                                    }
-                                };
-                                DocumentSource::Base64 {
+                            DocumentMediaType::PDF => match data {
+                                DocumentSourceKind::Base64(data)
+                                | DocumentSourceKind::String(data) => DocumentSource::Base64 {
                                     data,
                                     media_type: DocumentFormat::PDF,
+                                },
+                                DocumentSourceKind::Url(url) => DocumentSource::Url { url },
+                                _ => {
+                                    return Err(MessageError::ConversionError(
+                                        "Only base64 encoded data or URLs are supported for PDF documents".into(),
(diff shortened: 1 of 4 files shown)


gold-silver-copper:
Thank you for your contribution 🙏  It has been merged as stacked pr #2215

Issue #2156, opened 2026-07-15
Update Ollama implementation


gold-silver-copper:
Currently we treat the Ollama native API as if it was the OpenAI compatible API. We should distinguish and gives users the option between switching between the Ollama native, OpenAI, or Anthropic compatible APIs.

Inspired by this message from Erik Tews: There is one thing that is a bit surprising for me. Rig has a OllamaCompletionRequest with a max_tokens element, but apparently in Ollama there is no such option in the API: https://github.com/ollama/ollama/blob/main/docs/api.md

bugprone:
I'd like to work on this. I dug into the current implementation and here is what I found:

The endpoint is already the native `/api/chat`, but the request payload is still OpenAI-shaped: `OllamaCompletionRequest` serializes top-level `max_tokens` and `temperature`, neither of which exists in the native API, so Ollama silently ignores them (this is exactly the `max_tokens` mismatch Erik hit). The native equivalent of `max_tokens` is `options.num_predict`.

Proposed first PR, kept small per CONTRIBUTING:
- Make the serialized request strictly match the native API: map `CompletionRequest::max_tokens` to `options.num_predict` (without clobbering a user-provided `num_predict` in `additional_params`), keep `temperature` only inside `options`, and drop the two top-level fields
- Update the serialization tests that currently assert the OpenAI-shaped payload

For the second half (letting users switch between Ollama native / OpenAI-compatible / Anthropic-compatible APIs): before I write anything, what design do you have in mind? An API-mode option on `ollama::Client` that switches endpoint and wire types, or steering users to the existing `openai`/`anthropic` providers pointed at Ollama's compat endpoints, with docs? Happy to take that as a follow-up PR once the direction is agreed.

PR #2185, opened 2026-07-18
fix(ollama): send max_tokens as options.num_predict in native requests

bugprone:
Part of #2156 (the `max_tokens` half; the API-mode switching discussed there is left for a follow-up once the direction is agreed).

## Problem

The provider posts to the native `/api/chat` endpoint, but the payload was still OpenAI-shaped: top-level `max_tokens` and `temperature` fields that do not exist in the native API, so Ollama silently ignored them. This is the mismatch reported in #2156. The native equivalent of `max_tokens` is the `num_predict` model parameter inside `options`.

## Implementation

- Drop the top-level `max_tokens` and `temperature` fields from `OllamaCompletionRequest`
- Map `CompletionRequest::max_tokens` to `options.num_predict` and keep `temperature` inside `options` only
- An explicit `num_predict` or `temperature` in `additional_params` still wins, preserving the existing merge precedence
- One deliberate side effect: when `temperature` is unset, `options` no longer carries `"temperature": null` — the key is omitted instead

Updated the serialization tests that asserted the old payload shape and added a test for the `additional_params` precedence.

## Verification

Against a live Ollama (qwen2.5:0.5b) with `.max_tokens(10)`:

- before: 410-word response (`max_tokens` ignored)
- after: response truncated at 10 tokens

`cargo test -p rig-core --lib` (1341 passed), `cargo clippy --all-features`, and `cargo fmt --check` are all clean.


gold-silver-copper:
awesome thx!

Issue #1713, opened 2026-05-02
feat: Support AWS_BEARER_TOKEN_BEDROCK


DavidSouther:
- [x] I have looked for existing issues (including closed) about this

## Feature Request
Support AWS_BEARER_TOKEN_BEDROCK for Bedrock inference calls.

### Motivation
Bearer tokens make prototype iteration slightly easier when working with Bedrock. 

### Proposal
Haven't looked deeply; this might "just work" if this crate uses the aws-sdk-bedrock crate.

### Alternatives

The standard AWS_SECRET_ACCESS_KEY works for production workloads.

fangkangmi:
@gold-silver-copper I followed up on the note in #2309. Current main locks aws-sdk-bedrockruntime 1.124.0; that exact SDK version loads AWS_BEARER_TOKEN_BEDROCK through its Bedrock service configuration, selects HTTP bearer authentication, and includes an upstream request-capture test asserting the resulting Authorization: Bearer header. Rig constructs both its builder and from-env clients through aws_config followed by aws_sdk_bedrockruntime::Client::new, so no additional Rig implementation appears necessary. PR #2190 concerns the separate Mantle path and is not required for existing Bedrock Runtime/Converse support. Could we close #1713 as already supported? I have not made a live AWS call with a real bearer token; this conclusion is based on the exact locked SDK source, its upstream regression test, and Rig's current client construction path.

PR #2190, opened 2026-07-19
feat(rig-bedrock): add Bedrock Mantle OpenAI-compatible client

### LINKED ISSUE #2189: feat: Support Amazon Bedrock Mantle (OpenAI-compatible) for OpenAI-on-Bedrock models

## Feature Request
Add first-class support for **Amazon Bedrock Mantle** (OpenAI-compatible Chat Completions / Responses APIs) so Rig can call OpenAI-on-Bedrock models such as GPT OSS and other `openai.*` Mantle model ids.

Today `rig-bedrock` only speaks the Bedrock **Converse** API via `aws-sdk-bedrockruntime`. OpenAI models on Bedrock are primarily served through Mantle:

- Base URL shape: `https://bedrock-mantle.{region}.api.aws/openai/v1`
- Auth: Bedrock API key / short-term IAM bearer token (`Authorization: Bearer …`), not the Converse SDK client alone
- Model ids look like `openai.gpt-oss-120b`, `openai.gpt-oss-20b` (and other OpenAI-on-Mantle ids), not classic Converse FM ids alone

There is no existing issue covering Mantle. Closest related work is #1713 (`AWS_BEARER_TOKEN_BEDROCK`), which is about bearer auth for inference but does not introduce the Mantle OpenAI transport.

### Motivation
- OpenAI open-weight / OpenAI-on-Bedrock models are available on Mantle with OpenAI-compatible endpoints.
- Callers currently have to hand-roll an `openai::Client` with a custom base URL + token minting outside Rig.
- Converse model-constant additions alone do not cover Mantle Responses / Chat Completions usage (tools, reasoning params, etc.).

### Proposal
Extend `rig-bedrock` (preferred over a new companion crate) with a **Mantle path** that reuses Rig’s existing OpenAI-compatible client stack (Responses by default, matching `rig-core`’s OpenAI provider):

1. **Client construction**
   - `Client` / builder helpers that produce an OpenAI-compatible client pointed at Mantle for a given AWS region.
   - Auth options:
     - Short-term IAM token generated from the default AWS credential chain (presign `CallWithBearerToken`, encode as `bedrock-api-key-…` — same algorithm as the official AWS Bedrock token generators).
     - Optional long-lived / env bearer via `AWS_BEARER_TOKEN_BEDROCK` (can fully or partially address #1713 for the Mantle path).
2. **Model constants** for common Mantle OpenAI model ids (`openai.gpt-oss-20b`, `openai.gpt-oss-120b`, …). Free-form model strings remain supported.
3. **Docs + example** showing Mantle vs Converse: Converse stays the default for Claude / Nova / Llama; Mantle is the path for `openai.*` OpenAI-compatible models.
4. **Tests**: unit tests for token formatting / base URL; cassette-backed regression tests where feasible (same bar as other providers).

Non-goals for the initial PR:
- Replacing or rewriting the existing Converse client.
- Forcing model-id auto-routing magic that switches Converse vs Mantle based on string prefix (optional later; explicit Mantle client / constructor is clearer for v1).

### Alternatives
- Document “use `openai::Client` with a custom Mantle base URL” only (works today, but no first-class Bedrock auth, constants, or discoverability).
- Add Converse-only model id constants for OpenAI OSS runtime ids without Mantle (incomplete for Responses/Chat Completions and Mantle-only models).
- New companion crate `rig-bedrock-mantle` (extra surface area; Mantle is still Bedrock).

### References
- AWS Bedrock OpenAI models / Mantle endpoints: https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-openai.html
- Bedrock Mantle OpenAI-compatible base URL pattern: `https://bedrock-mantle.{region}.api.aws/openai/v1`
- Related: #1713

Happy to implement this against `rig-bedrock` following the OpenAI-compatible provider checklist in CONTRIBUTING.

felipefdl:
## Summary
Adds a first-class **Amazon Bedrock Mantle** path to `rig-bedrock` for OpenAI-on-Bedrock models (OpenAI-compatible Responses and Completions APIs).

Today `rig-bedrock` only speaks **Converse** via `aws-sdk-bedrockruntime`. OpenAI models on Bedrock are also served through **Mantle** (OpenAI-compatible HTTP). Converse is **unchanged**. There is **no** model-id auto-routing between Converse and Mantle.

### Live verification (`us-east-1`)

| Base URL | Models | API |
|----------|--------|-----|
| `https://bedrock-mantle.{region}.api.aws/v1` (**default**) | GPT-OSS and most Mantle catalog ids (`openai.gpt-oss-20b`, …) | Completions **and** Responses |
| `https://bedrock-mantle.{region}.api.aws/openai/v1` | GPT-5.x family (`openai.gpt-5.4`, `openai.gpt-5.6-luna` / `sol` / `terra`, …) | Responses |

Auth: short-term IAM bearer token (`bedrock-api-key-…`) or `AWS_BEARER_TOKEN_BEDROCK`.

### Public API (`rig_bedrock::mantle`)

- **First-class Mantle clients** (not OpenAI type aliases): `MantleResponsesExt` / `MantleCompletionsExt` with Mantle default base URL
- `ResponsesClient` / `CompletionsClient` (named to avoid collision with Converse `client::Client`)
- `ClientBuilder` → `build()` (Responses) / `build_completions()`
- `ClientBuilder::from_env()` is **sync** (env defaults only); free `from_env()` / `from_env_completions()` build clients async
- Short-term IAM token mint (SigV4 `CallWithBearerToken`); process-wide **multi-entry** cache keyed by `(region, access_key_id)`
- Token effective TTL = **min(12h, source AWS credential session expiry)**; cache refresh uses wall-clock **`SystemTime`**
- Token is **snapshotted at client build**; long-lived processes must rebuild before effective TTL
- Default base `/v1`; GPT-5.x helper `openai_gpt5_base_url` → `/openai/v1`
- Unversioned Mantle model constants (`OPENAI_GPT_OSS_20B` / `120B`, `OPENAI_GPT_5_4` / `5_5`, `OPENAI_GPT_5_6_LUNA` / `SOL` / `TERRA`)
- Versioned Runtime/Converse ids on `completion::OPENAI_GPT_OSS_*_VERSIONED` (not under Mantle)
- Typed `MantleError` with `#[source]` chains (`Token`, `Credentials`, `ClientBuild`, `CredentialsExpired`, `NoCredentialsProvider`)
- GenAI telemetry provider name: **`aws_bedrock`** (same family as Converse), including Responses streaming spans
- Capabilities: **chat completion only**; embeddings / image / audio / transcription / listing are `Nothing`
- Prefer `.additional_params(json!({"store": false}))` on Mantle Responses agents
- TLS features: `rustls` (default) / `native-tls`; Mantle HTTPS requires one of them (documented)

### rig-core interop

- Responses `created_at` accepts JSON integer **or float** (Mantle GPT-OSS returns floats); public type stays `u64`
- `ResponsesProviderExt::PROVIDER_NAME` so OpenAI-compatible backends can report a non-`openai` provider on Responses spans (including streaming)

### Example

```shell
export AWS_REGION=us-east-1
# optional: export AWS_BEARER_TOKEN_BEDROCK=bedrock-api-key-...
cargo run -p rig-bedrock --example agent_with_bedrock_mantle
```

### Docs

- Crate README dual-path table, token lifetime, TLS note, `store: false` gotcha
- Root README: TagoIO under “Who is using Rig”

Default Mantle base is `/v1`, not `/openai/v1`. Use `openai_gpt5_base_url` for GPT-5.x Responses (Luna / Sol / Terra). Versioned `-1:0` ids are Converse/Runtime only. Token algorithm matches official AWS Bedrock token generators. Non-goals for this PR: per-request bearer refresh on a long-lived client; Converse path changes; defaulting `store: false` inside `MantleResponsesExt` (needs a Responses request-shaping hook first).

## Test plan

- [x] `cargo test -p rig-bedrock --lib mantle`
- [x] `cargo test -p rig-core --lib created_at` (float + integer)
- [x] `cargo clippy -p rig-bedrock --all-targets -- -D warnings`
- [x] Live Mantle smoke (Completions + Responses, `openai.gpt-oss-20b`) via ignored tests / record runs
- [x] HTTP ProviderCassette replay (no AWS credentials):
  - `tests/cassettes/bedrock/mantle/agent/completion_smoke.yaml`
  - `tests/cassettes/bedrock/mantle/agent/responses_smoke.yaml`
  - `tests/cassettes/bedrock/mantle/agent/responses_streaming_smoke.yaml`
  - `tests/cassettes/bedrock/mantle/agent/gpt5_luna_responses_smoke.yaml`
  - `tests/cassettes/bedrock/mantle/agent/gpt5_sol_responses_smoke.yaml`
  - `tests/cassettes/bedrock/mantle/agent/gpt5_terra_responses_smoke.yaml`
- [x] Cassette safety registration for Mantle wrappers
- [x] Review feedback addressed (typed errors, wall-clock cache, multi-slot cache, streaming cassette, `pub(crate)` token helpers, versioned ids on `completion`, TLS docs)

## Related

- Closes #2189
- Partially addresses #1713 (`AWS_BEARER_TOKEN_BEDROCK` on the Mantle path)


FabianoEger (commented):
Reviewed against the provider implementation checklist in CONTRIBUTING.md, plus a closer pass over the token minting path and the rig-core interop changes.

## Overview

Adds a first-class Mantle (OpenAI-compatible) path to `rig-bedrock` alongside the untouched Converse path: dedicated `ResponsesClient` / `CompletionsClient` provider extensions, short-term IAM bearer token minting (SigV4 `CallWithBearerToken`), model constants, docs, an example, and cassette + ignored live tests. Two small rig-core changes support it: `ResponsesProviderExt::PROVIDER_NAME` for telemetry attribution and float `created_at` deserialization.

## What looks good

- Checklist compliance: Completions go through `GenericCompletionModel` via `OpenAICompatibleProvider`, Responses through `GenericResponsesCompletionModel` — no hand-rolled `CompletionModel` or request conversion. `Capabilities` marks everything but chat completion as `Nothing`, and the `ClientBuilder` API-key generic matches `ProviderBuilder::ApiKey` (`BearerAuth`).
- The token algorithm matches the official AWS Bedrock token generator (presign, scheme strip, `&Version=1` after signing, base64, prefix), and `effective_token_ttl` correctly caps `X-Amz-Expires` at the source credential session expiry — this is easy to get wrong and it is unit-tested both ways.
- Security posture is solid: `ClientBuilder`'s `Debug` redacts the key (tested), both provider builders default to Mantle base URLs with explicit tests asserting the bearer can never go to `api.openai.com`, cassettes are redacted, and the new cassette wrappers are registered with the safety scanner.
- `deserialize_created_at` rejects negative and non-finite values instead of silently truncating garbage, and keeps the public field `u64` so this is not a breaking change. The default `PROVIDER_NAME = "openai"` keeps existing `ResponsesProviderExt` impls source-compatible.

## Suggestions

1. **`MantleError` variants carry pre-formatted `String`s** (`Token(String)`, `Credentials(String)`, `ClientBuild(String)`). CONTRIBUTING lists `String` error types among the patterns to avoid; wrapping the underlying errors (`aws_sigv4` signing errors, `CredentialsError`, the client build error) with `#[source]`/`#[from]` would preserve the error chain for callers instead of flattening it at each `map_err`.
2. **`CachedToken.refresh_after` uses `Instant` for a wall-clock deadline** (the doc comment even says "wall-clock instant"). The token's real expiry is wall-clock (`X-Amz-Date` + `X-Amz-Expires`), but `Instant` is monotonic and can pause across host suspend on some platforms — a laptop that sleeps 3 hours can wake with a cache entry that looks fresh but is past its actual expiry. A `SystemTime` deadline avoids that class of stale-token bug.
3. **The token cache is single-slot.** It is keyed by `(region, access_key_id)` but stores only one entry, so a process alternating between two regions or credential sets re-resolves and re-mints on every alternation — and `resolve_credentials` runs a full `aws_config` load before every cache lookup. Not hot given tokens are minted at client build, but a small map keyed by `(region, access_key_id)` would make the cache actually hold.
4. **Public surface of the `token` module is broad**: `format_api_key_token`, `refresh_after_from_ttl`, and `TOKEN_TTL_SECS` (a duplicate of `TOKEN_TTL`) all become semver commitments. Unless they are intended as public API, `pub(crate)` would keep iteration room.
5. **The `store: false` gotcha is documented in five places** (crate docs, README, example, tests). If `ResponsesProviderExt` ever grows a request-shaping hook, `MantleResponsesExt` defaulting `store: false` would remove the footgun entirely — fine as a follow-up, but worth tracking.
6. **No Mantle streaming coverage.** The PR routes the Responses streaming span through `Ext::PROVIDER_NAME`, but there is no Mantle streaming cassette exercising that path. One streaming smoke (like the existing bedrock streaming cassettes) would lock in both the telemetry change and stream deserialization against Mantle's wire output.
7. Placement nit: `OPENAI_GPT_OSS_20B_VERSIONED` / `_120B_VERSIONED` are documented as Runtime/Converse ids yet live in the `mantle` module. The Converse `completion` module (next to the other Converse model constants) seems like the less confusing home.
8. Scope nit: the root README "Who is using Rig" addition is unrelated to the feature and could be split out per the small-and-focused PR guideline.

## Risks

- `rig-bedrock` now unconditionally enables rig-core's `image`, `audio`, and `reqwest` features for `Capabilities` associated-type unification, so Converse-only users pick up those dependencies. The inline comment explains why, which helps, but if the capability associated types ever become feature-independent this should be reverted.
- With the new `rustls`/`native-tls` features, a downstream `rig-bedrock` dependency with `default-features = false` and neither TLS feature will build but fail at runtime on the Mantle HTTPS path. A one-line note in the crate docs that Mantle requires one of the TLS features would save someone a confusing debugging session.
- The snapshotted-token limitation (no per-request refresh on long-lived clients) is clearly documented and listed as a non-goal, which is the right call for this PR — worth an issue so it does not get lost.

Overall this is a careful, well-tested addition that follows the generic client architecture as intended. Nothing here strikes me as blocking; items 1, 2, and 6 are the ones I would prioritize before marking ready.

felipefdl:
@FabianoEger thanks for the thorough pass. Addressed on `26c4de74` (and description updated):

1. **`MantleError`** — dropped flattened `String` variants. Now `Token` / `Credentials` / `ClientBuild` keep `#[source]` chains; added `CredentialsExpired` and `NoCredentialsProvider`.
2. **`refresh_after`** — switched from `Instant` to wall-clock `SystemTime` so host suspend cannot leave a stale “fresh” cache entry past real token expiry.
3. **Token cache** — multi-entry `HashMap` keyed by `(region, access_key_id)` instead of a single slot.
4. **Token module surface** — `format_api_key_token`, `refresh_after_from_ttl` (and the old `TOKEN_TTL_SECS` duplicate) are `pub(crate)` / removed from the public re-export; public mint path stays `TOKEN_TTL` + the generate helpers.
5. **`store: false` default in the ext** — left as a follow-up; still documented in several places. Needs a Responses request-shaping hook before `MantleResponsesExt` can set it without a footgun of its own.
6. **Streaming** — added `mantle/agent/responses_streaming_smoke` cassette (live-recorded, redacted, replay-green) so the Responses stream path and `PROVIDER_NAME = aws_bedrock` are locked in.
7. **Versioned model ids** — `OPENAI_GPT_OSS_*_VERSIONED` moved to `completion` next to the other Converse/Runtime constants.
8. **README “Who is using Rig”** — kept intentionally (TagoIO credit line). Happy to split if maintainers prefer a separate PR.
9. **TLS** — crate + Mantle docs note that Mantle HTTPS needs `rustls` (default) or `native-tls`.
10. **Snapshotted token** — still a documented non-goal for this PR (mint/cache at client build only). App can rebuild on a timer; live per-request refresh can be a follow-up issue if you want it tracked upstream.

Marking ready for review/merge. Happy to iterate on anything still off.

FabianoEger (approved):

PR #2196, opened 2026-07-21
OpenAI Responses API: Prevent duplicate reasoning stream events when encrypted_content is empty

boondocklabs:
## Description

OpenAI-compatible Responses API providers may return `encrypted_content` as an empty string on completed reasoning items.

Rig previously treated any present value, including `""`, as encrypted reasoning and emitted an additional empty `ReasoningContent::Encrypted` event. This resulted in consumers receiving duplicate reasoning completion events: one containing the visible reasoning text and another with empty content.

This change filters out empty encrypted reasoning while preserving non-empty encrypted payloads.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

Added a regression test confirming that a completed reasoning item containing visible reasoning text and `encrypted_content: ""` emits only the visible reasoning event.

## Checklist

- [x] I have added tests that prove the fix is effective

## Notes

Non-empty encrypted reasoning content continues to be emitted unchanged.


gold-silver-copper:
AWESOME PR THX

PR #2202, opened 2026-07-23
fix(openai) preserve multipart tool result content

### LINKED ISSUE #2201: OpenAI adapters lose multipart tool-result boundaries
## Description

When a Rig tool result contains multiple `ToolResultContent` blocks, the OpenAI provider adapters may concatenate those blocks before serializing the request.

For example:

```rust
OneOrMany::many(vec![
    ToolResultContent::text("first"),
    ToolResultContent::text("second"),
])
```

#### Responses API

When a tool result contains multiple text or JSON blocks and no image, the blocks are joined with newline characters and emitted as a single string:

```text
first
second
```

The original block boundaries are lost.

Mixed results containing an image are instead serialized as ordered rich-content entries.

#### Chat Completions API

Multiple text or JSON blocks are joined during the initial conversion into the OpenAI message type.

As a result, enabling `tool_result_array_content` produces a one-element content array containing the already-concatenated string, rather than one array element per original `ToolResultContent`.

### Expected behavior

Tool-result content boundaries and ordering should be preserved until the provider-specific wire-format normalization step.

For the Responses API:

* A single text or JSON block may continue to use the compact string representation.
* Multiple blocks should be serialized as ordered function-output content entries.
* Mixed text, JSON, and image content should preserve its original order.

For Chat Completions:

* Multiple text and JSON blocks should initially convert into distinct content entries.
* When `tool_result_array_content` is enabled, those entries should remain separate.
* When `tool_result_array_content` is disabled, the entries may be deliberately flattened into a newline-separated string for compatibility with providers that require string-valued tool results.

### Why this matters

`ToolResult.content` is represented as `OneOrMany<ToolResultContent>`. Flattening during initial provider conversion loses information represented by that type and prevents provider-specific serialization from preserving multipart content where the wire format supports it.

This is also ambiguous when individual text blocks contain newlines, because the consumer cannot reliably distinguish original block boundaries from newline characters within a block.

### Acceptance criteria

* Multiple text tool-result blocks retain their order and boundaries in the Responses API.
* Mixed text and JSON blocks retain their order and boundaries.
* Existing mixed text/image behavior remains ordered.
* A singleton text or JSON result retains the existing compact representation where applicable.
* Chat Completions array mode emits one content part per original text or JSON block.
* Chat Completions string mode continues to emit a newline-separated string.
* Regression tests cover both provider conversion and serialized wire output.

### Rig revision

Tested against:

```text
87f3f5b77a3caeffa10d60225c41e386753bf05e
```

KaasPeer1:
Fixes #2201

## Summary

Preserve multipart tool-result content through the OpenAI provider conversion paths.

### Responses API

* Use the compact string representation for a single text or JSON result.
* Serialize multiple tool-result blocks as ordered function-output content entries.
* Preserve ordering across text, JSON, and image content.

### Chat Completions API

* Convert multiple text and JSON tool-result blocks into distinct provider content entries.
* Preserve those entries when `tool_result_array_content` is enabled.
* Flatten multipart content only during request normalization when string-form content is required.

## Compatibility

The default string-form behavior for OpenAI-compatible providers is unchanged.

When `tool_result_array_content` is disabled, multipart text content is still serialized as a newline-separated string. The behavioral change applies when the selected wire format supports multipart content.

## Testing

Added regression coverage for:

* Singleton text and JSON tool results.
* Multiple ordered text blocks.
* Mixed text and JSON blocks.
* Mixed text and image blocks.
* Chat Completions array-form tool results.
* Chat Completions string-form compatibility behavior.
* Serialized provider request output.

## Type of change

* [x] Bug fix
* [ ] New feature
* [ ] Breaking change
* [ ] Documentation update

## Validation

* [x] `cargo fmt --all -- --check`
* [x] Targeted OpenAI provider unit tests
* [x] `cargo test -p rig-core --all-features --lib`
* [x] `cargo clippy --all-targets --all-features`

### DIFF
--- a/crates/rig-core/src/providers/openai/completion/mod.rs
+++ b/crates/rig-core/src/providers/openai/completion/mod.rs
@@ -602,27 +602,29 @@ impl TryFrom<message::ToolResult> for Message {
     type Error = message::MessageError;
 
     fn try_from(value: message::ToolResult) -> Result<Self, Self::Error> {
-        let text = value
+        let parts = value
             .content
             .into_iter()
-            .map(|content| {
-                match content {
-                message::ToolResultContent::Text(message::Text { text, .. }) => Ok(text),
-                message::ToolResultContent::Json { value } => Ok(value.to_string()),
+            .map(|content| match content {
+                message::ToolResultContent::Text(message::Text { text, .. }) => Ok(ToolResultContent::from(text)),
+                message::ToolResultContent::Json { value } => Ok(ToolResultContent::from(value.to_string())),
                 message::ToolResultContent::Image(_) => Err(message::MessageError::ConversionError(
-                    "OpenAI does not support images in tool results. Tool results must be text."
+                    "OpenAI Chat Completions does not support images in tool results. Tool results must be text."
                         .into(),
                 )),
-            }
             })
-            .collect::<Result<Vec<_>, _>>()?
-            .join("\n");
+            .collect::<Result<Vec<_>, _>>()?;
+
+        let content = match parts.as_slice() {
+            [part] => ToolResultContentValue::String(part.text.clone()),
+            _ => ToolResultContentValue::Array(parts),
+        };
 
         Ok(Message::ToolResult {
             // `call_id` carries the provider-issued call id when it differs
             // from the rig-level tool-result id (e.g. Mistral, llama.cpp).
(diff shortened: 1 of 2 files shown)


gold-silver-copper:
Merged as stacked PR #2217, ty for ur contribution

PR #2207, opened 2026-07-24
refactor(derive)!: single resolution authority, coherent required semantics, dependency hygiene

gold-silver-copper:
**TL;DR:** `rig-derive` gets one source of truth for each of the three things it
was previously patching locally: crate identity (one `CrateRefs` resolver shared
by path emission *and* context classification), `required` semantics (derived
from the parameter types, so the advertised schema always matches the
deserializer), and generated-code dependencies (resolved through `rig-core`
re-exports instead of the caller's Cargo.toml).

## Why

- **Renamed-dependency bug:** the context classifier hardcoded `rig`/`rig_agent`
  while path generation used `proc_macro_crate`, so
  `&mut rig_facade::tool::ToolContext` under a renamed dependency was
  misclassified as a model parameter and failed with a confusing derive error.
- **`required(...)` was a schema-only claim:** `required(a)` on
  `fn add(a: i32, b: i32)` advertised `b` as optional, but the generated
  deserializer still demanded it — a guaranteed runtime failure whenever the
  model legitimately omitted `b`. The macro also blindly overwrote the
  `required` array schemars computes correctly, and typos in
  `params(...)`/`required(...)` were silently ignored.
- **Dependency leaks:** generated code referenced bare `serde`/`serde_json`
  (every downstream crate needed both as direct deps — the rename fixtures
  prove it), and the `Embed` derive emitted an *unqualified* `impl Embed` that
  only compiled if the caller had the trait imported.

## What changed

- **`CrateRefs` (new `resolve.rs`)** — crate identity resolved once per
  expansion; the classifier matches resolved names (renames + `crate`
  self-references included), the four near-duplicate resolver functions are
  gone, and a contextual tool with neither `rig` nor `rig-agent` reachable gets
  a targeted `compile_error!` instead of an unresolved `::rig_agent` path.
- **Type-driven `required`** [**breaking**] — without `required(...)`,
  non-`Option` params are required and `Option<T>` params are optional
  (previously `Option` params were advertised required); with `required(...)`,
  omitted params get `#[serde(default)]`, so a non-`Option`, non-`Default` type
  is a compile error instead of a production deserialization failure. Names in
  `params(...)`/`required(...)` must match real parameters; malformed or
  duplicate attribute entries error instead of being silently dropped.
- **Dependency hygiene** — `rig-core` now re-exports `serde` and `serde_json`
  (next to the existing `schemars`); generated code resolves all three through
  the core namespace with `#[serde(crate)]`/`#[schemars(crate)]`, and the
  `Embed` derive emits fully qualified impls and where-clause bounds. The
  rename fixtures dropped their direct `serde`/`serde_json` deps — compiling
  them now *proves* the hygiene permanently.
- **Schema generation** — `schema_for!(...).to_value()` (infallible) inside a
  `LazyLock`: no generated `expect` (downstream `deny(clippy::expect_used)` is
  unaffected), schema built once instead of per call.
- **Structure** — `lib.rs` (867 lines) split into `resolve.rs`,
  `tool/{args,classify,expand}.rs`, and `embed/`.

## Migration

See the new `MIGRATING.md` section. The two observable breaking changes:
`Option` params are no longer listed in `required` by default (list them
explicitly if a provider needs that), and inconsistent `required(...)` lists
that previously compiled-but-failed-at-runtime now fail at compile time.

## Verification

- `cargo build --workspace --all-features --all-targets` — clean
- `cargo clippy --workspace --all-features --all-targets -- -D warnings` — exit 0
- `cargo test --workspace --all-features` — pass (153 green suites, incl. all
  cassette tests; the one cassette-pinned `#[rig_tool]` schema has all
  non-`Option` params, so its wire format is unchanged)
- `cargo fmt --all --check` — clean
- WASM lanes: `rig-core --features wasm`, `rig-agent`, `rig` on
  `wasm32-unknown-unknown` — clean (`LazyLock` is wasm-safe)
- `cargo check -p rig --no-default-features --features {derive,test-utils,agent}` — pass
- New tests: unmarked fully-qualified renamed-context fixtures (the exact case
  that broke), a compile-fail fixture asserting the targeted no-runtime-crate
  diagnostic, 5 trybuild cases for the strict attribute grammar, and runtime
  tests proving an omitted optional-by-`required(...)` param deserializes via
  `Default` while a listed one stays required.


gold-silver-copper:
Pushed `30a63ba4` addressing two review findings:

1. **`required(...)` now rejects `Option<T>` parameters** with a spanned compile error. Empirically, schemars excludes `Option` fields from the schema's `required` array regardless of attributes, and serde deserializes a missing `Option` to `None` — so listing one was silently ignored on both sides rather than honored. A new trybuild case (`fail_required_option_param`) pins the diagnostic.
2. **Docs corrected**: the MIGRATING.md claim that listed `Option` params "must then be present (possibly `null`)" was false and is rewritten; both changelogs and the macro docs now also note this rejection and that a wildcard context binding (`#[rig(context)] _: &mut ToolContext`) errors — name it `_context` instead.

Verification: `cargo test -p rig-derive`, `cargo clippy --workspace --all-features --all-targets -- -D warnings`, and `cargo fmt --all --check` all clean.

PR #2236, opened 2026-07-31
fix(anthropic): surface streaming stop_reason and message_delta input_tokens

### LINKED ISSUE #2235: bug(anthropic): streaming drops stop_reason and message_delta input_tokens

## Bug Report

The Anthropic streaming provider parses two pieces of metadata off the wire and
then drops them before the consumer can see them.

**1. `stop_reason` never reaches the consumer.** `message_delta.delta.stop_reason`
is read in `providers/anthropic/streaming.rs` only to decide when to break out of
the SSE loop; the value itself is discarded. `StreamingCompletionResponse` carries
`usage` and nothing else, so a `max_tokens` truncation is indistinguishable from a
normal `end_turn`.

This matters for any consumer that has to refuse to act on a truncated response.
In our case the model's tool call may be cut mid-arguments — we need to know the
turn ended because of `max_tokens` so we can decline to execute it. Today that
signal is unavailable through the public streaming API.

The OpenAI-compatible side has the same gap for the same reason:
`providers/internal/openai_chat_completions_compatible.rs` collapses finish
reasons to `pub(crate) enum CompatibleFinishReason { ToolCalls, Other }`, so
`length` is lost at `providers/openai/completion/streaming.rs`. This issue only
proposes fixing the Anthropic side, since that is where a typed
`stop_reason` already exists on the response object.

**2. `input_tokens` is read only from `message_start`.** In the same function,
`input_tokens` is captured at `message_start` and the `message_delta` event's
`usage.input_tokens` is ignored, even though `PartialUsage` already models the
field. Anthropic itself reports input usage on `message_start`, so this works
against the real API — but Anthropic-*compatible* gateways do not all behave that
way. OpenRouter's Anthropic Messages endpoint sends `input_tokens: 0` on
`message_start` and the real count on `message_delta`:

```
message_start  {"input_tokens": 0, "output_tokens": 0, ...}
message_delta  {"input_tokens": 9, "output_tokens": 4, ...}
```

The result is a silent `Usage { input_tokens: 0 }`, which is worse than a missing
value: a consumer that sizes its context window from input tokens sees zero and
never acts.

## Reproduction

Against OpenRouter's Anthropic Messages API
(`https://openrouter.ai/api/v1/messages`, model `anthropic/claude-haiku-4.5`):

```rust
let client = anthropic::Client::builder()
    .api_key(&key)
    .base_url("https://openrouter.ai/api")
    .build()?;
let model = client.completion_model("anthropic/claude-haiku-4.5");
let mut stream = model.stream(request).await?;
    if let Ok(StreamedAssistantContent::Final(final_response)) = item {
        // No stop_reason available at all.
        // final_response.usage.input_tokens == Some(0), even though the
        // provider sent 9 on message_delta.
    }
}
```

Both are also reproducible offline from a recorded interaction — see the cassette
in the accompanying PR, which replays exactly this shape.

## Expected behavior

- `anthropic::streaming::StreamingCompletionResponse` exposes the provider's
  `stop_reason`, so consumers can distinguish `max_tokens` from `end_turn`.
- `input_tokens` is taken from `message_delta.usage` when the provider supplies
  it there, falling back to `message_start` otherwise — leaving behaviour against
  Anthropic proper unchanged.

## Additional context

Closed #1297 and #1775 extended `Usage` for similar reasons, so surfacing
provider-reported metadata that the crate already parses seems in scope.

A PR implementing both, with a cassette-backed regression test, follows.

shlim33:
Fixes #2235

The Anthropic streaming provider parses two pieces of metadata off the wire and
then drops them before the consumer can see them. This surfaces both, without
changing behaviour against Anthropic proper.

## 1. Surface `stop_reason` on the streaming response

`message_delta.delta.stop_reason` was read only to decide when to break out of
the SSE loop; the value was discarded. `StreamingCompletionResponse` carried
`usage` and nothing else, so a `max_tokens` truncation was indistinguishable
from a normal `end_turn`.

```rust
pub struct StreamingCompletionResponse {
    pub usage: PartialUsage,
    #[serde(default)]
    pub stop_reason: Option<String>,   // added
}
```

The value is captured in the same block that already builds `final_usage`, and
filled in at the `FinalResponse` construction site. The two test-fixture
construction sites get `None`.

This matters for consumers that must refuse to act on a truncated response — a
tool call cut mid-arguments has to be declined, and `stop_reason` is the only
signal that says so.

## 2. Prefer `message_delta.usage.input_tokens` when the provider sends it

`input_tokens` was captured at `message_start` only, and `message_delta`'s
`usage.input_tokens` was ignored even though `PartialUsage` already models the
field. Anthropic reports input usage on `message_start`, so this worked against
the real API — but Anthropic-compatible gateways do not all behave that way.
OpenRouter's Anthropic Messages endpoint sends:

```
message_start  {"input_tokens": 0, "output_tokens": 0, ...}
message_delta  {"input_tokens": 9, "output_tokens": 4, ...}
```

which yielded a silent `Usage { input_tokens: 0 }` — worse than a missing value,
because a consumer sizing its context window from input tokens sees zero and
never acts.

```rust
input_tokens: usage
    .input_tokens
    .filter(|v| *v > 0)
    .or_else(|| usize::try_from(input_tokens).ok()),
```

**Behaviour against Anthropic proper is unchanged**: `PartialUsage.input_tokens`
is `#[serde(default)] Option<usize>`, and Anthropic's `message_delta` payload
does not carry the field, so the expression always falls through to the
`message_start` value there.

## Tests

`streaming_metadata_from_message_delta` — a cassette-backed regression covering
both changes in one recorded interaction: `message_start` reports
`input_tokens: 0` while `message_delta` carries `stop_reason: "max_tokens"` and
`input_tokens: 9`.

I verified the test is load-bearing rather than incidentally green: reverting
the `input_tokens` change alone turns it red with `left: Some(0), right: Some(9)`,
and reverting the `stop_reason` change removes the field the test reads.

`cargo test -p rig --test anthropic` → 109 passed / 0 failed.
`cargo clippy -p rig-core --all-targets` → clean.

## Notes

- The OpenAI-compatible side has the same gap for finish reasons
  (`CompatibleFinishReason { ToolCalls, Other }` is `pub(crate)` and `length` is
  lost at `providers/openai/completion/streaming.rs`). I left it out to keep this
  PR focused on the provider where a typed `stop_reason` already exists on the
  response object. Happy to open a separate issue for it if that is wanted.
- The two changes are in adjacent lines of the same function and share a root
  cause — the final response not carrying what `message_delta` reported. If you
  would rather review them separately I can split the PR.

### DIFF
--- a/crates/rig-core/src/providers/anthropic/streaming.rs
+++ b/crates/rig-core/src/providers/anthropic/streaming.rs
@@ -313,9 +313,20 @@ where
                                             // cache_creation_input_tokens and cache_read_input_tokens
                                             // are cumulative totals on message_delta.usage per the
                                             // Anthropic streaming API spec — use them directly.
+                                            // Prefer `message_delta.usage.input_tokens` when the
+                                            // provider supplies it. Anthropic itself reports input
+                                            // usage on `message_start`, but Anthropic-compatible
+                                            // gateways (e.g. OpenRouter) send `input_tokens: 0`
+                                            // there and the real count on `message_delta`. Reading
+                                            // `message_start` alone silently yields a zero prompt
+                                            // count against those providers, which breaks any
+                                            // consumer that sizes its context from input tokens.
                                             let usage = PartialUsage {
                                                  output_tokens: usage.output_tokens,
-                                                 input_tokens: usize::try_from(input_tokens).ok(),
+                                                 input_tokens: usage
+                                                     .input_tokens
+                                                     .filter(|v| *v > 0)
+                                                     .or_else(|| usize::try_from(input_tokens).ok()),
                                                  cache_creation_input_tokens: usage.cache_creation_input_tokens,
                                                  cache_read_input_tokens: usage.cache_read_input_tokens
                                             };
--- a/crates/rig-core/src/providers/anthropic/streaming.rs
+++ b/crates/rig-core/src/providers/anthropic/streaming.rs
@@ -196,6 +196,12 @@ struct ThinkingState {
 #[derive(Clone, Debug, Deserialize, Serialize)]
 pub struct StreamingCompletionResponse {
     pub usage: PartialUsage,
+    /// Anthropic `message_delta.stop_reason` (`"end_turn"`, `"tool_use"`,
+    /// `"max_tokens"`, `"stop_sequence"`). Surfaced so consumers can
+    /// distinguish a `max_tokens` truncation from a normal stop — the
+    /// streaming API otherwise drops it.
+    #[serde(default)]
+    pub stop_reason: Option<String>,
 }
 
 impl GetTokenUsage for StreamingCompletionResponse {
@@ -283,6 +289,7 @@ where
             let mut sse_stream = Box::pin(stream);
             let mut input_tokens = 0;
             let mut final_usage = None;
+            let mut final_stop_reason: Option<String> = None;
 
             let mut text_content = String::new();
 
@@ -316,6 +323,7 @@ where
                                             let span = tracing::Span::current();
                                             span.record_token_usage(&usage);
                                             final_usage = Some(usage);
+                                            final_stop_reason = delta.stop_reason.clone();
                                             break;
                                         }
                                     }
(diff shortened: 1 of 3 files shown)


gold-silver-copper:
Superseded by #2279, I have added you as coauthor to that PR. Thank you for your report and your work!

Issue #2118, opened 2026-07-12
epic(agent): roadmap for  interactive coding agents


gold-silver-copper:
## Summary

Track the roadmap for making Rig a strong foundation for production-grade, interactive coding agents: agents that can be steered while running, execute tools safely, preserve and resume sessions, compose child agents and tools, and support extensible host-defined workflows.

Rig already has many of the right building blocks:

- a shared streaming/non-streaming agent drive loop;
- `AgentHook`, `HookStack`, `StepEvent`, and fail-closed `Flow` semantics;
- dynamic `ToolServerHandle` registration;
- structured `ToolExecutionResult`, `ToolFailure`, and typed tool extensions;
- `ConversationMemory`, memory policies, demotion, and compaction;
- per-request tool selection and tool concurrency.

The remaining work is mostly about connecting these pieces into a coherent host/runtime contract. This epic coordinates existing issues and identifies missing child issues. It is not intended to land as one PR.

## Motivation

Good coding agents need more than a ReAct loop. Interactive harnesses such as Pi demonstrate the value of a small core with strong host control:

- steering and follow-up messages during an active run;
- cancellation that reaches model requests, tools, and child work;
- resumable, branchable sessions with compaction;
- dynamically registered and selected tools;
- lifecycle interception for policy, approvals, context, and observability;
- safe local, remote, and sandboxed execution;
- progressive-disclosure skills;
- subagent and composite-tool orchestration.

Rig should provide provider-agnostic primitives for these capabilities without hard-coding one TUI, sandbox, scripting language, storage backend, or orchestration style.

## Design principles

1. **Host control over product opinion.** `rig-core` should expose lifecycle and execution primitives; applications decide UI, permissions, storage, and workflow.
2. **One runtime contract.** Streaming and non-streaming paths must preserve the same control-flow and tool semantics.
3. **Cancellation is end-to-end.** Model calls, native tools, MCP tools, nested dispatch, and child agents should observe one run-scoped cancellation/deadline contract.
4. **Native tools remain first-class.** MCP and provider-hosted tools should integrate without forcing in-process Rust tools through an external protocol.
5. **Structured internally, serialized only at boundaries.** Host metadata and rich tool outputs should not depend on magic JSON string conventions.
6. **Concurrency must be explicit.** Call-scoped state, file mutations, output ordering, and parent/child relationships must remain correct under parallel execution.
[...]

Issue #2116, opened 2026-07-12
feat: Checkpoint and manual interrupt/resume operations


may-fly:
Why aren't mechanisms like checkpoints supported for manually triggering interruptions within tool methods? The hook-based approach isn't flexible enough.

gold-silver-copper:
Could you provide some more details as to what exactly you are trying to achieve?

nazq:
Opened #2244 as a focused decomposition of the resume half of this (per the #2118 Phase 1 task to clarify and decompose #2116): resuming an already-persisted AgentRun through the runner surface, scoped to the safe boundaries AgentRun already serializes. Implementation PR to follow there.

may-fly:
Okay, got it, thanks

PR #2245, opened 2026-08-03
feat(agent): resume a suspended AgentRun through the runner surface

### LINKED ISSUE #2244: feat(agent): resume a persisted AgentRun through the runner surface
Decomposed from #2116 per the #2118 Phase 1 task "clarify and decompose
#2116". Refs #1968; complementary to #2121, which adds safe-boundary
checkpoint capture for manually driven runs and lists runner-level durable
resumption as remaining work.

## Summary

`AgentRun` is a sans-IO, `Serialize + Deserialize` state machine, and its
module docs advertise persisting a run between steps and resuming it in
another process. The capture half works today by hand-driving `next_step()` /
`model_response()` / `tool_results()`.

The resume half is missing: `AgentRunner::run` / `stream` build their
`AgentRun` internally and cannot accept a restored one. Anyone resuming a
persisted run must hand-write the entire drive loop for the resume leg — and
thereby loses the runner's hook stack, tool-server dispatch (including MCP
tools), conversation-memory append, and telemetry. A process restart or an
approval-gate suspension leaves the second half of the run without the
runner's integrations.

## Proposal

One additive entry point that hands a deserialized run back to the existing
shared drive loop, so a resumed run behaves identically to one that was
never suspended:

```rust
let run: AgentRun = serde_json::from_str(&persisted)?;
let response = agent.resume(run).run().await?;   // or .stream().await
```

Scoped to the epic's stated boundary — "define safe pause/checkpoint
boundaries; do not attempt to serialize or suspend arbitrary Rust futures":
resume operates only on the states `AgentRun` already serializes (model
boundary, pending tool batch). No future suspension, no new checkpoint
mechanism — this is the missing consumer of the serialization `AgentRun`
already has.

## Acceptance criteria

- [ ] A restored run drives through the same engine as `run()`/`stream()` —
      no second loop; hooks, tool server, memory, and telemetry wired.
- [ ] A run restored with pending tool calls re-enters execution preserving
      `internal_call_id`s; behavior when the tool registry changed across
      the suspend/resume boundary is defined and documented.
- [ ] Runner builder options (`max_turns`, `add_hook`, …) keep working after
      resume, seeded from the suspended run's values.
- [ ] Conversation memory is not loaded on resume (the run carries its own
      history) and the completed run's messages are appended exactly once
      across the boundary.
- [ ] Usage aggregates across the suspending and resuming processes.

I have an implementation ready and will open a PR against this issue.

nazq:
Fixes #2244.

Related: #2116 (checkpoint / interrupt–resume operations), #2118 (interactive coding-agent roadmap). Complementary to #2121, which adds safe-boundary checkpoint capture for *manually driven* runs and explicitly lists runner-level durable checkpoint resumption as remaining work — this PR is that piece for the current runner, kept small and additive.

## Motivation

`AgentRun` is a sans-IO, `Serialize + Deserialize` state machine, and its module docs advertise exactly the workflow servers need: serialize a run between steps (for example while tool calls are pending), persist it, and resume it later in another process. The capture half of that story works today by hand-driving `next_step()` / `model_response()` / `tool_results()`.

The resume half is currently missing: the high-level driver (`AgentRunner::run` / `stream`) builds its `AgentRun` internally and has no way to accept a restored one. Anyone resuming a persisted run must hand-write the entire drive loop for the resume leg — and thereby loses the runner's hook stack, tool-server dispatch (including retrieval and MCP-registered tools), conversation-memory append, telemetry spans, and structured-output handling. In practice a process restart, migration, or approval-gate suspension leaves the second half of the run without those integrations.

This PR closes the loop with one additive entry point:

```rust
let run: AgentRun = serde_json::from_str(&persisted)?;
let response = agent.resume(run).run().await?;      // or .stream().await
```

## Design

**Public surface (additive only):**

- `Agent::resume(run: AgentRun) -> AgentRunner<M>` — mirrors `Agent::runner(prompt)`.
- `AgentRunner::resume(agent, run)` — mirrors `AgentRunner::from_agent(agent, prompt)`.

Everything else on the runner keeps working: `add_hook`, `max_turns`, `tool_concurrency`, `conversation` / `without_memory`, and both `run()` and `stream()`.

**One drive loop, not a fork.** `run()` and `stream()` already share the single engine (`drive_agent`), which takes an `AgentRun` as input. Resume reuses that seam: the runner carries an optional restored run, and both surfaces drive it through the identical loop, so hooks, tool execution, fail-closed semantics, memory append, and telemetry behave exactly as they do for a run that was never suspended. No second driver exists to drift.

**Division of authority.** The restored run is authoritative for loop state: prompt, accumulated history, pending tool calls, turn budget, invalid-tool-call retry budget, tool choice, and aggregated usage (a resumed run's final usage spans both processes). The agent is authoritative for the environment the run resumes into: model, hooks, tool server, preamble, request parameters, structured-output configuration, and memory. The run's budgets and tool-choice policy seed the runner on `resume()`, so builder overrides default to the suspended values and can still be changed before driving — e.g. `agent.resume(run).max_turns(extended)` deliberately raises an exhausted budget.

**Mid-tool-batch re-entry.** A run restored in the pending-tools state re-enters the drive loop directly on `CallTools` (the state machine already re-emits the pending calls idempotently, preserving `internal_call_id`s). Within one process the engine pins each turn's tool batch to the registry snapshot advertised to the model; those pinned handles cannot survive serialization, so a resumed batch takes a fresh snapshot and binds to the tool server's *current* registry. This is only permitted on the first step of a resumed drive — for every later step, a missing snapshot remains the driver-bug error it is today.

**Memory semantics.** Resume never *loads* conversation memory (the restored run carries its own history — loading would corrupt it), but a configured backend still receives the completed run's messages at `Done`, exactly once across the suspend/resume boundary: the suspending process never reached `Done`, so it never appended. `without_memory()` opts out, as usual.

## Alternatives considered

- **Making the private drive loop public** so callers can drive a restored run themselves. Rejected: it exposes a large internal surface (`TurnSource`, snapshot threading, span shaping) that would freeze implementation details into API, and every caller would still have to reassemble hooks/memory/telemetry correctly. `resume()` keeps one blessed driver.
- **A free-standing `resume(agent, run)` function.** Equivalent power, but the runner-builder shape is the repo's idiom and gives the override story (`max_turns`, `add_hook`, …) for free.
- **Re-pinning the original turn's tool definitions on mid-batch resume.** Not achievable — tool executors are not serializable — so the limitation is documented rather than worked around. The behavior (current-registry dispatch) is documented instead; the run still validates calls against the tool names that were advertised on the suspended turn.
- **Blocking memory append on resume** (treating a restored run like caller-supplied explicit history, which bypasses memory entirely). Rejected: the restored run's messages are precisely this conversation's new messages, and appending once at completion is what would have happened without the interruption; skipping it would silently drop the exchange from memory.

## Serialization caveats (documented on `Agent::resume` and the `run` module)

- Run state embeds the full conversation accumulated so far; persisting it inherits the conversation's sensitivity.
- The format is JSON-tolerant (serde defaults absorb absent fields) but carries **no cross-version stability guarantee yet**: resume with the same rig version that suspended the run.
- Runs must be suspended at a step boundary (the state observed between `next_step()` calls, e.g. pending tool calls). A run snapshotted while a model response was outstanding cannot be advanced and fails with the existing protocol error.
- The resuming agent should be configured equivalently to the one that started the run (tools, output mode, preamble); the run does not carry the environment.

## Test coverage

New tests in `rig-agent` (all synthetic — `MockCompletionModel`, `MockAddTool`, `CountingMemory`; no network):

- `resumed_mid_tool_batch_run_completes_with_hooks` — hand-drive a run to its tool boundary, serialize to JSON, drop it, restore, and resume via `agent.resume(run).run()`: the pending tool executes through the live tool server, `ToolCall`/`ToolResult`/`CompletionCall` hooks fire on the resumed leg, usage aggregates across the boundary, and the final history contains the pre-suspension turn.
- `resumed_run_streams_tool_activity_and_final_response` — the same restored run through `stream()`: tool result surfaces as a stream item, hooks fire, final response arrives.
- `resume_seeds_and_overrides_the_suspended_turn_budget` — an exhausted budget is preserved by default (`MaxTurnsError`) and extendable via `.max_turns()` before driving.
- `resumed_run_appends_to_memory_without_loading` — zero loads, exactly one append, stored history equals the response messages.
- `resume_at_the_model_boundary_runs_like_a_new_run` — a run serialized before its first model call resumes indistinguishably from a fresh run.

Validation: `cargo fmt --check`, `cargo clippy -p rig-agent --all-targets --all-features` (clean), `cargo test -p rig-agent --all-features` (492 lib tests + doctests, green) on the repo toolchain (1.94.0).

### DIFF
--- a/crates/rig-agent/src/agent/completion.rs
+++ b/crates/rig-agent/src/agent/completion.rs
@@ -1,6 +1,6 @@
 use super::hook::{HookStack, RequestPatch};
 use super::prompt_request::{self, PromptRequest};
-use super::run::OutputMode;
+use super::run::{AgentRun, OutputMode};
 use super::runner::AgentRunner;
 use crate::{
     agent::prompt_request::streaming::StreamingPromptRequest,
@@ -619,6 +619,45 @@ where
     /// [`AgentRunner::add_hook`], then call [`AgentRunner::run`].
     pub fn runner(&self, prompt: impl Into<Message>) -> AgentRunner<M> {
         AgentRunner::from_agent(self, prompt)
+    }
+
+    /// Build an [`AgentRunner`] that continues a previously suspended
+    /// [`AgentRun`] — one serialized between steps, persisted, and restored
+    /// (possibly in another process) — with this agent's hooks, tool server,
+    /// memory, and telemetry. [`AgentRunner::run`] and [`AgentRunner::stream`]
+    /// share their drive loop with a fresh run, so a resumed run behaves
+    /// identically to one that was never suspended.
+    ///
+    /// The restored run supplies the loop state (prompt, history, pending tool
+    /// calls, budgets, aggregated usage); this agent supplies the environment
+    /// it resumes into and should be configured like the one that started the
+    /// run — pinned per-turn tool handles do not survive serialization, so
+    /// pending tool calls execute against this agent's current registry.
+    /// Conversation memory is never loaded for a resumed run (the run carries
+    /// its own history), but a configured backend still receives the completed
+    /// run's messages at the end. See the [`run`](crate::agent::run) module
+    /// docs for serialization caveats.
+    ///
+    /// # Example
+    /// ```no_run
+    /// use rig_agent::{agent::run::AgentRun, prelude::*};
+    /// use rig_core::{client::ProviderClient, providers::openai};
+    ///
+    /// # async fn run() -> Result<(), Box<dyn std::error::Error>> {
+    /// let openai = openai::Client::from_env()?;
+    /// let agent = openai.agent(openai::GPT_5_2).build();
+    ///
+    /// // A run suspended earlier (possibly by another process).
+    /// let json = std::fs::read_to_string("suspended-run.json")?;
+    /// let run: AgentRun = serde_json::from_str(&json)?;
+    ///
+    /// let response = agent.resume(run).run().await?;
+    /// println!("{}", response.output);
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn resume(&self, run: AgentRun) -> AgentRunner<M> {
+        AgentRunner::resume(self, run)
     }
 
     /// Resolve the provider-facing tool definitions available for a prompt.
--- a/crates/rig-agent/src/agent/prompt_request/streaming.rs
+++ b/crates/rig-agent/src/agent/prompt_request/streaming.rs
@@ -505,6 +505,9 @@ where
         // immediately following CallTools step. This keeps the sans-IO run state
         // serializable while pinning execution to the definitions sent that turn.
         let mut pending_tool_snapshot: Option<Arc<ToolRegistrySnapshot>> = None;
+        // Only a restored run may open on CallTools without a pinned snapshot;
+        // on any later step a missing snapshot is a driver bug.
+        let mut first_step = true;
 
         'outer: loop {
             let step = match run.next_step() {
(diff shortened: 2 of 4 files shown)


gold-silver-copper:
Solved by #2443

Issue #2090, opened 2026-07-10
feat(core): expose a normalized finish reason / stop status on CompletionResponse


domenic-donato:
# Summary

The generic `CompletionResponse<T>` exposes no normalized finish reason / stop status — the signal that says *why* generation ended (natural stop, token-limit truncation, content filter, tool-call handoff). Each provider receives it and drops it in the mapping, so distinguishing "the model chose to stop" from "the output was cut off" requires re-parsing `raw_response` per provider — exactly the coupling the normalized type exists to remove. Additive fix: an `Option<FinishReason>` on `CompletionResponse`, populated per provider.

References against `56e15e16`.

# Current state

- `crates/rig-core/src/completion/request.rs:488` — `CompletionResponse<T>` carries `choice`, `usage`, `raw_response`, `message_id`; no finish/status field.
- `crates/rig-vertexai/src/types/completion_response.rs:15` — the `TryFrom<VertexGenerateContentOutput>` never reads `Candidate::finish_reason` (the only `set_finish_reason` calls in the crate are `#[cfg(test)]` fixture builders).
- openai Responses API — the `TryFrom` into the generic response ignores `response.status` (`Completed`/`Incomplete` + `incomplete_details`), dropping the same signal a second way.

Consequence: a truncated completion arrives looking identical to a finished one — partial text in `choice`, normal-looking `usage`, nothing that says "cut off." Retry/escalation policy and anything built on `AgentRun` (an empty-but-truncated turn vs an empty-but-complete one) can't branch on it without provider-coupled raw parsing. The OTel GenAI semantic conventions model `finish_reasons` as a first-class completion attribute for the same reason.

# Notes

- The enum design is the thing worth agreeing on first — something like `Stop`, `Length`, `ContentFilter`, `ToolCalls`, `Other(String)`, mapped per provider from the signal each already receives.
- `Option<FinishReason>` keeps it additive: providers that haven't wired it yet return `None`, no existing consumer breaks.
- Happy to open the PR (core type + vertexai and openai mappings) once the variant set looks right to you.

# Acceptance criteria

- `CompletionResponse` exposes a normalized finish reason; vertexai and openai (Responses) populate it from their existing signals; a truncated completion is distinguishable from a natural stop without touching `raw_response`.

---

*Disclosure: this issue was researched and written with AI assistance, under human direction and review.*

gold-silver-copper:
implemented in #2257

PR #2249, opened 2026-08-03
feat(openai): surface finish_reason on chat-completions streaming final responses

### LINKED ISSUE #2248: bug(openai): chat-completions-compatible streaming drops finish_reason
Refs #2090, #2235. This is the OpenAI-compatible mirror of #2235, which #2236
fixes for Anthropic and whose notes explicitly leave the OpenAI-compatible
side open.

## Summary

The Chat Completions streaming path parses each choice's `finish_reason`,
uses it once to decide when to flush pending tool calls, and then drops it.
The streaming `StreamingCompletionResponse` (in
`providers/openai/completion/streaming.rs`) carries `usage` and nothing else,
so a `max_tokens` truncation is indistinguishable from a natural stop. The
non-streaming response keeps the value (`choices[].finish_reason`); only
streaming consumers lose it.

Because the shared state machine in
`providers/internal/openai_chat_completions_compatible.rs` backs
`.completions_api()` and every provider riding `GenericCompletionModel`
(Azure, Groq, DeepSeek, Mistral, OpenRouter, …) plus Copilot chat, the gap
applies across all of them.

## Reproduce

Any streaming completion capped by `max_tokens`: the final response reports
normal-looking usage and no signal that the output was cut off. A consumer
that must refuse to act on a truncated response has nothing to branch on
without re-parsing provider SSE itself.

## Proposed fix

Provider-native, additive — the same shape #1774/#1776 established for
Gemini and #2236 uses for Anthropic: track the last streamed reason in the
shared compatible-stream state machine and surface it as an
`Option<FinishReason>` on the streaming final response. No change to the
generic `CompletionResponse` types — the normalized cross-provider enum is
#2090's design discussion and this doesn't pre-empt it.

I have an implementation with cassette-backed regression tests ready and
will open a PR against this issue.

nazq:
Fixes #2248.

Refs #2090, #2235.

The OpenAI Chat Completions streaming path parses each choice's `finish_reason`,
uses it once to decide when to flush pending tool calls, and then drops it. The
streaming `StreamingCompletionResponse` carried `usage` and nothing else, so a
`max_tokens` truncation was indistinguishable from a natural stop — the same gap
#2236 closes for Anthropic, and the one its notes explicitly left open for the
OpenAI-compatible side. The non-streaming response keeps the value
(`choices[].finish_reason` on the raw response); only streaming consumers lose it.

Reproduce with any streaming completion capped by `max_tokens`: the final
response reports normal-looking usage and no signal that the output was cut off.
A consumer that must refuse to act on a truncated response has nothing to branch
on without re-parsing provider SSE itself.

## What this does

`StreamingCompletionResponse` gains the field:

```rust
pub struct StreamingCompletionResponse<U = Usage> {
    pub usage: U,
    pub finish_reason: Option<FinishReason>,   // added
}
```

`FinishReason` is the typed enum this module already deserializes from the wire
(`ToolCalls`, `Stop`, `ContentFilter`, `Length`, `Other(String)`); it now derives
`Clone`, `Serialize`, and `Eq` so it can live on the response.

Internally, the shared compatible-stream state machine
(`providers/internal/openai_chat_completions_compatible.rs`) previously
normalized finish reasons into `CompatibleFinishReason { ToolCalls, Other }`,
which conflated "stopped for length", "stopped naturally", and "no finish reason
on this chunk". The enum now carries the full set, the choice field is
`Option<CompatibleFinishReason>` so absence is distinct from nonstandard, the
loop tracks the last streamed reason, and `build_final_response` receives it.

Because the state machine is shared, the fix lands in one place for:

- OpenAI's chat-completions dialect (`.completions_api()`), and every provider
  driving completions through `GenericCompletionModel` (Azure, Groq, DeepSeek,
  Mistral, OpenRouter, Together, Moonshot, MiniMax, Z.ai, Hugging Face,
  Perplexity, and the other compatible providers);
- Copilot chat, whose profile previously collapsed everything except
  `tool_calls` — it now surfaces the same typed value.

Two mapping details, both pinned by tests: the deprecated `function_call` wire
value keeps its existing tool-calls semantics and surfaces as
`FinishReason::ToolCalls`; nonstandard values some gateways emit (e.g.
DeepSeek's `insufficient_system_resource`) survive verbatim as
`Other(String)`.

## Design choice

This mirrors #1776, which added `finish_reason` to Gemini's streaming response
type, and #2236, which does the same for Anthropic: the provider-native typed
value on the provider's streaming final response. It deliberately does **not**
touch the generic `CompletionResponse` / `streaming::StreamingCompletionResponse`
types — #2090 tracks a normalized cross-provider `FinishReason`, and the variant
set there is still open. Once that lands, this field is exactly the input the
OpenAI-family mapping needs.

Alternatives considered:

- **Normalized enum on the generic response** — rejected here as #2090's design
  discussion; doing it per-provider first matches how `Usage` grew.
- **Raw `Option<String>`** — rejected; the typed enum already exists on the wire
  path and `Other(String)` preserves fidelity for nonstandard values.
- **Capturing inside each profile** — not possible; profiles are stateless by
  design, so the value has to flow through the shared loop.

## Compatibility

Additive. The field is `Option`, skipped when `None` on serialization, and
`Option` fields default on deserialization, so payloads persisted before this
change round-trip unchanged, and payloads written by this change deserialize
under older readers that ignore unknown fields. The reshaped
`CompatibleFinishReason` and the `build_final_response` signature are
`pub(crate)`. No public API is removed or changed in shape.

## Tests

Unit tests in `providers/openai/completion/streaming.rs`:

- `stop` / `length` / `content_filter` each survive to the final response, with
  usage intact;
- `tool_calls` surfaces as `ToolCalls`; deprecated `function_call` normalizes to
  `ToolCalls`;
- nonstandard `insufficient_system_resource` is preserved as `Other`;
- a stream with no finish reason yields `None`;
- serde round-trip: legacy payloads without the field deserialize, `None` keeps
  the serialized form unchanged, standard and nonstandard reasons round-trip.

Copilot: `chat_stream_surfaces_finish_reason_on_final_response` pins the typed
value through `CopilotStreamingResponse::Chat`.

Cassette: `completions_api_stream_surfaces_length_finish_reason` replays a
`max_tokens`-truncated chat-completions stream and asserts `Length` on the final
response. The fixture is authored in scrubbed cassette form (it passes the
cassette-safety scan and the registered-scenario check).

The tests are load-bearing, not incidentally green: reverting only the surface
(`finish_reason: None` at the final-response construction site) turns four of
them red.

Verified locally on the pinned toolchain:

- `cargo fmt --check`
- `cargo clippy -p rig-core --all-targets --all-features` — clean
- `cargo test -p rig-core --lib` — 952 passed
- `cargo test -p rig --all-features --test openai` — 130 passed
- `cargo test -p rig --all-features --test copilot` — 67 passed
- full cassette suites for groq, deepseek, mistral, openrouter, perplexity,
  doubleword, llamafile, chatgpt — all green

### DIFF
--- a/crates/rig-core/src/providers/copilot/mod.rs
+++ b/crates/rig-core/src/providers/copilot/mod.rs
@@ -1557,6 +1557,18 @@ enum ChatFinishReason {
     Other(String),
 }
 
+impl From<&ChatFinishReason> for CompatibleFinishReason {
+    fn from(reason: &ChatFinishReason) -> Self {
+        match reason {
+            ChatFinishReason::ToolCalls => Self::ToolCalls,
+            ChatFinishReason::Stop => Self::Stop,
+            ChatFinishReason::ContentFilter => Self::ContentFilter,
+            ChatFinishReason::Length => Self::Length,
+            ChatFinishReason::Other(other) => Self::Other(other.clone()),
+        }
+    }
+}
+
 #[derive(Deserialize, Debug)]
 struct ChatStreamingChoice {
     delta: ChatStreamingDelta,
@@ -1598,11 +1610,7 @@ impl CompatibleStreamProfile for CopilotChatCompatibleProfile {
                 data.usage,
                 &data.choices,
                 |choice| CompatibleChoiceData {
-                    finish_reason: if choice.finish_reason == Some(ChatFinishReason::ToolCalls) {
-                        CompatibleFinishReason::ToolCalls
-                    } else {
-                        CompatibleFinishReason::Other
-                    },
+                    finish_reason: choice.finish_reason.as_ref().map(Into::into),
                     text: choice.delta.content.clone(),
                     reasoning: choice.delta.reasoning_content.clone(),
                     tool_calls: openai_chat_completions_compatible::tool_call_chunks(
@@ -1614,9 +1622,14 @@ impl CompatibleStreamProfile for CopilotChatCompatibleProfile {
         ))
     }
 
-    fn build_final_response(&self, usage: Self::Usage) -> Self::FinalResponse {
+    fn build_final_response(
+        &self,
+        usage: Self::Usage,
+        finish_reason: Option<CompatibleFinishReason>,
+    ) -> Self::FinalResponse {
         CopilotStreamingResponse::Chat(openai::completion::streaming::StreamingCompletionResponse {
             usage,
+            finish_reason: finish_reason.map(Into::into),
         })
     }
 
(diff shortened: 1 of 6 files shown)


gold-silver-copper:
implemented in #2257

PR #2251, opened 2026-08-03
test(bedrock): cassette coverage for multi-tool_use streaming

### LINKED ISSUE #2250: bug: rig-bedrock streaming drops every non-last tool call in a multi-tool_use message

## Bug Report

When Claude on Bedrock emits **more than one `tool_use` block in a single streamed assistant message** (parallel tool calls), `rig-bedrock`'s Converse stream aggregation silently discards all but the **last** tool call.

Each earlier call streams its `ToolCallDelta`s completely — the argument JSON is fully present on the wire and assembles to valid, schema-matching JSON — but never yields a final `RawStreamingChoice::ToolCall`, so it is absent from the aggregated response. The model then typically re-issues the lost call on a later turn, which makes the failure silent: no error, just extra turns, extra tokens, and degraded multi-step behaviour.

### Root cause

`crates/rig-bedrock/src/streaming.rs` (line refs against `6cfae6d829da21f9dc9e775e065fee157b264f7e`) keeps a **single** in-flight tool-call slot:

- **L138** — `let mut current_tool_call: Option<ToolCallState> = None;`
- **L209** — the `ContentBlockStart::ToolUse` arm does `current_tool_call = Some(ToolCallState { .. })`, **overwriting** whatever was in the slot. A previous call that has not yet been emitted is silently lost here.
- **L224** — the `ContentBlockStop` arm finalizes `current_reasoning` only. It **never flushes the tool-call slot**, even though Converse guarantees the block and its `input` JSON are complete at this point.
- **L230–L245** — `MessageStop(StopReason::ToolUse)` does `current_tool_call.take()` and emits exactly **one** `RawStreamingChoice::ToolCall` — whichever call happened to be last.

Net effect: N `tool_use` blocks in one message yield N `ToolCallDelta` streams but only 1 completed tool call.

## Reproduction

Any Claude model on Bedrock via `CompletionModel::stream`, with a prompt and toolset that induce two tool calls in one turn:

```rust
let model = client.completion_model(bedrock::completion::ANTHROPIC_CLAUDE_SONNET_4);
let request = model
    .completion_request("Look up the athlete's readiness, and record a note about it.")
    .tool(echo_lookup_definition())   // two distinct tools
    .tool(record_note_definition())
    .build();

let mut stream = model.stream(request).await?;

let mut started = 0;
let mut completed = 0;
    match item? {
        StreamedAssistantContent::ToolCallDelta { content: ToolCallDeltaContent::Name(_), .. } => started += 1,
        StreamedAssistantContent::ToolCall { .. } => completed += 1,
        _ => {}
    }
}

// Observed: started == 2, completed == 1
// Expected: started == 2, completed == 2
```

### Observed stream shape for a 2-tool turn

Normalized events from one run (`echo_lookup` is issued first, `record_note` second):

```
tool_call_start   {id: tooluse_TJwgcVhLfytd6IfJub9BnE, name: echo_lookup}
tool_call_delta   x7    -> assembles to {"query": "..."}
tool_call_start   {id: tooluse_XB3V38Cw6PNj3JHDJ4pmsN, name: record_note}   <- slot overwritten here
tool_call_delta   x21   -> assembles to a valid nested object
tool_call_complete {id: tooluse_XB3V38Cw6PNj3JHDJ4pmsN, name: record_note}  <- only the LAST call completes
usage / message end        (aggregated response contains 1 tool call, not 2)
```

The dropped call's deltas are complete and well-formed — reassembling the fragments for `echo_lookup` above yields:

```json
{"query": "Jo Rider goal event readiness"}
```

So the data arrives intact; only the client-side aggregation drops it.

### Determinism

Measured over a 50-run harness (3 tools, a scripted scenario that forces a 2-tool first turn), model `us.anthropic.claude-sonnet-4-6`, region `us-east-1`, plain `ConverseStream` via `CompletionModel::stream`:

| | |
|---|---|
| runs | 50 |
| turns containing tool calls | 170 |
| **multi-tool turns** | **69** |
| **calls lost** | **69** |
| lost in non-last position | **69** |
| lost in last position | **0** |
| lost in single-tool turns | **0** |

Every non-last call in every multi-tool turn was lost, and in all 69 cases the streamed fragments concatenated to valid JSON matching the tool's schema.

## Expected behavior

Every `tool_use` block in the message should yield a `RawStreamingChoice::ToolCall`, so a message with N tool calls aggregates to N tool calls.

Downstream already supports this — `rig-core`'s `StreamingCompletionResponse` pushes each `RawStreamingChoice::ToolCall` onto `assistant_items` (a `Vec`), which is why the OpenAI path handles parallel calls correctly. The fix is confined to `rig-bedrock`.

### Proposed fix

Flush the tool call where the Converse protocol says it is complete — `ContentBlockStop` — instead of at `MessageStop`:

```rust
aws_bedrock::ConverseStreamOutput::ContentBlockStop(_event) => {
    if let Some(tool_call) = current_tool_call.take() {
        let tool_input = if tool_call.input_json.is_empty() {
            serde_json::json!({})
        } else {
            serde_json::from_str(tool_call.input_json.as_str())?
        };
        yield Ok(RawStreamingChoice::ToolCall(
            RawStreamingToolCall::new(tool_call.id, tool_call.name, tool_input)
                .with_internal_call_id(tool_call.internal_call_id)
        ));
    }
    if let Some(reasoning_state) = current_reasoning.take()
        && let Some(choice) = finalize_reasoning(reasoning_state) {
            yield Ok(choice)
        }
},
```

Notes:

- Text blocks also emit `ContentBlockStop`; the slot is `None` there, so `take()` is a no-op.
- The `MessageStop(StopReason::ToolUse)` arm currently yields `ProviderError("Failed to call tool")` when the slot is empty. After this change the slot is normally empty by then, so that `else` needs to become a no-op. Nothing in the tree asserts that error string — it appears only at its own definition site.
- Existing Bedrock cassette coverage (`tests/providers/bedrock/cassette/{streaming,raw_streaming}.rs`) exercises single-tool turns only, which is why this wasn't caught. A two-tool-use cassette in `raw_streaming.rs` asserting `observation.tool_calls.len() == 2` locks it in.

I have this fix working locally and am happy to open a PR with the cassette-backed regression test.

kjones:
## Description

**Rescoped — the fix this PR originally carried is superseded by #2257.** What remains is the cassette regression test.

This PR opened against #2250 with a fix moving the Bedrock tool-call emit from `MessageStop` to `ContentBlockStop`. #2257 has since landed a better version of the same idea: in-flight calls keyed by `contentBlockIndex` rather than a single slot, a defensive drain at `MessageStop` for streams that omit the stop event, and correct `MaxTokens` handling that drops an unfinished block instead of fabricating `{}` args. I've dropped my commit and rebased onto `main`.

The test is worth keeping on its own merits. #2257 added unit tests over `process_event` with synthetic `ConverseStreamOutput` values, which is the right place for the state-machine logic. This adds coverage one level out: a cassette recorded live against `amazon.nova-lite-v1:0` in `us-east-1`, replayed through the real AWS SDK event-stream decoding path, asserting that a message carrying two `tool_use` blocks surfaces both calls.

**It independently validates #2257 against real wire bytes** — I re-ran it against `4f0bc70` and it passes.

Fixes #2250

## Type of change

- [x] Bug fix — regression coverage for a fix that landed in #2257

## Testing

`raw_stream_surfaces_two_distinct_tool_calls` in `tests/providers/bedrock/cassette/raw_streaming.rs`, using the existing `TWO_TOOL_STREAM_*` fixtures and the `AlphaSignal`/`BetaSignal` zero-argument tools. No new fixtures were needed.

- [x] **Passes against `main` as of #2257** — `cargo test -p rig --test bedrock --features bedrock` → **58 passed, 0 failed**, including `cassette_safety::cassette_files_match_registered_scenarios` and `cassettes_do_not_contain_obvious_secrets`.
- [x] **Confirmed it catches the original defect.** Replayed against the pre-#2257 aggregation, the same cassette yields:
  ```
  every tool_use block should surface a tool call;
  missing lookup_harbor_label, saw ["lookup_orchard_label"]
  ```
  Only the last `tool_use` block in the message survived — so the test is a real regression guard, not a tautology against current behaviour.
- [x] `cargo clippy -p rig --test bedrock --features bedrock` clean; `cargo fmt --check` clean.

## Checklist:

- [ ] I have made corresponding changes to the documentation — test-only, none needed

## Notes

- **Assertions are inline rather than reusing `assert_raw_stream_contains_distinct_tool_calls_before_text`.** That helper also requires the tool calls to precede any text, and Nova Lite narrates at length before calling its tools. Ordering isn't what this test is about, so asserting it would couple the test to unrelated model behaviour.
- **Nova Lite rather than a Claude model.** The defect was in the stream aggregation, not the model, so anything emitting two `tool_use` blocks exercises it — and Nova keeps the new cassette consistent with the existing Bedrock ones. The original report was measured on `us.anthropic.claude-sonnet-4-6`.

Happy to close this instead if you'd rather not carry a cassette for a case the unit tests in #2257 already cover.


gold-silver-copper:
Good test TY!

PR #2262, opened 2026-08-06
Stream parts become entities: lifecycle grammar, opaque keys, and tool names as data (the 84a43e9e C→B→A program)

gold-silver-copper:
## What this is

The C → B → A follow-up to #2258, from the independent review at `84a43e9e`: its five latent findings collapse to three root causes, and this PR removes all three — plus one live wire bug the recording program found before any fix was written. One PR, three phases in order, each landed green.

The thesis, one sentence: **#2258 stopped identity leaking by typing provenance; this PR stops identity doing lifecycle's and data's jobs — parts get an explicit lifecycle and an opaque key, and data (provider ids, tool names) gets its own fields.**

## The recording program (first, and it paid off immediately)

Live recording against real providers preceded the fixes, per the record-first doctrine #2258 wrote into `tests/README.md`:

- **A live data-loss bug on main, found by the first recording attempt:** the Gemini Interactions wire announces a function call in `step.start` with `"arguments": {}`, streams the real payload as `arguments_delta` fragments, and closes with `step.stop`. rig neither modeled the frame (the stream *errored*) nor assembled the fragments (calls aggregated with `{}` args). Fixed by routing the wire's own start/delta/end lifecycle through the shared assembly; recorded live in `interactions_same_tool_twice`, which asserts the exact per-call payloads the pre-fix code dropped.
- New live recordings: ollama and interactions **same-tool-called-twice** turns (the same-name shape the old name-as-id scheme collapsed), an interactions **signature-without-summaries** turn (the empty-buffer signature arm, live), and **cross-provider tool-result replays** to Gemini and Ollama from an OpenAI-Chat-shaped history — the provider accepting `functionResponse.name: "add"` (not `call_abc`) is the evidence for phase C.
- Not live-coaxable, documented as such: a repeated `output_item.done` (finding #1 — real OpenAI won't repeat one; pinned at unit/property level) and an id-less OpenRouter encrypted detail (finding #4 — pinned at unit level).

## Phase C — `ToolResult` carries the executed tool's name as data (finding #5)

`message::ToolResult` gains `name: Option<String>` (serde-skipped when absent, so persisted histories round-trip). The agent drivers populate it with the *executed* call's name on every construction route, both blocking and streamed (parity pinned). Serializers read `name` first; the ~90-line history-pairing heuristic is demoted to a back-compat shim for `name: None` results — and its shape-3 branch is fixed: an id matching a paired call's identity is proof it's an identifier, so the resolved name is that call's function name, never `call_abc` replayed verbatim. One deliberate re-record: `repair_renames_tool_call_and_executes_it`, whose second-turn request now carries the history-consistent repaired name instead of the legacy name-in-id artifact.

Also in C (review nits, made structural): `adapter::warn_unmodeled` is the one payload-redaction policy (kind + byte size, never the payload) with a `driver_adoption` scan failing CI on any direct `warn!(?...)` payload capture; and `provider_layout` gains `WireCoverage::SharedWireOwnSemantics`, so openrouter's profile-level streaming semantics stop hiding behind a wire-verbatim exemption.

## Phase B — opaque stream keys; durable handles and correlators separate (findings #3, #4)

The raw-event identity is now `StreamPartId`: `Eq + Hash + Clone + Debug` and **nothing else** — no `Serialize`, no rendering, no accessor into the durable id space (extended trybuild compile-fail suite pins all three; the representation was later made fully private, and a `Composite` variant shipped here was removed once it proved caller-less). The durable handle travels separately as `WireId` (`provider_id` on reasoning events, `tool_id` on `RawStreamingToolCall`/`ToolInputEnd`), whose only constructor rejects the empty string — absence is `None`, the fabricated `Wire("")` class is unrepresentable, and its per-serializer empty-string filters are deleted. Public delta ids become rig-generated correlators (`ReasoningDelta` gains `provider_id`; `ToolCallDelta` drops `id`; the agent hook payload drops `tool_call_id`). This also closed a **latent leak**: the agent's delta-only assembly path could commit a minted key's public rendering as a durable `Reasoning::id` and send it upstream — the assembled block's id now comes only from `provider_id`. Finding #3's `Output:0` overlap is unobservable once nothing renders the key.

## Phase A — the raw grammar is a part lifecycle (findings #1, #2, and the F1b/B4/G5 class)

`{Text, Reasoning, ToolInput} × {Start, Delta, End}`: `RawStreamingChoice` gains `ReasoningStart`/`ReasoningEnd { reasoning, signature }`/`TextEnd`, and `ReasoningSignature` is deleted — a trailing signature is an `End` arriving late, and **one end primitive** covers a signature closing an open block, a trailing signature after interleaved output, and a signature-only stream. The accumulator reduces to open-maps into an arrival-ordered part list; **idempotence belongs to the entity**: a finished-set populated by every finalization route makes a repeated `ToolInputEnd` — even with an authoritative name/arguments payload, the exact finding-#1 bypass — finalize nothing, pinned by a route-generic proptest law. Boundary-less wires synthesize their ends (one `reasoning_open` bit replaced each thought/restatement buffer; the follow-up `chunk_lifecycle` refactor then moved even that bit out of the adapters — they declare a chunk's parts and a shared `MintedReasoningLifecycle` derives order and boundary ends); anthropic's stop event maps to `ReasoningEnd` with the assembled signature and its thinking-text buffer is gone. Consumers see the completed `Reasoning` block when it closes with a wire-authoritative payload; bare synthesized ends stay silent.

**Deleted:** `PartKey`/ordinals, `open_ordinal`, `open_minted_reasoning`, `close_minted_reasoning`, `ManagedPart::{DeltaBuilt, Complete}`, the sibling-append special case, `closed_by_full_call`, `reasoning_full`, `reasoning_signature`, `ReasoningSignature`, `PartId::render()`/`as_wire()`, and every adapter-side thought/restatement buffer.

## Findings, at close

| 84a43e9e finding | Status |
| --- | --- |
| #1 repeated authoritative end duplicates the call | structurally impossible (entity finished-set, every route) |
| #2 interactions missing the signature branch | the branch class no longer exists (one end primitive) |
| #3 `Output:0` kind overlap observable via `render()` | `render()` deleted; keys unobservable |
| #4 `Wire("")` bypasses provenance | unrepresentable (`WireId::new` rejects empty; minted key + `None`) |
| #5 identifier replayed as function name | name is data (`ToolResult::name`); shim resolves identity matches to the call's name |

## Cassettes

Recorded new: gemini `interactions_same_tool_twice`, `interactions_signature_without_summaries`, `chat_sourced_history_replay`; ollama `same_tool_twice`, `chat_sourced_history_replay`. Deliberately re-recorded (named in its commit): gemini `agent_run_recovery/repair_renames_tool_call_and_executes_it` (the shape-3 fix changes its second-turn request; the re-record also drifts recorded usage — `thoughtsTokenCount` 37→48, `totalTokenCount` 221→232). Every other recording replays byte-identical — zero drift at the final gate.

## Verification (at head, all phases)

`cargo nextest run --all-features` **3457 passed / 0 failed** (default-members set); the out-of-facade guard step **44/44**; `clippy --workspace --all-targets --all-features -D warnings` clean; `fmt` clean; `RUSTDOCFLAGS="-D warnings" cargo doc` clean. MIGRATING carries a section per phase; changelogs carry the `[**breaking**]` entries (rig-core and rig-agent).

## Deliberate deviations from the planning prompt

- Findings #1/#2 got no interim symptom patches — phase A (same PR) removes both classes structurally; the pre-fix defect shapes are pinned by recordings/tests instead.
- ~~The lifecycle laws live as accumulator-level unit + proptest cases rather than an extension of the public-item validator~~ — closed by the follow-up sequence-law validator: `providers/internal/sequence_law.rs` now checks the raw adapter event stream inside both shared drivers under `cfg(any(test, debug_assertions))`, so every conformance fixture and cassette replay validates what adapters actually emit; `assert_valid_event_stream` keeps its public-level laws and still runs over every corpus scenario and cassette drain.
- Bare synthesized ends don't yield public completion events (the prompt's "consumers get reasoning end" is satisfied only for wire-authoritative closes) — yielding fabricated completion events changed observable history-building, and the wire-said-something rule is the honest line.


g13g13:
## Code review findings\n\nReviewed commit c25f30b32da2721772a94e0a1246320b94d9b42d (PR #2262).\n\n### P1 — malformed tool input can resurrect a finished call\n\nIn crates/rig-core/src/streaming/parts.rs:685 (https://github.com/0xPlaygrounds/rig/blob/c25f30b32da2721772a94e0a1246320b94d9b42d/crates/rig-core/src/streaming/parts.rs#L685), the UnparseableToolInput::Error branch returns Err without inserting the stream key into finished_tools. The stream deliberately continues consuming events (streaming/mod.rs:1114).\n\nReproduction:\n1. Open a tool call and send malformed argument fragments.\n2. Finalize with UnparseableToolInput::Error and receive the error.\n3. Send another ToolInputEnd for the same key.\n\nBecause the key was never marked finished, the second end takes the no-open-call path and emits a new ToolCall. This violates the lifecycle idempotence guarantee and can cause a tool to execute after the consumer already observed a malformed-call error.\n\nPlease mark the key terminal before returning the error (or track failed keys separately if key reuse is intentional), and add a regression test for Error -> repeated end/continuation.\n\n### P2 — OpenAI Chat conversion fabricates an empty required tool name\n\nThe OpenAI Chat conversion writes name = "" at providers/openai/completion/mod.rs:1006 (https://github.com/0xPlaygrounds/rig/blob/c25f30b32da2721772a94e0a1246320b94d9b42d/crates/rig-core/src/providers/openai/completion/mod.rs#L1006). The canonical ToolResult.name is required for name-keyed providers (completion/message.rs:223).\n\nWhen an OpenAI Chat-shaped history is converted to the core message model and replayed through Gemini or Ollama, the request can carry an empty functionResponse.name or tool name instead of the executed tool name. That can make the provider reject the request or break cross-provider replay.\n\nPlease resolve the name from the paired assistant tool call when converting a history, or return an explicit conversion error rather than silently fabricating an invalid empty name. Add a cross-provider replay test covering this conversion path.\n\nI did not modify the repository or rerun CI for this review.

PR #2263, opened 2026-08-07
fix(cohere): Cohere provider sends request shapes the v2 chat API rejects, and ships removed model IDs

### LINKED ISSUE #2241: bug: Cohere provider sends request shapes the v2 chat API rejects, and ships removed model IDs

## Bug Report

Several parts of the Cohere provider (`crates/rig-core/src/providers/cohere/`) do not
match Cohere's v2 chat API. Each item below was reproduced against the live API on
2026-08-01 with a real `COHERE_API_KEY` on `rig` 0.41.0.

**1. Tool calling is broken end to end.** `ToolResultContent` is missing the
`#[serde(tag = "type")]` that every sibling content enum in the same file has, so tool
results serialize externally-tagged as `{"Text":{"text":"-3"}}`. Cohere returns 422, so
the second turn of every tool round trip fails.

**2. Context documents are rejected.** `CohereCompletionRequest.documents` is typed
`Vec<rig::completion::Document>` and serializes Rig's own `{id, text, ..}` shape. Cohere
requires `{id, data}`. Any agent built with `.context(..)` gets a 400. Note that the
correct `impl From<completion::Document> for Document` already exists in
`completion.rs` — it is simply never used for requests.

**3. `ToolChoice::Specific` is rejected.** Rig's `ToolChoice` serializes to a tagged
object; Cohere accepts only the bare strings `REQUIRED` / `NONE`, so `Specific` returns
422. (`Required` and `None` happen to work — Cohere accepts them case-insensitively —
so this affects `Specific` only.)

**4. Five of six completion model constants are dead.** Cohere removed `command`,
`command-light`, `command-r`, and `command-r-plus` on 2025-09-15;
`command-light-nightly` also 404s. `COMMAND_R` is the constant used by the crate-level
doc example, `examples/debate`, and `examples/agent_with_context`.

| Constant | Model ID | Live status |
|---|---|---|
| `COMMAND_R` | `command-r` | 404 |
| `COMMAND_R_PLUS` | `command-r-plus` | 404 |
| `COMMAND` | `command` | 404 |
| `COMMAND_LIGHT` | `command-light` | 404 |
| `COMMAND_LIGHT_NIGHTLY` | `command-light-nightly` | 404 |
| `COMMAND_NIGHTLY` | `command-nightly` | 200 (works, but absent from Cohere's published catalogue) |

No constants exist for anything current: `command-a-plus-05-2026`, `command-a-03-2025`,
`command-a-reasoning-08-2025`, `command-a-vision-07-2025`, `command-a-translate-08-2025`,
`command-r7b-12-2024`, `command-r-plus-08-2024`, `command-r-08-2024`, or `embed-v4.0`.

**5. Error status and body are lost.** `CompletionModel::completion` wraps the transport
error with `.map_err(|e| http_client::Error::Instance(e.into()))`, which hides the
`InvalidStatusCodeWithMessage` that `CompletionError::provider_response_status()` and
`provider_response_body()` read from. Both return `None` on every real HTTP failure, so
callers cannot see the status or Cohere's `{"id", "message"}` error envelope. The
existing unit tests pass only because `RecordingHttpClient` returns `Ok` with a status
rather than `Err`.

**6. Token usage is under-reported and inconsistent.** `impl GetTokenUsage for Usage`
reads `usage.billed_units`, while `TryFrom<CompletionResponse>` and the streaming final
read `usage.tokens` — so `response.usage` and the telemetry span disagree for the same
response. `billed_units` excludes cached and system-prompt overhead; on a recorded
response the two differ by ~20x (27/66 vs 556/69). Cohere's `usage.cached_tokens` is
also dropped entirely, though `rig::completion::Usage::cached_input_tokens` exists to
receive it (512 of 556 input tokens were cached on that same response).

**7. `max_tokens` is silently dropped.** `CohereCompletionRequest` has no `max_tokens`
field, so `.max_tokens(n)` never reaches `/v2/chat` even though the endpoint supports it
— a request with `"max_tokens":24` returns `finish_reason: MAX_TOKENS` at exactly 24
output tokens. Callers get unbounded generations and unbounded cost.

## Reproduction

Each of these fails on 0.41.0:

```rust
use rig::completion::Prompt;
use rig::prelude::*;
use rig::providers::cohere;

# async fn run() -> Result<(), Box<dyn std::error::Error>> {
let client = cohere::Client::from_env()?;

// (4) 404: `command-r` was removed by Cohere on 2025-09-15.
let agent = client.agent(cohere::COMMAND_R).build();

// (2) 400: documents[0].data is required and must have at least one field
let agent = client
    .agent("command-a-03-2025")
    .context("A glarb-glarb is an ancient farming tool.")
    .build();
agent.prompt("What is a glarb-glarb?").await?;

// (1) 422 on the second turn: tool results serialize as {"Text":{...}}
let agent = client.agent("command-a-03-2025").tool(Subtract).build();
agent.prompt("Calculate 2 - 5.").await?;
# Ok(())
# }
```

The equivalent raw requests, with Cohere's actual responses:

```bash
# (2) documents
-d '{"model":"command-a-03-2025","messages":[{"role":"user","content":"hi"}],
     "documents":[{"id":"doc_1","text":"hello"}]}'
# 400 {"message":"documents[0].data is required and must have at least one field"}

# (1) tool result content
-d '... "messages":[..., {"role":"tool","tool_call_id":"subtract_1",
                          "content":[{"Text":{"text":"-3"}}]}] ...'
# 422 {"message":"invalid type: parameter 'messages.2.content.0.text' is of type
#      object but should be of type string."}
# The same request with {"type":"text","text":"-3"} returns 200.

# (3) tool_choice
-d '... "tool_choice":{"specific":{"function_names":["subtract"]}} ...'
# 422 {"message":"invalid type: parameter 'tool_choice' is of type object but
#      should be of type string."}
```

## Expected behavior

- Tool results serialize as `{"type":"text","text":"..."}` and tool round trips complete.
- `documents` serialize as `{id, data}` so `.context(..)` works.
- `ToolChoice::Specific` fails fast with a `CompletionError::RequestError` explaining
  that Cohere cannot force specific tools, instead of a 422 from the API.
- Model constants reflect Cohere's current catalogue; removed IDs are deprecated with
  a replacement named in the note.
- `provider_response_status()` returns the real status and `provider_response_body()`
  returns Cohere's error envelope, per the provider checklist in `CONTRIBUTING.md`.
- `response.usage` and the telemetry span report the same counters, sourced from
  `usage.tokens`, and populate `cached_input_tokens`.

## Additional context

There is currently no cassette suite for Cohere — `tests/providers/cohere/` holds only
four `#[ignore]`d live smoke tests, which is why none of this is caught in CI.

I have a working branch that fixes all six and adds nine live-recorded cassette
scenarios under `tests/providers/cohere/cassette/`, plus unit coverage. Happy to split
it as: model constants → cassette infrastructure → wire-format fixes → usage mapping,
per the "small and focused" guidance in `CONTRIBUTING.md`. Two things worth deciding
before I open PRs:

- The usage change is user-visible: telemetry token counts jump (~27 → ~556 input on a
  short prompt). Should that carry a `!` breaking marker?
- Adding `cached_tokens` to the public `Usage` struct breaks struct-literal construction
  downstream (it broke an existing test in-tree). Same question.

I'd like to pick this up if it's in scope.

rleisti:
# Fix Cohere provider: usage accounting, max_tokens, tool-call round-trips, and error surfacing

## Description

Fixes several correctness bugs in the Cohere provider (`crates/rig-core/src/providers/cohere/`) uncovered under RIG-1443, plus refreshes the model constant list against Cohere's current catalogue.

- **Token usage was inconsistent across surfaces.** `GetTokenUsage` read `usage.billed_units` while the non-streaming and streaming conversions read `usage.tokens`, so the same response reported different numbers depending on which path you asked. All three now map from `tokens` alone. `billed_units` excludes cached input and system-prompt overhead, so counts move up ~10-20x on the input side (e.g. 27 → 556 on a short prompt) — no code changes needed, but telemetry/cost dashboards keyed on Cohere counts will step up. `cached_input_tokens` is now populated from `usage.cached_tokens`, which was previously discarded.
- **`max_tokens` was silently dropped.** `CohereCompletionRequest` had no `max_tokens` field, so it never reached `/v2/chat`. It's now sent as a top-level parameter — agents that set `max_tokens` and saw no effect will now actually get cut off at that value.
- **Tool-result round-trips failed on the second turn.** `ToolResultContent` was missing `#[serde(tag = "type")]`, so it serialized as `{"Text":{"text":"-3"}}` instead of the tagged `{"type":"text","text":"-3"}` Cohere expects, causing a 422 on any multi-turn tool call. Fixed, with a corresponding note in MIGRATING.md for anyone who persisted the untagged JSON.
- **Transport errors hid status/body.** `CompletionModel::completion` boxed HTTP failures into `http_client::Error::Instance`, so `provider_response_status()`/`provider_response_body()` always returned `None`. Errors now propagate unwrapped as `InvalidStatusCodeWithMessage`.
- **`tool_choice` handling.** Replaced the passthrough of the full `ToolChoice` enum with a `CohereToolChoice` type that only accepts `Required`/`None` (matching what Cohere's API actually supports) and serializes as a bare string. `Auto` and `Specific` now fail fast client-side with an explanatory message instead of being silently mis-encoded.
- **Document conversion** now goes through `Document::from` so native documents serialize in Cohere's expected `{id, data}` shape (including extra fields via `additional_props`), rather than reusing `crate::completion::Document` directly.
- **Model constants refreshed:** added `COMMAND_A_*`, `COMMAND_R7B_12_2024`, `COMMAND_R_PLUS_08_2024`, `COMMAND_R_08_2024`, and `EMBED_V4`; deprecated the models Cohere retired on 2025-09-15 (`command-r-plus`, `command-r`, `command`, `command-light`, `command-light-nightly`) and `command-nightly`, each with a `#[deprecated]` note pointing at the replacement.

Also records the cassette for the previously-`#[ignore]`d `streaming_grammar::thinking_stream` Cohere test (no `COHERE_API_KEY` in CI), so it now replays from a fixture instead of being skipped.

Fixes #2241 
Fixes RIG-1443

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

Run `cargo test -p rig-core cohere::` and `cargo test --test cohere` to verify 

- [x] Unit tests for usage mapping, `ToolResultContent` tagging, `CohereToolChoice` serialization/validation, and document conversion (see `crates/rig-core/src/providers/cohere/completion.rs` test module)
- [x] Cassette-based integration tests under `tests/providers/cohere/cassette/` (agent, context, embeddings, errors, streaming, streaming_tools, tools) exercising the fixed request/response shapes against recorded fixtures
- [x] `streaming_grammar::thinking_stream` cassette recorded and de-`#[ignore]`d so it runs in CI without a live API key

## Checklist:

- [x] I have made corresponding changes to the documentation (MIGRATING.md)
- [x] New and existing unit tests pass locally with my changes (`cargo test -p rig-core cohere::` — 38 passed; `cargo test --test cohere` — 50 passed, 4 ignored live-API smoke tests; `cargo check --workspace --all-targets` clean)

## Notes

Four of the fixes here are breaking changes for anyone relying on the old (buggy) behavior; see the four new "Unreleased" entries in `MIGRATING.md` for details:
1. Cohere token counts now come from `tokens`, not `billed_units`
2. Cohere now honors `max_tokens`
3. Persisted `ToolResultContent` JSON from before this fix no longer round-trips (re-encode to the tagged form)
4. Cohere HTTP errors surface as `InvalidStatusCodeWithMessage` instead of `Instance`


gold-silver-copper:
Awesome, thank you!

PR #2264, opened 2026-08-07
fix(openai): forward additional_params in image generation request

Moeweb647252:
## Description

The OpenAI provider built the /images/generations body from model, prompt and size only, silently dropping ImageGenerationRequest.additional_params (e.g. quality, reference-image inputs). Merge them in after the response_format default so caller-supplied params take precedence.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

- [  ] I have added tests that prove the fix is effective

### DIFF
--- a/crates/rig-core/src/providers/openai/image_generation.rs
+++ b/crates/rig-core/src/providers/openai/image_generation.rs
@@ -103,6 +103,10 @@ where
             );
         }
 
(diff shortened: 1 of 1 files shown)


gold-silver-copper:
fixed in #2332

PR #2268, opened 2026-08-08
ci: split the PR gate from the full --all-features suite

gold-silver-copper:
Splits CI into a fast PR gate and a full `--all-features` lane (nightly + on-demand + every push to main), based on measured job timings rather than the assumed cost model.

## Phase 0 — measured baseline (median over the last 15 green runs)

| job | median | min | max |
|---|---|---|---|
| **stable / test** | **12.8m** | 10.6m | 32.9m |
| stable / doctest | 1.8m | 1.6m | 12.6m |
| stable / clippy | 1.7m | 1.3m | 8.4m |
| stable / doc | 1.4m | 1.3m | 8.1m |
| wasm checks (×6) | 0.6–0.8m | 0.5m | 1.2m |
| stable / fmt | 0.4m | 0.3m | 0.6m |

Critical path: `test` in 14 of 15 runs. The max column is cold-cache runs (lockfile/toolchain bumps).

Two premises of the original plan turned out stale on measurement:

1. **Caching already exists.** `actions-rust-lang/setup-rust-toolchain@v1` bundles `Swatinem/rust-cache` by default — that is why clippy finishes `--all-features --all-targets` in ~64s of lint time. There was no Phase-1 caching win left to take; the whole problem lives inside the `test` job.
2. **Execution is now a real share, not just compilation.** The final `--all-features` step in a representative warm run: ~2m compile + ~4.4m execution. The execution long pole is `portable_tool_facade_is_feature_additive` (123s of nested `cargo check`), then the Docker integration tests (mongodb 50s+44s, neo4j 46s, postgres 22s, lancedb 14s×2, qdrant 12s) and eight ~20s `cassette_safety` scans. (The lancedb tests **pass** on main — 14s each — contrary to the plan's premise; nothing needed fixing or dropping.)

Representative `test` job step timings (warm cache): setup+cache-restore 79s, protoc 29s, default-feature compile guard 49s, four targeted contract steps 137s, final `--all-features` run 391s.

## What changed

**PR gate (`ci.yaml`)** — `fmt`, the wasm checks, `check-rmcp-native-only`, `clippy` (still `--all-features --all-targets`), `doctest`, and `doc` are untouched. The `test` job splits in two parallel jobs:

- `test`: the default-feature compile guard (unchanged, still the only check of the pure default-feature test graph) + the workspace sweep as `cargo nextest run --features bedrock --retries 2`. `bedrock` rides along because the bedrock wire-conformance suite compiles into the facade's `bedrock` test binary and `tests/core/streaming_conformance_registry.rs` pins this step as the one executing it; rig-bedrock is already a workspace default member so the marginal compile cost is small. Provider API keys stay on this step (same-repo PRs keep exactly the pre-split live-test behaviour; fork PRs never had secrets).
- `test-guards`: the four existing contract tripwire steps, verbatim (agent unit tests, telemetry completion-parent contract with its deliberate no-retries, macro hygiene, out-of-facade conformance + structural guards), **plus one new step** running the full rig-core + rig-agent lib test sets under `--all-features` — without it, 81 fast feature-gated unit tests (rmcp tool registry, pdf/epub loaders, audio) would have demoted to nightly-only coverage. It reuses the telemetry step's compile artifacts.
- protoc is no longer installed in these two jobs: it exists only for rig-lancedb (not in either job's graph; rig-gemini-grpc vendors its own via `protoc-bin-vendored`). The three `--all-features` jobs keep it.

**Full lane (`nightly.yaml`, new)** — the exact `cargo nextest run --all-features --retries 2` step that left the gate, with protoc and the provider keys. Triggers: cron `0 6 * * *`, `workflow_dispatch` (a contributor touching a companion crate can run it on demand), and `workflow_call`.

**`cd.yaml`** — now calls `nightly.yaml` alongside `ci.yaml`, and `release-plz` needs both. Every push to main therefore still runs the full suite before any release; the split only changes *when* PRs pay for it, and reviewers keep the pre-split guarantee that nothing is released unverified. (This is why nightly.yaml needs no `merge_group` trigger: no merge-queue run has ever occurred on this repo, and the push-to-main path already covers landing.)

**Container tags** — `qdrant:latest` → `v1.19.0`, `neo4j:latest` → `5.26.29`, `mongodb-atlas-local:latest` → `8.0.5` (both call sites), matching the already-pinned `pgvector:pg17` / `scylla:5.4`. Floating tags defeat layer caching and let a rerun silently test a different database version.

## Test accounting (nothing deleted, nothing lost)

Diffed `cargo nextest list` between the gate's union (sweep + every test-guards step) and `--all-features`:

- **Gate ∪ nightly covers every test on main; zero tests run in neither.** (Also zero gate tests absent from the all-features set.)
- **Exactly 20 tests now execute only in the full lane** (nightly / push-to-main / on-demand): the 18 `tests/integrations.rs` tests (lancedb ×2, mongodb ×2, neo4j, postgres, qdrant, scylladb mock-setup, sqlite ×2, vectorize ×8), `portable_tool_facade_is_feature_additive`, and the two `image`-feature generation smoke tests (gemini nano-banana, xai).
- `--all-features` **compile** coverage is still on every PR: clippy lints `--all-features --all-targets`, doctest and doc build all features. A nightly failure can only be a runtime regression, never a stale compile break.

## Deviations from the plan, with reasons

- **Phase 1 (add caching) is a no-op** — caching was already in place (see above), which the plan itself anticipated ("one of the phases may make the others unnecessary").
- **`doctest`, `doc`, and full-fat `clippy` stay on the PR gate** rather than moving to nightly: at 1.4–1.8m each they are far off the critical path, moving them would save zero wall-clock, and they are what keeps all-features compile breaks out of main. This is the plan's own "prefer keeping clippy at `--all-features` if cheap" logic, applied to all three.
- **No containers were cut from the full lane**: they no longer cost any PR wall-clock, they run in parallel inside the sweep (biggest single contribution ~50s), the lancedb tests pass, and these are the only runtime coverage the companion crates' `VectorStoreIndex` implementations have. Cutting them would be all risk, no measurable win.

## Expected gate shape after the split

The gate's critical path becomes `test` ≈ setup + default+bedrock compile + ~2m execution, with `test-guards` similar and parallel — roughly **5–6m against the 12.8m baseline** on warm cache. The first run on this PR is a cold cache for the two new job cache keys; the number to judge is the steady state after merge. I will post the measured after-table as a comment once a few runs accumulate.

## Verification

- [ ] PR gate green with the new shape (this PR)
- [ ] Full-suite workflow proven green before merge via a dry-run branch (`ci/nightly-dryrun` — temporary `push` trigger, since `workflow_dispatch` cannot target a workflow that does not exist on the default branch yet)
- [x] `cargo nextest list` diff: every test accounted for in gate ∪ full lane

### DIFF
--- a/.github/scripts/check-toolchain-pin.sh
+++ b/.github/scripts/check-toolchain-pin.sh
@@ -1,0 +1,45 @@
+#!/usr/bin/env bash
+# `rust-toolchain.toml` is what cargo itself obeys, but every workflow also
+# carries a `RUST_VERSION` env used as the `setup-rust-toolchain` input. That
+# is four copies of one number, and a partial bump is silent: the gate goes
+# green on the new toolchain while a workflow that was missed — nightly.yaml
+# is also cd.yaml's pre-release gate — keeps compiling on the old one, so a
+# failure there gets misread as a runtime regression rather than a stale pin.
+#
+# Assert the copies agree. This is a text check on purpose: it must fail on a
+# workflow nobody remembered to update, which a parser keyed off the workflows
+# that *do* declare the variable would not do.
+set -euo pipefail
+
+cd "$(dirname "$0")/../.."
+
+channel=$(grep -E '^\s*channel\s*=' rust-toolchain.toml | head -1 | sed -E 's/.*"([^"]+)".*/\1/')
+if [ -z "$channel" ]; then
+  echo "::error::could not read [toolchain] channel from rust-toolchain.toml"
+  exit 1
+fi
+
+status=0
+found=0
+for workflow in .github/workflows/*.yaml .github/workflows/*.yml; do
+  [ -e "$workflow" ] || continue
+  # Only workflows that declare the variable are checked; a workflow with no
+  # `RUST_VERSION` takes its toolchain from rust-toolchain.toml already.
+  while IFS= read -r pinned; do
+    found=1
+    if [ "$pinned" != "$channel" ]; then
+      echo "::error file=$workflow::RUST_VERSION is $pinned but rust-toolchain.toml pins $channel"
+      status=1
+    fi
+  done < <(grep -E '^\s*RUST_VERSION:' "$workflow" | sed -E 's/.*RUST_VERSION:[[:space:]]*//' | tr -d '"'"'"'')
+done
+
+if [ "$found" -eq 0 ]; then
+  echo "::error::no workflow declares RUST_VERSION — this guard is checking nothing, so either restore the pins or delete the guard"
+  exit 1
+fi
+
+if [ "$status" -eq 0 ]; then
+  echo "ok: every workflow RUST_VERSION matches rust-toolchain.toml ($channel)"
+fi
+exit "$status"
--- a/.github/scripts/check-toolchain-pin.sh
+++ b/.github/scripts/check-toolchain-pin.sh
@@ -43,6 +43,19 @@ for workflow in .github/workflows/*.yaml .github/workflows/*.yml; do
     fi
   done < <(grep -E '^\s*RUST_VERSION:' "$workflow" \
     | sed -E 's/\r$//; s/^[[:space:]]*RUST_VERSION:[[:space:]]*//; s/[[:space:]]+#.*$//; s/[[:space:]]+$//; s/^"([^"]*)"$/\1/; s/^'"'"'([^'"'"']*)'"'"'$/\1/')
+
+  # Also catch the bypass this env-var check cannot see: a job that pins a
+  # literal version directly in `setup-rust-toolchain`'s `with:` block
+  # (`toolchain: 1.93.0`) declares no RUST_VERSION and would pass vacuously
+  # while silently building on a stale toolchain. `${{ env.RUST_VERSION }}`
+  # references are skipped — they resolve to a value already checked above.
+  while IFS= read -r inline; do
+    if [ "$inline" != "$channel" ]; then
+      echo "::error file=$workflow::inline toolchain pin is $inline but rust-toolchain.toml pins $channel"
+      status=1
+    fi
+  done < <(grep -E '^\s*toolchain:' "$workflow" | grep -vF '${{' \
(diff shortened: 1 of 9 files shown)


gold-silver-copper:
## Measured results (both verification runs green)

**PR gate, new shape** — run [31274290049](https://github.com/0xPlaygrounds/rig/actions/runs/31274290049), first attempt cold (new cache keys), rerun warm:

| job | baseline median | cold (attempt 1) | warm (attempt 2) |
|---|---|---|---|
| **stable / test** (critical path) | **12.8m** | 10.4m | **8.1m** |
| stable / test guards (new, parallel) | — | 6.0m | 3.5m |
| stable / clippy | 1.7m | 2.0m | 1.7m |
| stable / doctest | 1.8m | 1.9m | 1.8m |
| stable / doc | 1.4m | 1.7m | 1.3m |
| fmt / wasm checks (×7) | 0.4–0.8m | unchanged | unchanged |

**Critical path: 12.8m → 8.1m warm (−37%), and −4.7m ≈ −45% of the test job itself.** Warm `test` breakdown: 71s cache restore, 49s default-feature compile guard, then the sweep at 356s ≈ 246s compile + 110s execution (3442 tests, 47 binaries, all passed). The residual floor is workspace-crate compilation, which rust-cache deliberately never caches — cutting that needs sccache or similar, out of scope here.

**Full lane dry-run** — run [31274258218](https://github.com/0xPlaygrounds/rig/actions/runs/31274258218), green before merge as promised: 3556/3556 passed in 23.6m (fully cold cache; the job had never existed). All 17 Docker-backed integration tests passed against the **pinned** images (`qdrant:v1.19.0` 10.7s, `postgres` 16.5s, `lancedb` 22–23s, `mongodb-atlas-local:8.0.5` 37.5s+46.6s, `neo4j:5.26.29` 49.5s). The temporary `ci/nightly-dryrun` branch used to trigger it (workflow_dispatch cannot target a workflow absent from the default branch) has been deleted.

Remaining execution long-poles inside the gate sweep, for future reference: the `identity_leak` trybuild suite (23s) and the eight ~16s `cassette_safety` secret-scans.

Issue #2291, opened 2026-08-12
feat: Support Mistral document_url and image_url content parts


ac-machache:
- [x] I have looked for existing issues (including closed) about this

## Feature Request

Support Mistral's multimodal content parts — `document_url` and `image_url` —
so documents and images can be attached to a Mistral completion. This also fixes
the silent drop reported in #2290.

### Motivation

`UserContent` already models `Document` and `Image`, and Mistral accepts them as
base64 bytes in a content array. rig can't send either: `finalize_request_body`
flattens every content array to a plain string and keeps only `text` parts, so the
attachment is discarded before the request goes out.

Document Q&A and vision are unreachable through rig on Mistral today, even though
the provider supports both and rig already has the types for them.

### Proposal

In `MistralExt::finalize_request_body`, flatten only when every part is text.
Otherwise emit Mistral's parts — `text`, `image_url`, `document_url` — carrying the
base64 data rig already holds in `DocumentSourceKind::Base64`.

Drawback: Mistral's request shaping diverges further from the plain OpenAI shape it
currently borrows, so the two need to stay in step as OpenAI's content handling
changes.

### Alternatives

- **Return a `ConversionError` instead of dropping**, as Anthropic does for audio.
  Fixes the silent loss but leaves attachments unsupported.
- **Use a different provider for file turns.** Gemini handles them natively, but
  that's two providers to serve one content type and nothing improves in rig.

I'd be happy to implement this if it isn't already planned or assigned — the same
code path covers #2290 too.

gold-silver-copper:
Implemented on `main` — closing.

This landed with #2331 (`fix(mistral): four provider bugs found by live cassette recording`), alongside the silent-drop bug #2290 that was closed at the time; this feature issue was just left open.

The proposal was followed almost exactly. In `crates/rig-core/src/providers/mistral/completion.rs`:

- `normalize_request_content` flattens **only** when `parts.iter().all(is_text_part)`; anything else keeps the array and each part is rewritten by `into_mistral_chunk`.
- `into_mistral_chunk` emits Mistral's own chunk names — `text`, `image_url`, `input_audio`, `document_url`, `file` — rebuilding each payload rather than forwarding it, since every Mistral chunk forbids unknown fields.
- Inline document bytes become `document_url` carrying the base64 `data:` URI, with the filename in Mistral's own `document_name` field; an uploaded-file reference becomes the `file` chunk with `file_id` at the top level.
[...]

PR #2293, opened 2026-08-12
fix(mistral): preserve supported non-text content

### LINKED ISSUE #2290: bug: Mistral provider silently drops non-text content parts

## Bug Report

The Mistral provider silently discards every non-text content part. Attaching a
document, image or audio returns `Ok` with a normal completion — the part never
reaches the API, and nothing in the result says so.

`MistralExt::finalize_request_body` (`providers/mistral/client.rs`) calls:

```rust
flatten_text_content_parts(content, "", false)

false = flatten even when not all parts are text, and the helper keeps only
parts with a text/refusal key. file, image_url and input_audio parts are
removed there.

Reproduction

let agent = mistral::Client::new(&api_key)?
    .agent(mistral::MISTRAL_SMALL)
    .build();

// base64 of any small PDF
let msg = Message::document(pdf_base64, Some(DocumentMediaType::PDF));

// Succeeds, and answers as though nothing was attached.
println!("{}", agent.prompt(msg).await?);

Same with Message::image_base64(..) and Message::audio(..).

Expected behavior

Either translate the parts into Mistral's schema, or return a ConversionError
as Anthropic does ("Audio is not supported in Anthropic"). Not: succeed while
dropping the content.

Additional context

rig-core 0.41.0.

css521:
Closes #2290

## Summary

- classify serialized message content before Mistral text flattening
- preserve Mistral `ImageURLChunk` objects and convert Rig's OpenAI-shaped audio payload to the official Mistral `input_audio: string` form
- reject file, video, malformed, and unknown content before sending instead of silently dropping it
- cover blocking, streaming, supported multimodal, and fail-closed request paths
- pin the complete image wire shape and add a positive streaming control so future over-rejection is detected

## Tests

- `cargo fmt --all -- --check`
- `cargo test -p rig-core mistral` (23 passed)
- `cargo test -p rig-core --lib` (1262 passed, 3 ignored)

A full `cargo test -p rig-core` also passed the library and driver suites locally; its existing `identity_leak` trybuild snapshots differed only in current Rust diagnostic wording and the expanded local Cargo mirror path.

### DIFF
--- a/crates/rig-core/src/providers/mistral/client.rs
+++ b/crates/rig-core/src/providers/mistral/client.rs
@@ -68,11 +68,12 @@ impl crate::providers::openai::completion::OpenAICompatibleProvider for MistralE
                 let is_assistant =
                     message.get("role").and_then(serde_json::Value::as_str) == Some("assistant");
 
-                // Mistral takes message `content` as a plain string.
+                // Mistral takes text-only message `content` as a plain string
+                // and carries its own image/audio chunks as an array; content
+                // types it has no chunk for fail the request here rather than
+                // reaching the API with the content removed.
                 if let Some(content) = message.get_mut("content") {
-                    crate::providers::openai::completion::flatten_text_content_parts(
-                        content, "", false,
-                    );
+                    super::completion::normalize_request_content(content)?;
                 }
 
                 if is_assistant {
--- a/crates/rig-core/src/providers/mistral/completion.rs
+++ b/crates/rig-core/src/providers/mistral/completion.rs
@@ -62,6 +62,110 @@ where
         .unwrap_or_default())
 }
 
+/// Content chunk type Mistral carries for images, used by the Pixtral models.
+const IMAGE_CHUNK_TYPE: &str = "image_url";
+/// Content chunk type Mistral carries for prompt audio, used by the
+/// audio-input models that report `prompt_audio_seconds` usage.
+const AUDIO_CHUNK_TYPE: &str = "input_audio";
+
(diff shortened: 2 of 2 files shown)


gold-silver-copper:
Hi @css521 — I've opened #2331, which supersedes this PR for #2290. Flagging it here so the two don't get reviewed in parallel.

Your diagnosis was right, and #2331 keeps both it and your fail-closed instinct. Three things differ:

1. **Documents are mapped rather than rejected.** Mistral reads inline base64 PDFs through its `document_url` chunk — verified end to end against the live API, so `file` parts now become `document_url` (or Mistral's own `{"type":"file","file_id":…}` for an uploaded id) instead of erroring. That also implements #2291, which failing closed would have left unreachable.
2. **The audio note is corrected.** This PR sends the same bare string yours does, but not for the stated reason: I probed the live API and the OpenAI-shaped `{data, format}` object *is* accepted (the server flattens it and ignores `format`), returning 200 with identical `audio_tokens`. The bare string is still the right thing to send — it's the form Mistral's published schema documents — so the destination is unchanged, only the justification.
3. **It's backed by recorded live traffic**: 26 cassettes, 23 of which fail on `main` and pass with the fix, rather than unit tests alone.

Thanks for finding and characterising this one — you're credited in #2331's description.

css521:
Thanks for the heads-up and for carrying this forward. The document mapping and audio clarification make sense, and I appreciate the live-traffic coverage and co-author credit. Since #2331 has now merged and supersedes this PR, I’ll close #2293.

Issue #2079, opened 2026-07-09
refactor(providers): shared `fetch_model_list` helper for the ten hand-rolled model listers (+ fill `ModelListing = Nothing` gaps)


gold-silver-copper:
# Summary

The `ModelLister` implementation — a `ListModelsResponse { data: Vec<Entry> }` struct, a GET, a status check, `parse_error_with_context` — is copy-pasted across ten providers. A shared `fetch_model_list` helper in `providers/internal` collapses each to ~15 lines, and turns the providers that currently declare `ModelListing = Nothing` despite having a `/models` endpoint into one-line additions.

References against `d03689a1`.

# Current state

Hand-rolled copies of the same shape:

- `providers/openai/model_listing.rs:11`
- `providers/mistral/model_listing.rs:11`
- `providers/anthropic/model_listing.rs:11` (paginated variant)
- `providers/openrouter/model_listing.rs:11`
- `providers/gemini/model_listing.rs:15` (paginated, two listers)
- `providers/deepseek.rs:412`
- `providers/ollama.rs:830`
- `providers/xiaomimimo.rs:283`
- `providers/copilot/mod.rs:1448`

Related inconsistencies:

- groq, moonshot, zai, minimax, mira declare `ModelListing = Nothing` although their upstream APIs expose `/models` — capability gaps that become trivial once the helper exists.
- mira is worse: it ships a bespoke **off-trait** `Client::list_models() -> Vec<String>` with its own `MiraError` enum (`providers/mira.rs:219-245`, `:136`) — the only provider with a listing API outside the `ModelLister` trait. It should implement the trait via the helper and drop the one-off surface.

# Proposal

1. Add `providers/internal::fetch_model_list(client, path, provider_name, map_entry)` (plus a paginated variant for anthropic/gemini).
2. Collapse the ten existing listers onto it.
3. Follow-up (can be the same PR or a second one): add `ModelLister` impls for groq/moonshot/zai/minimax/mira; migrate mira off its bespoke `list_models`.

# Acceptance criteria

- One shared fetch path; each provider lister is a thin mapping; existing model-listing tests pass; no provider exposes an off-trait listing API.

Lstarsky0:
[...]

PR #2321, opened 2026-08-14
feat(providers): implement ModelLister for groq, moonshot, and minimax

Lstarsky0:
Refs #2079.

## Description

Steps 1 and 2 of #2079 have landed — `providers/internal/model_listing.rs` holds the shared `list_models` path and the `impl_model_lister!` macro, and mira is migrated off its bespoke off-trait `list_models`. This is the leftover from step 3: the `ModelListing = Nothing` capability gaps.

Groq, Moonshot and MiniMax each serve an OpenAI-style `GET <base>/models`, but all three declared `ModelListing = Nothing`, so `client.list_models()` simply didn't exist on them. Each now declares `model_listing` and defines its lister through the existing macro:

```rust
crate::providers::internal::model_listing::impl_model_lister!(
    /// [`ModelLister`](crate::client::ModelLister) implementation for the Groq
    /// API (`GET /models`), the same endpoint [`GroqExt::VERIFY_PATH`] already
    /// uses to verify credentials.
    GroqModelLister,
    Client<H>,
    crate::providers::internal::model_listing::ListModelEntry,
    "Groq",
    "/models"
);
```

Two hunks per provider — a capability line and a macro call. The request, status triage, envelope decoding and error context all come from the shared path, so there is no new hand-rolled listing code. The path is relative to the configured base URL, so Moonshot and MiniMax listings follow the client between their global and China hosts instead of pinning a host.

## Type of change

- [x] New feature

## What I can and can't vouch for on the endpoints

Groq is the solid one and it is provable in-tree: `GroqExt::VERIFY_PATH` is already `"/models"` (`groq.rs:37`), so this provider was hitting that exact path before this PR to verify credentials. With `GROQ_API_BASE_URL` at `https://api.groq.com/openai/v1`, the lister resolves `https://api.groq.com/openai/v1/models`, which is what the test asserts.

Moonshot and MiniMax are the OpenAI-compatible `/models` convention applied to base URLs already declared in this repo:

| provider | global | China |
| --- | --- | --- |
| Moonshot | `https://api.moonshot.ai/v1` | `https://api.moonshot.cn/v1` |
| MiniMax | `https://api.minimax.io/v1` | `https://api.minimaxi.com/v1` |

**I made no live API calls against any of the three**, so nothing here is verified against a running server — the tests prove the URL the client resolves and the decoding, not that the remote answers. If you want a live check before merging, that needs keys I don't have.

All three return extra per-entry fields beyond the four `ListModelEntry` reads. `ListModelEntry` doesn't set `deny_unknown_fields`, so they're ignored; the Moonshot fixture includes `context_length` to pin that.

## Testing

Six unit tests, no API keys needed. They drive the real client through `RecordingHttpClient` and assert the resolved URL rather than restating the constant, plus the mapped `Model` fields, the base-URL switch for the two dual-region providers, and that a 401 keeps provider/path/status/body in the error.

On the pinned toolchain (rustc 1.94.0), using CI's own commands, run twice end to end with identical results:

| Command | Exit | |
| --- | --- | --- |
| `cargo fmt -- --check` | 0 | |
| `cargo clippy --all-features --all-targets` | 0 | no warnings |
| `cargo nextest run --locked -p rig-core -p rig-agent --all-features --retries 2 -E 'not binary(macro_hygiene)'` | 0 | 1956 passed, 5 skipped |

The six new tests by name:

```
rig-core providers::groq::tests::list_models_uses_the_models_endpoint
rig-core providers::groq::tests::list_models_preserves_api_error_context
rig-core providers::moonshot::tests::list_models_uses_the_models_endpoint
rig-core providers::moonshot::tests::list_models_follows_the_configured_base_url
rig-core providers::minimax::tests::list_models_uses_the_models_endpoint
rig-core providers::minimax::tests::list_models_follows_the_configured_base_url
```

I also checked they can fail: changing Groq's `"/models"` to `"/modelz"` fails both Groq tests with

```
left:  "https://api.groq.com/openai/v1/modelz"
right: "https://api.groq.com/openai/v1/models"
```

which is the point of asserting the resolved URL rather than the constant.

## Checklist

- [x] I have made corresponding changes to the documentation — docstrings on each lister; no guide changes needed
- [x] I have added tests that prove my feature works

## Notes

**Z.ai is deliberately left out.** #2079 lists it, but one lister on `ZAiExt` would resolve `/models` against whichever base the client was built with, and this repo declares two:

```rust
/// General-purpose OpenAI-compatible base URL.
pub const GENERAL_API_BASE_URL: &str = "https://api.z.ai/api/paas/v4";
/// Coding-focused OpenAI-compatible base URL.
pub const CODING_API_BASE_URL: &str = "https://api.z.ai/api/coding/paas/v4";
```

Those are two different product surfaces by rig's own description, so `list_models()` would mean something different depending on configuration, and I can't confirm what either returns without live calls. Say the word and I'll add it — it should probably carry a doc warning if so. (I cited a Z.ai SDK issue for this on #2079 and have since retracted that: the reference no longer resolves.)

**MiniMax's Anthropic dialect** keeps no lister — `/anthropic/v1/models` returns a different shape that `ListModelEntry` wouldn't decode correctly. Noted in the macro docstring so it doesn't read as an oversight.

**Left as `Refs` rather than `Closes`** because #2079 names z.ai and this doesn't cover it. Flip it if you'd rather the issue close on merge.

Last thing: if #2228 is expected to land first I'm happy to hold or retarget, since it removes `client/model_listing.rs` and this would need redoing on the new shape. Per provider the production surface is one capability line plus one macro call, so redoing it is cheap either way — the bulk of the diff is the tests.

### DIFF
--- a/crates/rig-core/src/providers/groq.rs
+++ b/crates/rig-core/src/providers/groq.rs
@@ -84,6 +84,7 @@ client::impl_capabilities!(
     GroqExt,
     completion = CompletionModel<H>,
     transcription = TranscriptionModel<H>,
+    model_listing = GroqModelLister<H>,
 );
 
 impl DebugExt for GroqExt {}
@@ -106,6 +107,17 @@ pub type CompletionModel<H = reqwest::Client> =
 /// final item of the stream returned by `CompletionModel::raw_stream`. Shared
 /// with the OpenAI Chat Completions path, usage payload included.
 pub type StreamingCompletionResponse = openai::StreamingCompletionResponse;
+
+crate::providers::internal::model_listing::impl_model_lister!(
+    /// [`ModelLister`](crate::client::ModelLister) implementation for the Groq
+    /// API (`GET /models`), the same endpoint [`GroqExt::VERIFY_PATH`] already
+    /// uses to verify credentials.
+    GroqModelLister,
+    Client<H>,
+    crate::providers::internal::model_listing::ListModelEntry,
+    "Groq",
+    "/models"
+);
 
 client::impl_provider_client!(Client, input = String, api_key_env = "GROQ_API_KEY");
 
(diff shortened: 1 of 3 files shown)


Lstarsky0:
Superseded by #2339 — groq, moonshot and minimax all have `impl_model_lister!` on `main` now. Closing.

gold-silver-copper:
@Lstarsky0 Sorry about that, I should've given you coauthor credit there. It just kind of evolved on it's own and I forgot to :(

Lstarsky0:
No worries at all — yours did more than mine anyway, with the shared pagination loop and the termination guard.

Happy to take some of the providers that still have no listing (`together`, `xai`, `zai`, `hyperbolic`, and a few others) if that's useful. Just say which ones you're not already on and I'll pick those up.

PR #2323, opened 2026-08-14
fix(gemini): stop injecting a hardcoded maxOutputTokens=4096 into structured-output and image requests

### LINKED ISSUE #2322: bug: Gemini injects a hardcoded maxOutputTokens=4096 and temperature=1.0 into structured-output and image requests

## Bug Report

`GenerationConfig::default()` for the Gemini provider hardcodes `max_output_tokens: Some(4096)` and `temperature: Some(1.0)`:

https://github.com/0xPlaygrounds/rig/blob/d4e4c988211b136778c174689f1bd85e2e6a9fe6/crates/rig-core/src/providers/gemini/completion.rs#L1730-L1752

Two call sites still seed a request from that `Default`, so both silently ship a 4096-token cap the caller never asked for and pin `temperature` to `1.0`:

1. **Structured output** — the `output_schema` arm of `create_request_body`:
   https://github.com/0xPlaygrounds/rig/blob/d4e4c988211b136778c174689f1bd85e2e6a9fe6/crates/rig-core/src/providers/gemini/completion.rs#L217-L222
2. **Image generation** — `..Default::default()` in the request literal:
   https://github.com/0xPlaygrounds/rig/blob/d4e4c988211b136778c174689f1bd85e2e6a9fe6/crates/rig-core/src/providers/gemini/image_generation.rs#L113-L120

#2283 fixed exactly this hazard for the plain `temperature` / `max_tokens` path, and the comment it left behind spells out why `Default::default()` must not be used as a seed — but the `output_schema` arm immediately above it was not updated:

https://github.com/0xPlaygrounds/rig/blob/d4e4c988211b136778c174689f1bd85e2e6a9fe6/crates/rig-core/src/providers/gemini/completion.rs#L224-L234

Every field of `GenerationConfig` is `skip_serializing_if = "Option::is_none"`, so leaving these two fields unset keeps them off the wire entirely and lets Gemini apply the model's own output limit — which is what the caller expects when they never set `max_tokens`.

Note this only bites callers who rely on the model default: an explicit `max_tokens` is applied afterwards at
[completion.rs#L246-L248](https://github.com/0xPlaygrounds/rig/blob/d4e4c988211b136778c174689f1bd85e2e6a9fe6/crates/rig-core/src/providers/gemini/completion.rs#L246-L248) and overwrites the injected 4096.

### Secondary: streaming reports a truncated turn as a successful, empty completion

This is what hides the bug above. The two paths disagree:

- **Non-streaming** raises a clear `ResponseError` naming the finish reason when a candidate has no content ([completion.rs#L438-L454](https://github.com/0xPlaygrounds/rig/blob/d4e4c988211b136778c174689f1bd85e2e6a9fe6/crates/rig-core/src/providers/gemini/completion.rs#L438-L454)), and additionally guards emptiness via `require_non_empty_response` ([L516](https://github.com/0xPlaygrounds/rig/blob/d4e4c988211b136778c174689f1bd85e2e6a9fe6/crates/rig-core/src/providers/gemini/completion.rs#L516)).
- **Streaming** logs the same condition at `debug!` and then emits `FinalResponse` anyway, so the stream completes normally with zero assistant content:
  https://github.com/0xPlaygrounds/rig/blob/d4e4c988211b136778c174689f1bd85e2e6a9fe6/crates/rig-core/src/providers/gemini/streaming.rs#L286-L307

`require_non_empty_response` is applied only on non-streaming paths. `finish_reason: Length` *is* propagated on the final response, so this is detectable — but nothing enforces it, so hitting `maxOutputTokens` is indistinguishable from a normal empty completion unless the caller explicitly inspects `finish_reason`.

## Reproduction

Deterministic, no API key needed — `create_request_body` is `pub(crate)`, so this runs as an in-crate test in `providers::gemini::completion`:

```rust
let request = CompletionRequest {
    model: None,
    preamble: None,
    chat_history: vec!["Hello".into()],
    documents: vec![],
    tools: vec![],
    temperature: None,
    max_tokens: None,          // caller relies on the model's own limit
    tool_choice: None,
    additional_params: None,
    output_schema: Some(
        schemars::Schema::try_from(serde_json::json!({
            "type": "object",
            "properties": {"a": {"type": "string"}}
        }))
        .unwrap(),
    ),
    record_telemetry_content: false,
};

let body = create_request_body(request).unwrap();
println!("{}", serde_json::to_string(&body.generation_config).unwrap());
```

Actual output on `main` (`d4e4c98`):

```json
{"responseMimeType":"application/json",
 "responseJsonSchema":{"properties":{"a":{"type":"string"}},"type":"object"},
 "maxOutputTokens":4096,
 "temperature":1.0}
```

The same request with `output_schema: None` correctly produces `null` (no `generationConfig` on the wire), which is the post-#2283 behaviour.

## Expected behavior

A request that does not set `max_tokens` or `temperature` should not send `maxOutputTokens` or `temperature` at all, regardless of whether an `output_schema` is present — Gemini should apply the model's own output-token limit.

Secondarily, a streaming turn terminated by `MAX_TOKENS` with no content should not surface as a successful empty completion.

## Additional context

Suggested fix — changing the `Default` itself resolves all call sites at once:

1. Set `temperature: None` and `max_output_tokens: None` in `impl Default for GenerationConfig`. That makes `::default()` safe as a seed, fixes both the `output_schema` and image-generation call sites, and lets the existing `skip_serializing_if` attributes do their job. The bespoke struct literal added by #2283 at [L236-L240](https://github.com/0xPlaygrounds/rig/blob/d4e4c988211b136778c174689f1bd85e2e6a9fe6/crates/rig-core/src/providers/gemini/completion.rs#L236-L240) can then collapse back to `GenerationConfig::default`.
2. Separately decide whether the streaming path should error (or at minimum `warn!`) on a content-less terminal chunk, matching the non-streaming path.

`GenerationConfig` is public, so (1) is a user-visible change to its `Default` and likely wants a changelog entry even though it is a bug fix.

Scope check — these are **not** affected:
- `interactions_api::GenerationConfig` derives `Default`, so all fields are `None`.
- `rig-vertexai` builds its own `vertexai::model::GenerationConfig::new()` and treats the typed `max_tokens` as authoritative.
- Deserializing a partial `generationConfig` from `additional_params` is fine — serde yields `None` for absent fields rather than going through `Default`.

Verified present on `main` at `d4e4c988211b136778c174689f1bd85e2e6a9fe6`.

gold-silver-copper:
Closes #2322.

## The bug

`gemini::completion::GenerationConfig`'s `Default` hardcoded `temperature: Some(1.0)` and `max_output_tokens: Some(4096)`. Two request paths seeded themselves from that `Default`, so both silently shipped a 4096-token cap and a pinned temperature that no caller asked for:

- **Native structured output** — the `output_schema` arm of `create_request_body`, i.e. *every* `output_schema` turn.
- **Image generation** — the `..Default::default()` request literal.

A caller who configured a 16k budget got 4096. Only callers relying on the model's own default were affected: an explicit `max_tokens` is applied afterwards and overwrote the injected value, which is why the plain-agent fix in #2283 did not surface this.

#2283 fixed the same hazard for the `temperature`/`max_tokens` arm and left a comment spelling out why `Default::default()` must not be used as a seed — but the `output_schema` arm directly above it kept doing exactly that.

## The fix

Rather than clear the two fields at a third call site, the footgun is removed at the root. `Default` is now **derived**, so it is all-`None`, and every field is already `skip_serializing_if = "Option::is_none"`. A default config therefore puts *nothing* on the wire and Gemini applies each model's own documented limit. The bespoke struct literal from #2283 collapses back to `GenerationConfig::default()`, and a doc comment on the type says not to reintroduce non-`None` defaults, with the reason.

This fixes both call sites at once, and any future one.

## Test matrix

Live-recorded against Gemini. Each asserts on the **request** boundary per `tests/README.md` — the recorded body is the assertion. The new `assert_recorded_sampling_fields` helper parses the cassette YAML and indexes into the decoded JSON rather than substring-matching the raw file, so a field name occurring incidentally in a response body or a JSON schema property cannot make an absence assertion silently pass or fail.

| scenario | schema | `max_tokens` | `temperature` | recorded `generationConfig` |
| --- | --- | --- | --- | --- |
| `agent_max_tokens_without_additional_params` (C15, existing) | no | 512 | unset | `maxOutputTokens` only |
| `structured_output_without_max_tokens` **(new)** | yes | unset | unset | neither field |
| `structured_output_with_max_tokens` **(new)** | yes | 16384 | unset | `maxOutputTokens` only |
| `temperature_without_max_tokens` **(new)** | no | unset | 0.0 | `temperature` only |
| `nano_banana_image_generation_smoke` (existing, assertion added) | n/a | n/a | n/a | neither field |

The 16384 case pins the reported scenario directly: a budget above the old hardcoded cap now reaches Gemini intact. C15 and the new temperature-only case cover both halves of the shared seed, so a reintroduced non-`None` default is caught whichever field it lands on.

## Cassette churn

18 existing Gemini cassettes were re-recorded live:

- **2 changed because of the bug** — `structured_output_smoke` and `nano_banana_image_generation_smoke` were recording `"maxOutputTokens":4096,"temperature":1.0` on requests that never set either field. These diffs *are* the bug, captured in the committed corpus.
- **16 changed as a consequence** — their tests pass `GenerationConfig::default()` through `additional_params`, which now serializes to `{}` instead of the two injected fields. Test code is unchanged; I did not redesign those tests as part of this fix.

Request bodies lose only those two fields. Response bodies differ because they are fresh recordings. **No test assertion was weakened or removed to make anything pass.**

## Verification

```
cargo test -p rig --all-features --test gemini -- --test-threads=1
    181 passed; 0 failed; 1 ignored

cargo test -p rig --all-features --test core -- --test-threads=1
    188 passed; 0 failed

cargo test -p rig-core --lib          1304 passed; 0 failed
cargo test -p rig-vertexai              55 passed; 0 failed
cargo fmt --all -- --check              clean
cargo clippy -p rig-core --all-features clean
```

`rig-vertexai` consumes this type as `GeminiGenerationConfig` but builds its own `vertexai::model::GenerationConfig::new()` and treats the typed `max_tokens` as authoritative, so it is unaffected — confirmed by its suite. The Interactions API's `GenerationConfig` already derived `Default` (all-`None`) and was never affected.

## Breaking change

`GenerationConfig::default()` no longer yields `temperature: 1.0` / `max_output_tokens: 4096`. Callers who *want* those values must now set them explicitly. Changelog entry added under `### Fixed` and marked `[**breaking**]`.

## Deliberately not addressed

The streaming path reports a `MAX_TOKENS` turn carrying no content as a **successful, empty completion** — `streaming.rs` logs at `debug!` and emits `FinalResponse`, where the non-streaming path raises a `ResponseError` and applies `require_non_empty_response`. That asymmetry is what hid this bug in production: the truncation surfaced as an unexplained empty response rather than an error. `finish_reason: Length` *is* propagated on the final response, so it is detectable but not enforced.

Left for a follow-up: the fix direction (hard error vs. `warn!`) is a maintainer call, it is a separate behavior change with its own blast radius, and it is not needed to close #2322. Flagged in the issue.


gold-silver-copper:
Superseded by #2324, which includes this commit unchanged and closes the rest of the chain: the finish reason lost in rig-agent, the empty-truncated-turn success, and the discarded `outputTokenLimit`.

Leaving this open for now — the two are separable, and this one's green CI is useful evidence that the `Default` change alone is safe. Close it in favour of #2324 whenever you prefer.

Issue #2045, opened 2026-07-08
refactor(core)!: add #[non_exhaustive] to core error enums (PromptError, ToolError, ToolSetError, VectorStoreError, StructuredOutputError)


gold-silver-copper:
# Summary

`#[non_exhaustive]` coverage on rig-core's error enums is backwards: peripheral enums have it, but the core enums users match on constantly do not — so adding any variant to them is a semver-breaking change. The codebase is already contorting itself to avoid that. Fix in the next breaking release (there is an open window: the last four refactor commits were all `!`).

References against `695490d9`.

# Current state

**Have `#[non_exhaustive]`:** `CompletionError` (`completion/request.rs:88`), `EmbeddingError`, `TranscriptionError`, `ImageGenerationError`, `AudioGenerationError`, `RerankError`, `MemoryError`, `ClientBuilderError`, `ProviderClientError`, `VerifyError`.

**Lack it:** `PromptError` (`completion/request.rs:148`), `StructuredOutputError` (`request.rs:249`), `ToolError` (`tool/mod.rs:46`), `ToolSetError` (`tool/mod.rs:534`), `VectorStoreError` (`vector_store/mod.rs:33`).

# The cost is already visible in the code

`completion/request.rs:191-198` — the `From<MemoryError> for PromptError` impl smuggles memory errors through `CompletionError::RequestError(Box::new(err))`, and its own doc comment says why:

```rust
/// Surface [`crate::memory::ConversationMemory`] failures through the existing
/// [`CompletionError::RequestError`] variant so adding memory support does not
/// require a new top-level [`PromptError`] arm in downstream exhaustive matchers.
```

I.e. a feature had to hide its error type inside a `Box<dyn Error>` because `PromptError` can't grow. With `#[non_exhaustive]`, memory failures could get a proper typed variant (and callers could actually match on them).

Related smell that a typed variant would also fix: `ToolError`'s `Display` does string-prefix sniffing (`tool/mod.rs:65`, `error_str.starts_with("ToolCallError: ")`) to dedupe recursively-nested agent-as-tool errors.

# Proposal

1. Add `#[non_exhaustive]` to the five enums above (breaking: downstream exhaustive matches need a `_` arm).
2. In the same release, add the variants the workarounds were avoiding:
   - `PromptError::MemoryError(MemoryError)` — retire the `RequestError` smuggling.
   - Consider a typed nested-tool variant on `ToolError` to remove the Display string-sniffing.
[...]

PR #2324, opened 2026-08-14
fix(gemini, agent): close the whole output-budget truncation chain, not just the 4096 cap

### LINKED ISSUE #2322: bug: Gemini injects a hardcoded maxOutputTokens=4096 and temperature=1.0 into structured-output and image requests

gold-silver-copper:
Closes #2322. **Supersedes #2323** — that PR fixed the trigger; this one closes the whole chain. #2323's commit is included here unchanged, so its green CI still stands as evidence that the `Default` change alone is safe.

## The failure chain

What users saw was an unexplained blank answer. Five links got them there, and #2323 only cut the first:

| # | Link | Closed by |
| --- | --- | --- |
| 1 | `GenerationConfig::default()` hardcoded `max_output_tokens: Some(4096)`, injected into structured-output and image requests | `deda7a7d` (#2323) |
| 2 | A 16k-budget turn was truncated at 4096 | ↑ |
| 3 | rig-agent's assembler **discarded** `FinishReason::Length` | `9e172020` |
| 4 | The answer-less turn finalized as a successful `""` | `9e172020`, `9c1767e2` |
| 5 | Nothing in rig knew the real limit was 65536 | `9e172020` |

### Link 4 needed a second pass

`9e172020` gated the rule on `is_empty_assistant_turn`, which is true only for a turn with no items or one empty text block. A **reasoning-only** turn is neither — so it still finalized as a successful `""`.

That is the *common* shape of this bug, not a corner of it: providers bill thinking tokens against the output limit (our cassettes show `thoughtsTokenCount` of 176–307 on ordinary prompts), so a model that exhausts its budget mid-thought is cut off before emitting any answer text. The headline scenario of #2322 — thinking model, small budget, blank answer — was still broken.

`9c1767e2` corrects the predicate to `turn_delivered_no_answer` (no tool call, no non-empty text). `is_empty_assistant_turn` is untouched and still governs the history push, so reasoning is still recorded and a truncated thinking turn stays debuggable. Reasoning is not an answer; tool calls are, so a truncated tool-call turn still routes to execution.

## The reason was lost in rig-agent, not the Gemini adapter

**#2322's original text blames the wrong layer — corrected here.** The adapter is fine: `map_google_finish_reason` maps `MAX_TOKENS` → `FinishReason::Length`, and `StreamFinal::finish_reason` carries it correctly.

The loss was in `run/streamed.rs`: the `Final` arm read `usage` and `saw_text` off the terminal record and dropped everything else. `grep finish_reason crates/rig-agent/src/agent/run/` returned **nothing**. So on the agent API — how nearly everyone hits this — a truncated turn wasn't merely unenforced, it was **undetectable**.

`CompletionCall` now carries `finish_reason`, populated on both the blocking and streamed recorders from the same terminal record identity already comes from. Per call rather than per run: a multi-turn run has N reasons, and the truncated one is the diagnostic.

## An empty truncated turn is no longer a successful answer

A turn with **zero** content reporting `Length` or `ContentFilter` now fails with a `ResponseError` naming the reason — matching what the blocking Gemini path already did for a content-less candidate. The two surfaces disagreeing is precisely what let truncation surface as a blank.

The empty-success path is **intentional** (documented in `run/mod.rs`, pinned by a test), so the rule is deliberately narrow:

- partial output → truncated: **still valid**, reason preserved on the call
- textless with `Stop` / `ToolCalls`: **unchanged**
- `FinishReason::Other`: **not** truncation — it carries a provider's own wire spelling with no normalized meaning, so erroring would fail benign runs

`final_response_can_remain_empty_for_truly_textless_turns` passes **untouched** — its mock leaves `finish_reason` unset via `StreamFinal::new`, so the new rule cannot fire on it. Verified against the code, not assumed.

## Rig was discarding Gemini's real output limit

`ListModelEntry` parsed `inputTokenLimit` and dropped `outputTokenLimit`; the generic `Model` had nowhere to put it. So nothing in the library knew `gemini-2.5-flash` allows **65,536** output tokens — ~16× the hardcoded 4096. Our own committed listing cassette had the answer all along.

`Model::max_output_tokens` fixes that, distinct from `context_length` (input window vs output ceiling). It is deliberately **not** sent on requests: omitting an output limit is what lets the provider apply its own per-model default, and populating `maxOutputTokens` from the listing would reintroduce a rig-chosen cap by another route — the same bug class, differently spelled.

## Test matrix

| # | scenario | kind | asserts |
| --- | --- | --- | --- |
| 1 | schema, no `max_tokens` | live | no sampling fields |
| 2 | schema, `max_tokens` 16384 | live | `maxOutputTokens` only — a budget above the old cap arrives intact |
| 3 | `temperature` alone | live | `temperature` only |
| 4 | C15: `max_tokens` alone | live | `maxOutputTokens` only |
| 5 | image generation | live | no sampling fields |
| 6 | **thinking config, no `max_tokens`** | live *(new)* | `{"thinkingConfig":{…}}` alone |
| 7 | **streaming structured output** | live *(new)* | no sampling fields |
| 8 | **model listing** | existing cassette *(new assertion)* | 65536 for `gemini-2.5-flash`; >40 models annotated |
| 9 | **empty + `MAX_TOKENS`** | synthetic *(new)* | errors, naming the reason |
| 10 | **partial text + `MAX_TOKENS`** | synthetic *(new)* | stays valid, reason on the call |
| 11 | **empty + content filter** | synthetic *(new)* | errors |
| 12 | **empty + `Other`** | synthetic *(new)* | still finalizes |
| 13 | **reasoning-only + `MAX_TOKENS`** | synthetic *(new)* | errors — the common shape |
| 14 | **reasoning-only + content filter** | synthetic *(new)* | errors |
| 15 | **reasoning + text + `MAX_TOKENS`** | synthetic *(new)* | stays valid, reason on the call |
| 16 | **reasoning-only + `Stop`** | synthetic *(new)* | still finalizes |
| 17 | **reasoning reaches consumer before the error** | synthetic *(new)* | truncated thinking stays debuggable |

Scenarios 9–12 are synthetic because a provider cannot be made to emit an exactly-empty `MAX_TOKENS` turn or a safety filter on demand. Each states that in its doc comment, per `tests/README.md`'s rule for non-cassette tests of provider-facing behavior.

Scenario 6 matters more than it looks: it is the usage pattern that hid #2283, and the one most exposed to a restored non-`None` `Default`, because callers build these configs with `..Default::default()`.

**The guard tests were verified to fail without the fix.** With the rule disabled, 9 and 11 fail; 10 and 12 pass either way, as they should — so the pair pins the behavior rather than the implementation.

## Verification

```
cargo test --workspace --all-features     4384 passed; 0 failed
cargo test -p rig --all-features --test gemini    183 passed; 1 ignored
cargo fmt --all -- --check                clean
cargo clippy --all-targets --all-features clean
```

**Zero existing cassettes modified.** `git diff --stat <base> -- tests/cassettes/` is empty; the only cassette changes are the two new files. The 18 re-recorded in #2323 are carried over untouched.

## Breaking changes

Documented in both changelogs and `MIGRATING.md`:

- `AgentRun::record_streamed_completion_call` takes the finish reason as a third argument
- `StreamedTurnEvent::Completed` gains `finish_reason`
- `ModelTurn` / `StreamedTurn` gain `finish_reason` (both `#[non_exhaustive]`; `ModelTurn::new` unchanged)
- `Model` gains `max_output_tokens`
- **Behavior:** an empty turn reporting `Length`/`ContentFilter` now errors

`CompletionCall.finish_reason` is serde-defaulted, so pre-#2322 run JSON still loads.

## Reviewing this

`deda7a7d` (the `Default` fix, already reviewed as #2323) and `9e172020` (everything else) are separable — the second is a strict superset in value but the two can be judged independently. If the behavior change in link 4 is contentious, it is the only part that changes runtime behavior for existing users; W1's observability alone would still make truncation detectable, and I'd be glad to split it out.


gold-silver-copper:
Pushed `9c1767e2` — a P1 in this PR's own guard, found while re-reading the predicate.

**The shape that slipped through.** `9e172020` gated the rule on `is_empty_assistant_turn`, true for only two shapes: no items, or one empty unannotated text block. A turn carrying **one `Reasoning` item and nothing else** is neither, so the guard never fired and the run finalized successfully with `output == ""`.

Repro against `9e172020` — reasoning event + `FinalResponse(Length)` → `FINALIZED SUCCESSFULLY, output=""`.

**Why it's P1 rather than an edge case.** Gemini counts thinking tokens against `maxOutputTokens` — the committed cassettes show `thoughtsTokenCount` of 176–307 on ordinary prompts. A model that exhausts its budget while thinking is cut off *before* emitting any answer text, so the turn carries reasoning and no text. That is the headline scenario of #2322, and it was still broken on the branch claiming to fix it. The earlier tests passed because they only ever scripted zero-content or text-then-truncate, never reasoning-only.

**The fix is a predicate correction, not a wider rule.** "Does this turn belong in history" and "did this turn deliver an answer" are different questions that diverge on exactly this shape. `turn_delivered_no_answer` (no tool call, no non-empty text) is what the guard needed; `is_empty_assistant_turn` is untouched and still governs the history push, so reasoning is still recorded and a truncated thinking turn stays debuggable.

| turn shape | reason | outcome |
| --- | --- | --- |
| nothing | `Length` / `ContentFilter` | error |
| **reasoning only** | `Length` / `ContentFilter` | **error (the fix)** |
| any non-empty text | `Length` | valid, reason kept |
| reasoning + text | `Length` | valid, reason kept |
| tool calls | `Length` | routes to tools |
| any | `Stop` / `ToolCalls` / `Other` | unchanged |

**Verified to fail without the fix.** Reverting the predicate fails the two erroring tests while the three "stays valid" tests still pass — so the set pins behavior, not implementation. Five synthetic tests added (rows 13–17 in the updated matrix), each documenting why it can't be a cassette.

4389 workspace tests pass (+5), fmt and clippy clean, **no cassette churn**. Changelog and MIGRATING corrected — both said "produced zero assistant content", which described the old predicate.

gold-silver-copper:
Confirmed and fixed in `b3b8b53f`. Good catch — and the second time this predicate has been wrong, which changed how I fixed it.

**Repro:** `turn_delivered_no_answer(&[AssistantContent::Image(..)])` returned `true`.

Not hypothetical: ten providers emit assistant images, Gemini's blocking path among them (`PartKind::InlineData` → `image_base64`), and image tokens count against the same output budget this PR is about — so a truncated image turn is exactly the combination that would have hit it.

**The root cause was the catch-all, not the missing arm.** The predicate matched `Text` and `ToolCall` and sent everything else to `_ => false`. That default is backwards: "not one of the two shapes I thought of" is not evidence a turn answered nothing. The wildcard also made it silent — `Image` had existed all along, and any future variant would inherit the same wrong bucket without a compiler warning.

The match is now **exhaustive, no `_` arm**: `Image` is an answer, `Reasoning` is the single exclusion, and adding a variant to `AssistantContent` breaks this build. Verified by temporarily adding one — `E0004` fires at the predicate.

Given this predicate has now been wrong twice (`Reasoning`, then `Image`), a third targeted patch seemed like the wrong response; making the compiler enforce the enumeration is the fix that generalizes.

**Tests:** two unit tests at the level the defect actually lives at, rather than through a stream — this is pure classification over a rig-owned enum, and a variant inheriting the wrong bucket is invisible at the wire level. One pins every variant plus mixed shapes (reasoning+image answers; reasoning+empty-text does not); the other pins that the two turn predicates diverge on exactly one shape and agree everywhere else. Confirmed to fail with `_ => false` restored.

4391 workspace tests pass (+2), clippy clean under `-D warnings`, no cassette churn.

gold-silver-copper:
Both confirmed and fixed in `3521f86b`.

**P3-1 — the divergence is two shapes, not one.** Verified:

```
[Text{ text: "", additional_params: Some(..) }]
  is_empty_assistant_turn  = false   (annotated -> recordable)
  turn_delivered_no_answer = true    (no text -> no answer)
```

The behaviour was already right; the doc and test were wrong. The old test checked three agreeing shapes — `[]`, empty text, real text — none annotated, so it never touched the case it claimed to exclude. Same failure mode as the two P1s: the test encoded the documentation's assumption rather than probing it. The doc now names both divergent shapes, and the test enumerates the divergent set and the agreeing set explicitly (image and tool-call turns included).

**Metadata-only text is not an answer** — now stated as its own decision with its own test, per your request, rather than left to fall out of the predicate's code. Rationale: `assistant_text_from_choice` concatenates `text.text` alone, so an annotation-only turn hands the caller `""`. The metadata is worth recording — `prompt_request_preserves_metadata_only_text_turn_in_history` already pins that — but it is metadata *about* an answer, not the answer. The test also pins that an annotation does not suppress a real answer beside it, and records the run-level consequence if anyone reclassifies it: whether a truncated annotation-only turn errors or silently returns `""`.

**P3-2 — dangling pointer removed.** Correct: this path returns `Err`, so no `PromptResponse` is ever built and the advice pointed nowhere. The reason is named in the error itself, which is the only place it appears on this path; a comment records why the pointer must not come back.

I left `MIGRATING.md`'s reference to `completion_calls` alone deliberately — it describes the *success* path (partial output then truncation), where a `PromptResponse` genuinely is returned.

4392 workspace tests pass, gemini 183, clippy clean under `-D warnings`, no cassette churn.

gold-silver-copper:
Ran a review over this branch. Four findings; three applied in `5798c1a1`, one raised for a maintainer decision.

**F1 (medium) — the error gave advice that cannot help.** Correct and the most serious of the four. The guard fires for `ContentFilter` as well as `Length`, but the message ended unconditionally with "the turn was cut short … raise max_tokens". For a safety block (Gemini `SAFETY`/`PROHIBITED_CONTENT`, Anthropic refusal, OpenAI `content_filter`) that is both factually wrong — nothing was cut short — and actively misleading, since it points at a setting that cannot fix a filtered response. The remedy now branches on the reason.

Both filter tests only asserted that *some* error occurred, which is exactly why they missed it. They now assert the reason is named and that `max_tokens` is **not** suggested; the `Length` test asserts it is.

**F3 (low, but the worst-placed gap) — the blocking surface had no coverage.** All nine guard tests drove `stream_prompt`, and `MockTurn` had no way to set a finish reason, so `runner.rs`'s `.with_finish_reason(resp.finish_reason())` was never exercised — deleting that line failed nothing. Given this PR's premise is that *the two surfaces disagreeing* is what let truncation surface as a blank, that is the worst place to be untested. Added `MockTurn::with_finish_reason` and three blocking tests, and verified they close the gap: removing the `runner.rs` line now fails all three.

**F4 (low) — undocumented narrowing.** Also correct: the guard precedes the `OutputMode::Tool` re-prompt, so an answerless truncated turn now fails immediately rather than consuming a retry. I kept the behaviour and documented it in both CHANGELOG and MIGRATING — the re-prompt could not have helped, since the budget or the filter, not the phrasing, is what stopped the turn, so the retry would truncate again and fail less specifically. Happy to reorder instead if you disagree.

**F2 (low) — not applied, your call.** `Model` is not `#[non_exhaustive]` while `CompletionCall`/`ModelTurn`/`StreamedTurn` are, so adding a field to it breaks downstream struct literals. The inconsistency is real and this release is already breaking, so now is the cheapest time to fix it — but making it non-exhaustive permanently forbids the struct-literal construction the type's own doc example teaches, which is a separate decision about a shared public type and not one I want to make inside an output-budget bugfix. Say the word and I'll add it.

4395 workspace tests pass (+3), gemini 183, clippy clean under `-D warnings`, no cassette churn.

gold-silver-copper:
Filed F2 as #2325 rather than folding it into this PR — it is a decision about a shared public type, and settling it inside an output-budget bugfix would bury it.

The issue records the verified inconsistency (`Model`/`ModelList` lack `#[non_exhaustive]`; `CompletionCall`/`PromptResponse`/`ModelTurn`/`StreamedTurn` have it — the four that gained a field in this PR without breaking anything), the precedent from #2045, and the migration path: public fields stay writable, so callers move to `Model::from_id(..)` plus field assignment and no new API is needed. In-repo churn is two call sites.

If you'd rather take it here while the breaking window is open, say so and I'll add it to this branch.

PR #2330, opened 2026-08-15
refactor(core)!: add #[non_exhaustive] to the receive-side structs that grew fields this cycle

gold-silver-copper:
Closes #2325

## Why now

`#[non_exhaustive]` forbids struct-literal construction from other crates, so **adding it is itself a semver break** — it can only land inside a breaking release window. That window is open: release PR #2221 (v0.42.0) already flags `rig-core` as "⚠ API breaking changes", and `Model` just took exactly the break the attribute prevents — #2324 added `Model.max_output_tokens` and broke a downstream struct literal, while the four `rig-agent` types in the same PR (`CompletionCall`, `PromptResponse`, `ModelTurn`, `StreamedTurn`) gained fields harmlessly *because* they carry it. Outside a breaking window the attribute cannot be added at all, and #2321/#2079 will keep producing new listing fields in the meantime. #2045 made the same "there is an open window" argument for the error enums and was accepted.

## Scope: the release window's census, not a general sweep

#2221's `cargo-semver-checks` output is a ready-made census of every externally-constructible `rig-core` struct that gained a field this cycle — i.e. every type that *already demonstrated* it grows fields while lacking the attribute. That census is the scope line here, plus `ModelList` (named by #2325; it shares the module and the receive-mostly usage). Deduplicated it is 19 rows; every one is accounted for below, and each was verified to still exist on `origin/main` at the cited path before acting.

Note the raw output contains **four rows beyond the list in the issue thread** — `responses_api/streaming.rs::StreamingCompletionResponse`, `cohere/streaming.rs::StreamingCompletionResponse`, `message.rs::ToolResult`, and anthropic's `ToolDefinition`. They are included.

**Decision rule.** Apply the attribute when the type is a value the library hands *out* and callers overwhelmingly read rather than build. Do **not** apply it where external construction is the type's contract — concretely: *is the struct literal the documented/actual construction idiom, with no ergonomic alternative?* That test, not "is it request-shaped", is what separates the two skips below from the 18.

## Sweep table

| # | type | path | gained this cycle | verdict |
| --- | --- | --- | --- | --- |
| 1 | `model::Model` | `model/listing.rs` | `max_output_tokens` | **attribute** |
| 2 | `model::ModelList` | `model/listing.rs` | — (per #2325) | **attribute** |
| 3 | `completion::message::ToolCall` | `completion/message.rs` | `provider` | **attribute** |
[...]

Issue #2314, opened 2026-08-13
feat: carry the provider transport request id on completion errors (error-path identity capture)


gold-silver-copper:
## Feature Request

Carry the provider's transport request id (`request-id` / `x-request-id` / SDK metadata) on completion **errors**, so a failed call can be reported to provider support — the moment the id actually matters.

### Context

#2265 / #2313 added `provider_request_id` to every *successful* surface: core `CompletionResponse` and `StreamFinal`, provider raw responses, hook events, and `PromptResponse.completion_calls`. Error-path capture was deliberately scoped out and documented as a limitation.

The follow-up edge-case matrix on #2313 turned that limitation from a hypothesis into recorded evidence: **providers send the id on error responses too, and rig currently drops it.**

- `tests/cassettes/anthropic/response_identity_edge/provider_error_response_surfaces_cleanly.yaml` — a `404` on a nonexistent model whose recorded response headers include `request-id` (scrubbed to `req_REDACTED_1`).
- `tests/cassettes/openai/response_identity_edge/provider_error_response_surfaces_cleanly.yaml` — same shape, `x-request-id` present on the error response.

So a caller who hits a 4xx/5xx — the exact case where "give support this request id" applies — gets a `CompletionError` with the body and status, while the id that was on the wire beside them is discarded.

### Why this matters

- Support workflows: Anthropic and OpenAI both ask for the request id when investigating a failed call. Today rig callers must re-derive it from their own HTTP-level logging or lose it.
- Transcript/observability pipelines (the #2265 use case): a persisted run should record identity for failed attempts as well as completed ones — failed calls are the ones people go back and investigate.
- Retry diagnostics: a hook implementing retry-on-provider-error can tag each failed attempt with its id, matching the per-attempt semantics #2313 established for successes.

### Where the pieces already are

Everything but the error-type surface exists:

- **Blocking**: `send_completion` (`providers/internal/completion_send.rs`) already reads the provider-named header off the response *before* consuming the body — including on the non-success branch, where the captured value is currently unused. The error is minted one line later via `from_http_response(status, raw_body)`.
- **Streaming**: the SSE `RequestIdSlot` is populated at connect, before a stream can fail mid-flight, so a post-connect stream error already has the id available in the slot; a connect-time HTTP failure flows through the same response object the capture reads.
- **Bedrock**: the AWS SDK error types implement `RequestId` just as the success outputs do.
[...]

PR #2335, opened 2026-08-15
refactor: remove #[non_exhaustive] from the workspace

gold-silver-copper:
Removes all **53** `#[non_exhaustive]` attributes from the workspace — `rig-core` (26), `rig-agent` (18), `rig-bedrock` (8), `rig-candle` (1). Public structs read as plain data again: struct literals, functional update (`..Default::default()`) and exhaustive `match` all work from any crate.

Supersedes #2330, which was adding the attribute to 18 more types; that PR is closed in favour of this one.

## This is not a breaking change

`#[non_exhaustive]` only ever *restricted* downstream code, so removing it is permissive. `cargo semver-checks` v0.48.0 against the parent commit:

```
Checked 196 checks: 196 pass, 57 skip
Summary no semver update required
```

Nothing downstream stops compiling. One thing may start warning — see below.

## What it costs, stated plainly

The bargain inverts. From here on, **adding a field to any of these structs, or a variant to any of these enums, is a breaking change** and has to wait for a breaking release. The attribute also cannot be reintroduced outside a breaking window, so this is not cheap to revisit. That is the deliberate trade being made for construction ergonomics; it is written into `CHANGELOG.md` and `MIGRATING.md` so nobody has to rediscover it.

For scale, the churn this now exposes: in the current cycle alone, 19 externally-constructible `rig-core` structs gained a field, and #2329 had to write a breaking-change note for `anthropic::streaming::StreamingCompletionResponse` precisely because that type lacked the attribute. Those notes become the normal case.

## Fallout, and how it was handled

**Unreachable wildcards (build failures under `-D warnings`).** Exactly two in-tree matches covered every variant of `ReasoningContent` behind a `_` arm that is now unreachable:

- `tests/common/reasoning.rs:617` — this module is `#[path]`-included by **26** test binaries, so it fails 26 targets at once
- `tests/core/streaming_conformance.rs:524`

Both wildcards were already dead at runtime; both are deleted. Nothing asserts on the values they produced.

**Stale rustdoc.** Three comments asserted the attribute as the *reason* to use a constructor. Rewritten to recommend the constructors on their own merits — still true, still the better form, just no longer compiler-enforced:

- `provider_response.rs` (`ProviderResponseError`)
- `completion/request.rs` (`ProviderCapabilities`)
- `agent/hook.rs` (`ModelSelection::new`)

**One contortion unwound**, which is the readability this change is for — `tests/providers/openai/cassette/gpt_5_6_reasoning.rs` round-tripped a `Reasoning` through `serde_json::from_value` *because* the type was non-exhaustive. It is a plain struct literal now.

**Not touched:** comments about the AWS SDK's own `#[non_exhaustive]` types (`rig-bedrock/types/converse_output.rs`, `providers/internal/wire.rs`). Those attributes belong to `aws-sdk-bedrockruntime` and are unaffected.

## Two findings worth a maintainer's decision

A per-attribute audit of all 53 turned up three "risky" and one genuine correctness widening. Both are documented in `MIGRATING.md`; neither blocks CI; both are one-line to reverse if you disagree.

**1. `StreamFinal` / `completion::CompletionResponse` lose their last construction guard.** Both normalize an empty identifier string to `None`, and that rule lives *only* in the `with_*_id` setters generated by `response_metadata_setters!`, whose own doc says:

> An empty string is treated as absent: gateways that echo `""` for fields they don't populate must not produce a `Some("")` that differs between the buffered and streaming paths. **All identifier and model setters share this rule so the invariant lives here rather than at every provider call site.**

Both types deliberately route `Deserialize` through those setters too (`#[serde(from = "StreamFinalRepr")]` → `StreamFinal::new(..).with_optional_message_id(..)`). So every construction path normalized — and `#[non_exhaustive]` was the last thing preventing an external struct literal from bypassing them. After this change, downstream code can write `StreamFinal { message_id: Some(String::new()), .. }` and hand it straight back in through the public `StreamedAssistantContent::final_response(res: StreamFinal)`.

If that matters, the durable fix is to make the normalized fields private with accessors, so the invariant is structural rather than attribute-dependent. Keeping the attribute on these two types alone would also do it.

**2. `CandleError` can no longer be matched portably.** Three of its variants are `#[cfg(not(target_family = "wasm"))]`, and rig-candle ships a wasm consumer (`examples/candle_wasm_chat`). While the enum was non-exhaustive, one wildcard-bearing `match` compiled on both targets. Now an exhaustive match cannot: the arms are missing variants on wasm, and adding them plus a wildcard trips `unreachable_patterns` on wasm. Nothing in this workspace matches on it, so CI is unaffected — `MIGRATING.md` tells consumers to keep a wildcard.

## Superseded documentation

Shipped guidance describes ten-plus types as `#[non_exhaustive]` and tells readers to use constructors for that reason: "Core errors are `#[non_exhaustive]`" (0.39→0.40), `ProviderResponseError` (#2314), `CompletionResponse`, `ProviderCapabilities`, `ModelTurn`/`StreamedTurn` (#2322), `MultiTurnStreamItem`, `DocumentSourceKind`, `RerankError`, `MemoryError`, openai's `ToolChoice`.

Those passages are left intact — they are accurate history for the releases they document. The new `MIGRATING.md` section lists them as superseded and notes that the constructors they point at all still exist and are still recommended; only the compiler no longer insists.

## Verification

| check | result |
| --- | --- |
| `RUSTFLAGS="-D warnings" cargo clippy --workspace --all-targets --all-features` | clean |
| `cargo test --workspace --all-features` | exit 0 — 4,574 tests, 163 suites, 0 failures |
| `cargo test --doc --workspace --all-features` | exit 0 |
| `cargo check -p rig-core --target wasm32-unknown-unknown` | clean |
| `cargo doc --locked --no-deps --all-features` (`RUSTDOCFLAGS=-D warnings`, CI's exact invocation) | exit 0 |
| `cargo semver-checks --baseline-rev main -p rig-core --all-features` | 196 pass, no semver update required |

The 53 removals were also audited individually before being applied — public reachability, whether the attribute was already inert (9 were, behind private fields or non-public paths), every cross-crate `match`/destructure site that could become unreachable, whether a constructor enforced an invariant a literal would bypass, and which shipped docs referenced it. That audit is what produced the two findings above.


gold-silver-copper:
Filed #2336 for the one finding in this PR's audit that is a genuine correctness widening rather than a style change: the empty-string-is-absent rule on `StreamFinal` / `CompletionResponse` is enforced only in the `response_metadata_setters!` setters, so it is advisory for anyone constructing those types directly.

Worth stressing that #2336 is not a request to revert this PR. The fields were always `pub`, so `x.message_id = Some(String::new())` compiled even with `#[non_exhaustive]` — the attribute blocked origination, never mutation. The proposal there is `Option<WireId>` instead of `Option<String>`, which makes `Some("")` unrepresentable and needs no attribute and no getters. rig already uses exactly that pattern for `WireId`, whose doc states the same principle.

It is breaking, so it wants the window #2221 still has open.

PR #2343, opened 2026-08-15
feat(voyageai): expose embedding request options

sergiomeneses:
**Why:** Voyage AI supports retrieval-aligned embeddings via input_type ("query" / "document"), which prepends retrieval prompts such as “Represent the document for retrieval: “. measured on a 36-chunk corpus: chunks embedded with input_type="document" (voyage-4-large) plus queries with input_type="query" (voyage-4-lite) keep recall@10 at 12/12 and cut off-topic top-1 cosine from 0.335 to 0.222, enabling a higher similarity floor.

**What:** Adds a public EmbeddingOptions struct (input_type, truncation, output_dimension, all Option, default None) and EmbeddingModel::with_options(...). Options are serialized into the /embeddings request only when set; the existing constructor and embedding_model factory stay backward-compatible. The generic EmbeddingsBuilder is untouched (provider-agnostic).

**Deferred:** output_dtype (quantized int8/uint8/binary/ubinary) is documented as a TODO: those dtypes return integers, not floats, and need a separate response-parsing change.

Tests. Unit tests assert the request body contains the options when set and omits them when None. cargo test -p rig-core passes.

### DIFF
--- a/crates/rig-core/src/providers/voyageai.rs
+++ b/crates/rig-core/src/providers/voyageai.rs
@@ -90,6 +90,7 @@ impl<T> EmbeddingModel<T> {
             client,
             model: model.into(),
             ndims,
+            options: EmbeddingOptions::default(),
         }
     }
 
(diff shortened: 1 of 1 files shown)


gold-silver-copper:
awesome stuff!

PR #2344, opened 2026-08-15
fix(embeddings): builder input order

sergiomeneses:
`build_with_usage` reassembled results by iterating a `HashMap<usize, T>`, whose iteration order is arbitrary, so the returned `Vec<(T, Vec<Embedding>)>` did not match input order. Callers zipping positionally silently stored each embedding under the neighbor’s content (recall@5 dropped from ~0.92 to ~0.42 on a 36-chunk corpus).

Documents are now stored in a position-indexed `Vec<T>` and merged with `enumerate()`, preserving input order. Removed the four `result.sort_by(...)` workarounds that masked the defect, and added a batch-boundary test asserting exact input order.

Verified: cargo check -p rig-core, full cargo test -p rig-core, clippy.

### DIFF
--- a/crates/rig-core/src/embeddings/builder.rs
+++ b/crates/rig-core/src/embeddings/builder.rs
@@ -353,4 +353,27 @@ mod tests {
             )
         )
     }
+
+    #[tokio::test]
+    async fn test_build_preserves_input_order_across_batches() {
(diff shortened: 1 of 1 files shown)


gold-silver-copper:
good work!

PR #2370, opened 2026-08-17
feat(memory)!: `TemplateCompactor` caps are literal, with an explicit opt-out

### LINKED ISSUE #2181: Get rid of 0 -1 sentinel values
Rig occasionally uses 0 to mean infinite, should probably switch those to Option types or use NonZerou64 magic

gamboacarlos:
Fix #2181

### Problem

`TemplateCompactor::with_max_bytes(0)` previously meant **“unbounded”**. This made `0` a special value with semantics that differed from the rest of the API.

In particular, the cap was non-monotonic:

* `0` → unbounded
* `1` → header plus marker
* `2` → slightly more output

This is especially error-prone when the value is computed rather than explicitly typed. For example:

```rust
budget.saturating_sub(overhead)
```

or a configuration field left at its default value could unexpectedly produce `0`, turning a bounded compactor into an unbounded one.

This is particularly dangerous when used through `CompactingMemory`. The compactor sits outside the wrapped `MemoryPolicy`'s budget, so an accidentally unbounded compactor can cause the rollup to grow without limit: each pass embeds the previous summary verbatim.

### Changes

* `with_max_bytes(n)` now treats `n` **literally**. A value of `0` is a zero-byte cap rather than an unbounded setting.
* **New:** `without_max_cap()` explicitly opts out of the byte limit and restores unbounded rollup behavior.
* This provides a way to opt out at the call site instead of reserving `0` as a special value.
* Documentation and `CHANGELOG.md` have been updated to describe the literal-`0` behavior and the new unbounded escape hatch.

### Breaking behavior

This is a behavioral breaking change for callers that currently pass `0` as an “unbounded” switch.

The API signature remains unchanged, so existing code will still compile, but callers relying on `with_max_bytes(0)` being unbounded will now receive a zero-byte cap.

Migrate those callers to:

```rust
.without_max_cap()
```

### Verification

All checks pass:

* `cargo clippy --all-features --all-targets`
* `cargo fmt -- --check`
* `cargo test -p rig`
* `cargo test -p rig --all-features`
* `cargo test -p rig --test core`
* `cargo test -p rig-memory`

### DIFF
--- a/crates/rig-memory/CHANGELOG.md
+++ b/crates/rig-memory/CHANGELOG.md
@@ -6,6 +6,29 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
+### Added
+
+- *(memory)* `TemplateCompactor::without_max_bytes` clears a configured size
+  cap, restoring the default unbounded rollup. It is the inverse of
+  `with_max_bytes` and the migration target for callers that passed `0`.
+
+### Changed
+
+- *(memory)* [**breaking**] `TemplateCompactor::with_max_bytes(0)` no longer
+  means "unbounded" — the cap is now taken literally. Use `without_max_bytes`
+  for unbounded behaviour. The signature is unchanged, so this does not break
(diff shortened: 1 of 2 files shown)


gold-silver-copper:
There's many more cases than just this, I will be fixing this myself later. Thank you for your time but I will be closing this.

PR #2377, opened 2026-08-18
fix(rig-postgres): filtered and thresholded searches generate valid SQL

### LINKED ISSUE #2376: fix(rig-postgres): `member` filters, single-condition filters and thresholds generate invalid SQL
## Summary

`rig-postgres` generates SQL that Postgres rejects for any `VectorSearchRequest` that carries a `member` filter, a single-condition filter, or a `threshold`. When the threshold *would* have parsed, it compared in the wrong direction.

| # | Where (`crates/rig-postgres/src/lib.rs`) | Bug | Effect |
|---|---|---|---|
| 1 | `PgSearchFilter::member` | renders `id is in ($3, $4)` | `syntax error at or near "in"` — the unit test currently pins this string |
| 2 | `search_query` | threshold predicate `distance > $N` is placed in the inner `SELECT`'s `WHERE`, where `distance` is only a select-list alias | `column "distance" does not exist` for every `.threshold(..)` request |
| 3 | `search_query` | `>` on a pgvector *distance*, while `VectorSearchRequestBuilder::threshold` is documented as a **minimum similarity** | even if #2 parsed, it would keep the *least* similar rows |
| 4 | `search_query` | `"WHERE" + condition` with no separator | a single-condition filter renders `WHEREprice >= $3`; only compound filters (which start with `(`) ever parsed |

## Reproduction

```rust
let store = PostgresVectorStore::with_defaults(model, pool);

// 1: syntax error at or near "in"
store.top_n_ids(VectorSearchRequest::builder().query("q").samples(3)
    .filter(PgSearchFilter::member("document->>'name'".into(), vec![json!("a"), json!("b")]))
    .build()).await?;

// 2 + 3: column "distance" does not exist
store.top_n_ids(VectorSearchRequest::builder().query("q").samples(3).threshold(0.8).build()).await?;

// 4: syntax error at or near "WHEREprice"
store.top_n_ids(VectorSearchRequest::builder().query("q").samples(3)
    .filter(PgSearchFilter::gte("price".into(), json!(5)))
    .build()).await?;
```

Generated SQL on `main` for the three shapes (from `search_query`):

```
WHEREprice >= $3
WHEREdistance > $3
WHERE(distance > $3) AND ((kind = $4) AND (id is in ($5, $6)))
```

## Proposed fix (non-breaking)

- `member` renders `IN`.
- Apply the threshold as a minimum similarity on a per-operator expression that repeats the distance operator instead of naming the alias — `1 - (embedding <=> $1) >= $N` for cosine/jaccard, `-(embedding <#> $1) >= $N` for inner product, `-(embedding <-> $1) >= $N` for L2/L1/hamming — the same approach `rig-sqlite` takes. Returned scores stay raw distances in ascending order, so no public behavior changes for callers who never set a threshold.
- Separate `WHERE` from its first condition.
- Pin the generated SQL with unit tests (no database needed) and extend the Docker-backed integration test with a threshold and a `member` filter.

I have a PR ready for this.

jackthepunished:
Fixes #2376

`rig-postgres` produced SQL that Postgres rejects for any `VectorSearchRequest` that carries a `member` filter, a single-condition filter, or a `threshold`; and the threshold compared in the wrong direction. Found by auditing the vector-store crates against `rig-sqlite`, which gets all of this right.

| # | bug | severity |
|---|---|---|
| 1 | `PgSearchFilter::member` renders `id is in ($3, $4)` — not a Postgres operator | high (hard error) |
| 2 | threshold rendered as `distance > $N` in the inner `WHERE`, where `distance` is only a select-list alias → `column "distance" does not exist` | high (hard error) |
| 3 | `>` on a pgvector *distance* while `threshold` is documented as a minimum similarity → would keep the *worst* rows | high |
| 4 | `"WHERE" + condition` with no separator → `WHEREprice >= $3` for any single-condition filter | high (hard error) |

Generated SQL on `origin/main`, as printed by the new tests when run there:

```
WHEREprice >= $3
WHEREdistance > $3
WHERE(distance > $3) AND ((kind = $4) AND (id is in ($5, $6)))
```

and on this branch:

```
WHERE (price >= $3)
WHERE (1 - (embedding <=> $1) >= $3)
WHERE ((kind = $3) AND (id IN ($4, $5))) AND (-(embedding <-> $1) >= $6)
```

## Fix

- `member` renders `IN`.
- The threshold is applied as a **minimum similarity** on a per-operator expression that repeats the distance operator rather than naming the alias: `1 - (embedding <=> $1)` for cosine and jaccard, the negated distance for inner product / L2 / L1 / hamming (pgvector's `<#>` is already the negative inner product). This is the approach `rig-sqlite` takes. It is composed *after* the `$` renumbering pass so the `$1` inside the expression survives, and its bind value is appended after the filter's.
- `WHERE` is separated from its first condition.

**Non-breaking**: returned scores are still raw pgvector distances in ascending order; only requests that previously errored change behavior.

## Tests

- **4 unit tests fail on `origin/main` and pass here** (`threshold_is_a_minimum_similarity_on_the_repeated_expression`, `threshold_placeholder_follows_the_filter_placeholders`, `single_condition_filter_is_separated_from_where`, plus the existing `every_parameterised_operator_uses_dollar_placeholders`, whose assertion pinned `is in`). They call the private `search_query` on a store built with a lazy pool and `rig_core::test_utils::MockEmbeddingModel`, so no database is needed. `similarity_expression_inverts_each_distance_operator` covers all six operators.

Verified locally: `cargo fmt --all -- --check`, `cargo clippy -p rig-postgres --all-targets --all-features`, `cargo test -p rig-postgres`.

### DIFF
--- a/crates/rig-postgres/CHANGELOG.md
+++ b/crates/rig-postgres/CHANGELOG.md
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
+### Fixed
+
+- *(vector-store)* filtered and thresholded searches reach Postgres as valid SQL. `PgSearchFilter::member` rendered `is in`; a single-condition filter abutted the `WHERE` keyword (`WHEREprice >= $3`); and a threshold was rendered as `distance > $N` inside the inner select, where `distance` is only a select-list alias — and would have kept the least similar rows had it parsed. A threshold is now applied as a minimum similarity on the repeated distance operator (`1 - (embedding <=> $1) >= $N` for cosine and jaccard, the negated distance for the other operators), while returned scores remain raw distances in ascending order ([#2376](https://github.com/0xPlaygrounds/rig/issues/2376))
+
 ## [0.42.0](https://github.com/0xPlaygrounds/rig/compare/rig-postgres-v0.41.0...rig-postgres-v0.42.0) - 2026-08-16
 
 ### Other
--- a/crates/rig-postgres/Cargo.toml
+++ b/crates/rig-postgres/Cargo.toml
@@ -29,6 +29,10 @@ uuid = { workspace = true, features = ["v4", "serde"] }
 
 [dev-dependencies]
 anyhow = { workspace = true }
+rig-core = { path = "../rig-core", version = "0.42.0", default-features = false, features = [
+  "derive",
+  "test-utils",
+] }
 log = { workspace = true }
 tracing-subscriber = { workspace = true, features = ["env-filter"] }
 tokio-test = { workspace = true }
--- a/crates/rig-postgres/src/lib.rs
+++ b/crates/rig-postgres/src/lib.rs
@@ -424,8 +424,10 @@ where
 
 #[cfg(test)]
 mod tests {
-    use super::{PgSearchFilter, SearchFilter};
+    use super::{PgSearchFilter, PgVectorDistanceFunction, PostgresVectorStore, SearchFilter};
+    use rig_core::{test_utils::MockEmbeddingModel, vector_store::request::VectorSearchRequest};
     use serde_json::json;
+    use sqlx::postgres::PgPoolOptions;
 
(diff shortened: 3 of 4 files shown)


gold-silver-copper:
Solved by #2531

PR #2378, opened 2026-08-18
feat: add tool call result image

inqode-lars:
Added support for ToolResultContent with Image type.

This is (at least) supported for llama.cpp. But i'm unsure for the official OpenAI-API.

Would it be better to add a new provider for llama.cpp?


gold-silver-copper:
Based on your report, I added this feature in #2380 and #2381, I am also working on a PR that creates a full cassette test matrix for llama.cpp and greatly improves that provider support. I gave you coauthor status for #2381, to thank you for your contribution. I think I will make llama.cpp a separate provider in a follow up PR. Prost

gold-silver-copper:
@inqode-lars llama.cpp has been added as a new provider in #2382

PR #2407, opened 2026-08-23
feat!: transport middleware + run lifecycle hooks (on_run_start/on_run_settled, durable Scratchpad)

gold-silver-copper:
Completes the agent lifecycle with the seams that stopped at the semantic layer, inspired by [Pi's extension lifecycle](https://pi.dev/docs/latest/extensions#lifecycle-overview).

## 1. Transport-boundary middleware (`HttpMiddleware`)

Agent hooks patch a request's *meaning* (`RequestPatch`) but never see the serialized provider payload, the HTTP headers, or the raw HTTP response — that required implementing `HttpClientExt` yourself. New `rig_core::http_client::HttpMiddleware` runs inside `BoxedHttpClient` (#2401), so one implementation covers every provider and all three paths (unary, streaming, multipart):

- `before_request_headers` — mutates the outgoing `HeaderMap` in place (per-request beta headers, auth decoration)
- `before_request_body` — sees the serialized body, may replace it (logging/replay, payload patching); skipped for multipart, which has no single serialized payload
- `after_response` — status + headers as soon as they arrive, **before stream consumption** for streaming calls (rate-limit headers, request ids)

Attach with `BoxedHttpClient::with_middleware(..)`; middlewares compose in attachment order (all header hooks, then all body hooks seeing the final headers, then response hooks). Observe-only on the response, but any phase can *fail* the request through the normal `http_client::Error` path — nothing can be silently swallowed. `examples/http_middleware` demonstrates all three on a live Anthropic call.

## 2. `on_run_start` — pre-run prompt hook

The earliest interception used to be `on_completion_call`; nothing could rewrite or veto the user prompt itself. `AgentHook::on_run_start` fires once before the first model call with the initial prompt + input history. `RunStartAction::{Continue, Rewrite, Stop}` — rewrites chain in registration order (each hook sees the previous rewrite), first stop wins and terminates the run before any provider work with `PromptError::PromptCancelled`. Backed by `AgentRun::{initial_prompt, rewrite_initial_prompt, input_chat_history}` in rig-run so the rewrite is protocol data, not driver trickery.

## 3. `on_run_settled` — terminal event

Per-turn finishes (`ModelTurnFinished`, `StreamResponseFinish`) can be followed by hook-driven retries or tool turns; nothing distinguished "a turn ended" from "nothing follows automatically". `on_run_settled` fires exactly once per run with `SettledOutcome::{Response, Error}`, covering success and every error termination path (including a run-start stop). Every hook in a stack observes it (no short-circuit).

**Finding on the stream side** (the prompt asked for an explicit call): `MultiTurnStreamItem::FinalResponse` already *is* the settled signal on the success path, and error termination is the stream's terminal `Err` item — so no new stream variant is added; the docs on `FinalResponse` now state the equivalence instead of duplicating the event.

## 4. Durable hook state

`Scratchpad` gains string-keyed durable JSON entries (`put_durable`/`get_durable`/`remove_durable`) plus `export()`/`restore()` over a new serializable `rig_run::ScratchpadSnapshot`. `AgentRun` carries an optional `hook_state` slot (`set_hook_state`/`take_hook_state`; serde-default, so old serialized runs load unchanged), and both agent drivers restore a carried snapshot into the live scratchpad at run start. Typed `TypeMap` entries are deliberately not exported — durable state is an explicit, named contract.

## Breaking

- `StepEventKind` gained `RunStart` and `RunSettled` — exhaustive matches must add arms (`AgentHook` itself is unaffected; the new methods are default-implemented).
- MIGRATING.md updated under *0.41 → next*.

## Notes / follow-ups

- No cassette tests were added: the transport hooks are opt-in on `BoxedHttpClient` and provider request paths are unchanged (the middleware unit tests cover unary, streaming, multipart, ordering, and error propagation against the recorded test transports).
- `HookContext` write-back of durable state into `AgentRun` at settle time (fully automatic round-tripping through the futures driver) is left as a follow-up; the manual `export()` → `set_hook_state` path is documented on `Scratchpad`.

Tests: 4 middleware + 1 boxing test (rig-core), run-start chaining / first-stop-wins / scratchpad round-trip (hook.rs), rewrite-validity + hook-state serde (rig-run), and 4 end-to-end lifecycle tests on both surfaces (runner.rs). Full workspace: 5753 passed, clippy `-D warnings` clean, doctests green.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5526,6 +5526,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "http_middleware"
+version = "0.42.0"
+dependencies = [
+ "anyhow",
+ "bytes",
+ "http 1.4.2",
+ "rig",
+ "tokio",
+]
+
+[[package]]
 name = "httparse"
 version = "1.10.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/MIGRATING.md
+++ b/MIGRATING.md
@@ -806,6 +806,33 @@ handed back a silently short list.
 
 ## 0.41 → next
 
+### Run lifecycle hooks and transport middleware
+
+The agent hook surface gained a pre-run and a terminal event, and the erased
+HTTP transport gained a middleware seam. What breaks:
+
+- **`StepEventKind` gained `RunStart` and `RunSettled`.** Exhaustive `match`es
+  over `StepEventKind` (e.g. in an `AgentHook::observes` implementation) must
+  add the two arms. `AgentHook` itself is unaffected: the new `on_run_start`
+  and `on_run_settled` methods are default-implemented.
+- **`BoxedHttpClient` now applies attached `HttpMiddleware`** (new, opt-in via
+  `BoxedHttpClient::with_middleware`). Behavior without middleware is
+  unchanged. `BoxedHttpClient::ptr_eq` still compares the underlying transport
+  only, so two handles differing only in middleware compare equal.
+- **`AgentRun` gained an optional `hook_state` slot** (`hook_state`,
+  `set_hook_state`, `take_hook_state`, carrying the new serializable
+  `rig_run::ScratchpadSnapshot`) plus `initial_prompt`,
+  `rewrite_initial_prompt`, and `input_chat_history`. Runs serialized before
+  this release deserialize unchanged (the field defaults to absent). The agent
+  drivers restore a carried snapshot into the run's `Scratchpad` durable
+  entries at run start.
+
+New, non-breaking: `Scratchpad::{put_durable, get_durable, remove_durable,
+export, restore}` for hook state that must survive a serialized pause, and the
+`RunStart`/`RunStartAction`/`RunSettled`/`SettledOutcome` hook vocabulary. A
+`RunStartAction::Stop` terminates the run before any provider call with
+`PromptError::PromptCancelled`.
+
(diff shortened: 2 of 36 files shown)


gold-silver-copper:
Added a live-recorded cassette matrix for every feature in this PR, across three providers (anthropic, openai, gemini — `tests/providers/*/cassette/lifecycle_matrix.rs`, fixtures under `tests/cassettes/*/lifecycle_matrix/`):

| cell | proves |
|---|---|
| `middleware_phases_observe_a_unary_completion` | all three `HttpMiddleware` phases fire once on a real completion; serialized body and 200 status visible; header injection is replay-transparent |
| `middleware_response_phase_precedes_stream_consumption` | `after_response` sees status/headers on a live SSE call before the stream is consumed; the streamed run settles exactly once |
| `run_start_rewrite_reaches_the_provider` | an `on_run_start` rewrite is what the provider actually answers (marker-word assertion on the recorded reply) |
| `run_settles_once_across_a_multi_turn_tool_run_with_durable_state` | one settle across a real 2-call tool run; durable scratchpad counter survives to the settle export |

Every cell sends through `BoxedHttpClient` + a `WireProbe` middleware and a `LifecycleHookProbe` agent hook (shared in `tests/common/support.rs`); new per-provider wrappers are registered in the cassette-safety suite. Recorded live with `RIG_PROVIDER_TEST_MODE=record`; full `--test anthropic --test openai --test gemini` replay is green (1200/1200).

PR #2408, opened 2026-08-23
feat!: event-sourced hook state — RunEntry log on AgentRun replaces the durable Scratchpad snapshot

gold-silver-copper:
Replaces #2407's durable-Scratchpad snapshot with an event-sourced design mirrored from pi's extension state model (verified against pi's source: `appendEntry`/`CustomEntry`, `getBranch()` replay, and its docs' "State Management" guidance).

## Why the snapshot design lost

The KV-plus-snapshot shape (`put_durable`/`get_durable` → `export()` → `AgentRun::hook_state` → driver `restore()`) required a manual export at exactly the right moment — the "auto write-back" gap — and mixed two key models (type-keyed transient vs string-keyed durable) with silent deserialize-to-`None` reads.

## The replacement

**Hooks persist by appending entries to the run's record; durability holds by construction.**

- `rig_run::RunEntry { kind, turn, value }` — protocol data like `TurnTools`. `AgentRun` gains an append-only `entries: Vec<RunEntry>` (serde-default; pre-existing serialized runs load unchanged) with `append_entry` / `entries` / `entries_of` / `last_entry_of`. `turn` deliberately replaces pi's wall-clock timestamp: rig-run is deterministic sans-IO state, no clocks.
- `HookContext::append_entry(kind, &T)` (turn-stamped, fail-loud serialization, fire-and-forget like pi's `appendEntry`), `entries(kind)` (carried-in entries + this run's appends, in order — the replay), and `last_entry(kind)`. The documented default pattern is **snapshot + last-wins**, which pi's extensions use unanimously.
- The driver seeds the context from `run.entries()` at run start (replacing `restore()`) and **flushes pending appends into the run at every step boundary and in the Done arm** — the serialized run is always current; there is no export moment. Entries appended inside `on_run_settled` are documented as not persisted (the run is finished).
- Mirroring pi's storage-vs-context split: entries are storage, never context — they never reach a provider request, pinned two ways: a rig-run test that a run with entries emits identical `CallModel` steps, and the cassette matrix, whose replay server matches request bodies byte-exactly — replay staying green with the probe now appending entries **is** wire-level proof (no re-recording was needed; fixtures are untouched).
- rig-run stays hook-free (dependency-graph guard unchanged); out-of-process hosts append/read directly on `AgentRun`.

## Deleted (shipped unreleased in #2407, no shims per repo policy)

`Scratchpad::{put_durable, get_durable, remove_durable, export, restore}` and its `durable` field; `rig_run::ScratchpadSnapshot`; `AgentRun::{hook_state, set_hook_state, take_hook_state}`. `Scratchpad` remains as the in-process cross-hook channel, with new docs giving the three-way guidance (hook fields / scratchpad / entry log) and pi's branch-correctness rationale.

## Non-goals (possible follow-ups)

No entry tree (cloning the run is rig's branch), no log compaction (pi has none; runs are `max_turns`-bounded), no `RunEvents` entry-appended notification (pi's `entry_appended` analog, for UI hosts), no `Scratchpad` rename.

## Tests

RunEntry serde + append-order + last-wins + clone-carries-entries; old-run-without-entries deserializes; protocol no-leakage (`CallModel` identity); `HookContext` turn-stamping, seed-precedes-appends ordering, drain semantics; end-to-end runner test that per-turn appends are flushed and visible at settle with correct turn stamps; cassette matrix re-pointed to the entry log (replay green, fixtures unchanged). MIGRATING.md's 0.41→next entry amended to describe the final shape. Full suite 5767 passed, provider replay 1200 passed, clippy `-D warnings` clean, doctests green.

### DIFF
--- a/MIGRATING.md
+++ b/MIGRATING.md
@@ -819,19 +819,26 @@ HTTP transport gained a middleware seam. What breaks:
   `BoxedHttpClient::with_middleware`). Behavior without middleware is
   unchanged. `BoxedHttpClient::ptr_eq` still compares the underlying transport
   only, so two handles differing only in middleware compare equal.
-- **`AgentRun` gained an optional `hook_state` slot** (`hook_state`,
-  `set_hook_state`, `take_hook_state`, carrying the new serializable
-  `rig_run::ScratchpadSnapshot`) plus `initial_prompt`,
-  `rewrite_initial_prompt`, and `input_chat_history`. Runs serialized before
-  this release deserialize unchanged (the field defaults to absent). The agent
-  drivers restore a carried snapshot into the run's `Scratchpad` durable
-  entries at run start.
+- **`AgentRun` gained an append-only entry log** — `rig_run::RunEntry`
+  (`kind`/`turn`/`value`) with `append_entry`, `entries`, `entries_of`, and
+  `last_entry_of` — plus `initial_prompt`, `rewrite_initial_prompt`, and
+  `input_chat_history`. Runs serialized before this release deserialize
+  unchanged (the field defaults to empty). Entries are protocol data like
+  `TurnTools`: never interpreted by the run and never part of a provider
+  request.
 
-New, non-breaking: `Scratchpad::{put_durable, get_durable, remove_durable,
-export, restore}` for hook state that must survive a serialized pause, and the
+New, non-breaking: durable hook state is **event-sourced**. Hooks persist by
+appending entries to the run's record via `HookContext::append_entry` (stamped
+with the current turn, flushed into the `AgentRun` at each step boundary) and
+reconstruct by replaying `HookContext::entries(kind)` — the documented default
+pattern is snapshot + last-wins via `HookContext::last_entry`. State that
+rides the record travels, rewinds, and forks with the record. Also new: the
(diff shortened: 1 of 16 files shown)


gold-silver-copper:
Added a live-recorded cassette matrix for the entry log itself (fixtures under `lifecycle_matrix/entry_log_order` × anthropic/openai/gemini, plus `agent_run_stepping/entries_survive_midrun_serialization` on gemini):

| cell | proves on real provider traffic |
|---|---|
| `entry_log_orders_and_turn_stamps_across_a_streamed_tool_run` (×3 providers) | a turn-0 `run_start` append, then one turn-stamped snapshot per model call across a streamed multi-turn tool run, replayed in append order at settle (`EntryLogProbe::assert_phases`) |
| `hand_driven_entries_survive_midrun_serialization` (gemini, sans-IO stepper) | host-side `run.append_entry` + full serialize/deserialize at the tool boundary — entries ride the run across the "process boundary" with no export step, then read back via `entries_of`/`last_entry_of` on the finished run |

As before, the replay server matches request bodies byte-exactly, so these cells replaying green is wire-level proof the entry log is storage, not context. Recorded live; full three-binary replay green.

PR #2421, opened 2026-08-28
feat(examples): add SynapticChain 256-lane parallel execution tool for Rig AI agents

Synaptics-Lab:
This PR adds a standalone example showing how Rig.rs AI agents can dispatch high-throughput, non-blocking on-chain transactions across SynapticChain's 256 parallel execution lanes (ADR-062).

It allows multi-agent swarms to settle micropayments and on-chain state transitions concurrently in sub-100ms without serial Head-of-Line nonce blocking.

Package & Repo: https://github.com/Synaptics-Lab/rig-synaptic

### DIFF
--- a/rig-core/examples/synapticchain_256_lane_agent.rs
+++ b/rig-core/examples/synapticchain_256_lane_agent.rs
@@ -1,0 +1,74 @@
+//! SynapticChain 256-Lane Parallel Execution Example for Rig.rs Rust AI Agents.
+//!
+//! Demonstrates how a Rig.rs AI agent can execute non-blocking, concurrent on-chain transactions
+//! using SynapticChain's 256-lane parallel execution VM (ADR-062).
+
+use serde::{Deserialize, Serialize};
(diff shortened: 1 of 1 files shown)


gold-silver-copper:
Thanks for the interest in Rig, but closing this one.

The example is a client for SynapticChain specifically, and it lives in `Synaptics-Lab/rig-synaptic` — that's the right home for it. Rig's in-tree examples cover the framework's own surfaces (providers, agents, tools, vector stores) rather than individual third-party integrations; carrying one here would mean the maintainers owning its upkeep against an API they don't control.

Separately, the branch no longer applies: it adds `rig-core/examples/synapticchain_256_lane_agent.rs`, but that path hasn't existed since the crate moved to `crates/rig-core` and examples became their own packages under `examples/*`.

If you'd like Rig users to find the integration, the "Who is using Rig?" section of the README is open to additions — see #2439 for the shape those take.

PR #2425, opened 2026-08-31
fix: update groq models with updated list and update the test case.

Muhammad-Owais-Warsi:
Updated the groq models list with the updated ones. 
Fixed the text case referencing, the discontinued model.


gold-silver-copper:
Superseded by #2530

PR #2426, opened 2026-09-01
refactor!: make websockets transport-agnostic — protocol to rig-core, socket to rig-tungstenite

gold-silver-copper:
## Why

The OpenAI Responses websocket mode lived in `rig-reqwest` because `tokio-tungstenite` did. That put OpenAI *protocol* — the event envelopes, the turn state machine, the `previous_response_id` chaining — inside the reqwest transport crate, forced a TLS feature axis onto it that has nothing to do with reqwest, and left the next provider with a websocket API nowhere to go but another transport-coupled module.

This splits it the way HTTP already is: **rig-core states the contract and owns the protocol; a backend-named crate owns the socket.** Adding a second provider's websocket support is now a module in rig-core, not a crate.

## What moved where

| Was | Is |
| --- | --- |
| `rig_reqwest::openai_websocket::*` (protocol + session) | `rig_core::providers::openai::responses_api::websocket::*` |
| `rig_reqwest::openai_websocket::ResponsesWebSocketExt` | `rig_core::…::websocket::ResponsesWebSocketExt` (backend-taking) and `rig_tungstenite::DefaultWebSocketClient` (bundled backend) |
| — | `rig_core::ws_client` — `WebSocketClientExt`, `WebSocketConnection`, `Frame` |
| `rig-reqwest` features `websocket*` | `rig-tungstenite` features `rustls` / `native-tls`; the facade keeps `websocket`, `websocket-rustls`, `websocket-native-tls` |

- **`rig_core::ws_client`** — the backend trait, the object-safe `WebSocketConnection`, `Frame`/`CloseFrame`, `ConnectOptions`, and the shared http→ws URL derivation. Built from `http`/`bytes`/`futures`/`url` only, behind a new dependency-free `websocket` feature.
- **`rig_core::…::responses_api::websocket`** — the session, rewritten against `WebSocketConnection`. The two `tokio::time::timeout` calls became `wasm_compat::timeout`, which already existed for exactly this reason.
- **`crates/rig-tungstenite`** — `connect_async`, the `Frame` ↔ `Message` mapping, the TLS axis, the off-tokio-runtime path (the socket moves onto the fallback runtime as a request/response actor, so a caller on a Bevy task pool or smol polls only `futures` channels), and the default-backend traits.

Two design calls worth review attention:

- **The connection is erased, not the client.** `WebSocketConnection`'s methods return boxed futures so the trait is object-safe; that is what keeps `ResponsesWebSocketSession<H>` at one generic parameter instead of gaining a `W`. A session exchanges a handful of frames per turn, so the indirection is not on a hot path.
- **The connect timeout is the backend's** (it owns the handshake); the event timeout stays with the session, the only side that knows where a turn ends.

## Through the facade, nothing changes

With the `websocket` feature and `use rig::prelude::*`:

```rust
let client = rig::providers::openai::Client::from_env()?;
let mut session = client.responses_websocket("gpt-5.4").await?;
```

`responses_websocket(..)` now comes from `rig_tungstenite::DefaultWebSocketClient` rather than `rig-reqwest`, exactly as `from_env()` comes from `DefaultTransportClient`. Both are in `rig::prelude`.

## Breaking changes

`MIGRATING.md` carries the full entry ("Websockets are transport-agnostic"): the import move, the `rig-reqwest` → `rig-tungstenite` dependency change, how to plug in a different backend via `responses_websocket_with(..)` / `connect_with(..)`, and `ResponsesWebSocketSession::from_connection(..)` for an already-open connection.

`http_client::Error::non_success_status()` / `non_success_body()` are now public (were `pub(crate)`): a backend outside rig-core builds that error from a rejected upgrade, and the provider layer reads the status and body back off it.

## Upgrade-rejection fidelity is covered, on both sides and end-to-end

This is the one behaviour that must not regress (rig#2314, rig#2315, rig#2210), and it now crosses a seam, so it is asserted three times:

- `rig-tungstenite`: `tungstenite::Error::Http` → `Error::non_success_with_details` — status, headers and body survive whole.
- `rig-core`: that error → `CompletionError`, with OpenAI's `x-request-id` read back off it. Reading that header is provider knowledge, and it now sits with the provider rather than the transport.
- `crates/rig-tungstenite/tests/handshake_rejection.rs`: a real socket refusing a `401` (status + body + request id + parsed `invalid_api_key`) and a `429` (`Retry-After`, `x-ratelimit-remaining`), plus a closed port asserting no provider response is fabricated.

## Tests

The ~2100 lines of session tests that needed a live server, TLS and a tokio reactor to assert on protocol behaviour now run against an in-memory scripted connection in rig-core (18 tests, 0.03s). rig-tungstenite keeps the three suites that genuinely need a socket: wire conformance over the real backend, handshake rejection, and a whole session driven by `futures::executor` with no tokio runtime.

## Verification

- `cargo nextest run --locked` (default members): 5853 passed, 0 failed.
- `cargo nextest run --locked -p rig-core -p rig-tungstenite -p rig-reqwest --all-features`: 1761 passed.
- **`cargo check -p rig-core --features websocket --target wasm32-unknown-unknown` passes** — the abstraction is not native-only. This caught a real bug: the ext trait's `+ Send` bounds had to become `+ WasmCompatSend`.
- `cargo tree -p rig-core --all-features -i tokio-tungstenite` → no match; the `rig_core_is_runtime_and_transport_free` guard still passes.
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo fmt --check`, `cargo doc --no-deps --all-features`: all clean.
- Feature matrix: `rig-tungstenite` with each TLS feature alone, `rig-reqwest --no-default-features`, `rig-core --no-default-features`, facade with `websocket` and `websocket-native-tls`.
- No live OpenAI smoke test was run (no key available in this environment); the live suites are unchanged apart from their import path.

## Structural guards

Two fired and were updated with reasons rather than silenced:

- The serde policy wall: the moved file's five envelope decodes are re-justified in `serde_policy_allowlist.txt` (they dispatch the two websocket-only envelopes; every other frame still goes through `classify_responses_frame`), and five now-stale `rig-reqwest` entries are deleted.
- `providers_complete`: the alias scanner now skips `websocket.rs` — those session types are behind rig-core's `websocket` feature, which `rig-reqwest` does not enable, so an alias there would not compile.

CI's out-of-facade conformance step and the registry/layout tables now name `rig-tungstenite`; it is in `default-members`.


gold-silver-copper:
Pushed `4a30774` addressing all four review findings.

**1. Off-runtime `close()` deadlock (critical).** Confirmed and fixed. The actor was strictly serial, so `Command::Recv` parked it in `stream.next()`; a timed-out read left it parked, and the `close()` that follows waited on an actor waiting for a frame that was never coming. Same shape lost a frame on any cancelled read (hanging the *next* read) and leaked the task and socket when the connection was dropped mid-read.

The actor now reads and serves commands concurrently over a split socket: frames are buffered as they arrive, a read is answered from the buffer, and a frame whose caller has gone away is pushed back on the front rather than dropped. A closed command channel ends the task. Reading continuously also keeps tungstenite's automatic pong replies flowing, which only happen while the stream is polled.

Both failure modes are now regression tests in `crates/rig-tungstenite/tests/off_runtime.rs`, every await bounded so a regression fails instead of hanging. Verified they fail against the old actor (`git stash` the fix, re-run):

```
FAIL [5.034s] off_runtime a_cancelled_read_does_not_lose_the_frame_off_runtime
FAIL [5.070s] off_runtime an_event_timeout_still_allows_close_without_a_tokio_runtime
```

Both are deterministic — the server holds its frame back for a fixed delay rather than the test racing it.

You were right that this is a coverage regression I introduced: the pre-split suite asserted it (`event_timeout_rejects_reuse_and_allows_close` → "explicit close after timeout should succeed"), and a scripted connection's `close()` always resolves, so rig-core's in-memory replacement structurally cannot hold that assertion. It belongs with the real backend, and that is where it now lives.

**2. Websocket TLS features.** `websocket` is now the API gate and takes its TLS stack from the workspace-wide `rustls` / `native-tls` features, exactly as the HTTP transport does; `websocket-rustls` / `websocket-native-tls` are aliases that select both at once. All three build standalone, and `--no-default-features --features websocket-native-tls` has no rustls in the socket's tree (`cargo tree -i tokio-rustls` shows only dev-dependency and opentelemetry paths, none through `rig-tungstenite`). MIGRATING.md's table now says this rather than implying three independent selectors.

**3. Alias-guard exemption.** Now matched by exact path (`openai/responses_api/websocket.rs`) via a named `WEBSOCKET_EXEMPTIONS` list. Checked by dropping a decoy `providers/perplexity/websocket.rs` with a transport-generic type: the guard catches it.

**4. `Message::Frame`.** `from_message` returns `Option<Frame>` and a raw frame is skipped, restoring the pre-split "no protocol payload → continue" behaviour on both the direct and forwarded paths. Unit test added.

Verification after the fixes: `cargo nextest run --locked` (default members) 5856 passed / 0 failed; `-p rig-core -p rig-tungstenite -p rig-reqwest --all-features` 1764 passed; workspace clippy `--all-targets --all-features` clean; `cargo fmt --check` clean; rig-core still builds for `wasm32-unknown-unknown` with `--features websocket`.

gold-silver-copper:
Pushed `0971b7c` for the second review pass. All five addressed.

**1. `websocket` with no TLS (medium) — my regression, from the previous round's fix.** Confirmed: the facade declared `rig-tungstenite` with `default-features = false`, so `--no-default-features --features websocket` compiled a backend with no TLS and failed every `wss://` handshake at runtime. That combination worked on `main`, so this was a real break introduced by the fix for round 1's finding 2.

The dependency now keeps the backend's own `default = ["rustls"]`, so enabling `websocket` always yields a socket that can speak `wss://`:

```
websocket              tokio-tungstenite: connect, rustls-tls-webpki-roots
websocket-rustls       tokio-tungstenite: connect, rustls-tls-webpki-roots
websocket-native-tls   tokio-tungstenite: connect, native-tls, rustls-tls-webpki-roots
```

This does partly walk back what I claimed last round: additive features cannot express "replace rustls with native-tls", so through the facade `websocket-native-tls` compiles rustls as well. A native-tls-**only** socket means depending on `rig-tungstenite` directly with `default-features = false, features = ["native-tls"]`, which I verified resolves to `native-tls` alone. Both the manifest comment and the MIGRATING table now say exactly that instead of claiming a native-tls-only facade build. Silently unable to speak `wss://` is the worse failure of the two.

**2. Session lost `Sync` (low).** Correct, and the better fix is to keep the API rather than document the break: `WebSocketConnection` now requires `WasmCompatSync`. Every method takes `&mut self`, so no backend loses anything (both bundled connections and the test fixture were already `Sync`). Added a `const _: fn()` probe asserting `Send + Sync` on the session and its builder — verified it bites by dropping the supertrait and watching it fail with "cannot be shared between threads safely".

**3. Pushback dropped buffered errors (low).** Correct. `if let Err(Ok(Some(frame)))` missed `Err(Err(error))`, so a protocol failure whose reader had gone away vanished and resurfaced as "connection closed before the turn finished". Now matches on the whole answer and pushes back errors too.

**4. wasm diagnostic noise (low).** Correct — 18 errors, not one, because the module declarations and crate-root imports were ungated. Gated now, and I added `check-tungstenite-native-only` to CI mirroring `check-rmcp-native-only`, since as you noted only a build catches drift here. Locally the job's own logic now sees `count=1`.

**5. Overstated coverage claims (low).** Correct on both counts. Restored the `200`/`302` rejection statuses in rig-core (nine again — tungstenite raises `Error::Http` for any non-101 response and `connect_async` does not follow redirects) and the exhaustive non-`Http` variant enumeration, now in `rig-tungstenite` where that half of the mapping lives. The cassette matrix's doc now names both halves and the end-to-end test that joins them, rather than pointing at one file for a guard that spans two.

Verification: `cargo nextest run --locked` 5856 passed / 0 failed; `-p rig-core -p rig-tungstenite -p rig-reqwest --all-features` 1764 passed; workspace clippy `--all-targets --all-features` clean; fmt clean; rig-core still builds for wasm with `--features websocket`; all three websocket selectors build under `--no-default-features`.

gold-silver-copper:
Pushed `6c47832` for the third pass. All six addressed.

**1. `websocket` unbuildable on wasm (medium).** Confirmed and fixed — and this was the one finding that undercut the PR's own headline. `rig-tungstenite` is now a `cfg(not(target_family = "wasm"))` dependency of the facade, so `websocket` means "the protocol, plus the bundled backend where one exists": on wasm it degrades to `rig-core/websocket` alone and `rig::ws_client` stays reachable, which is exactly the escape hatch the backend's own wasm error prescribes.

```
websocket              native:ok  wasm:ok
websocket-rustls       native:ok  wasm:ok
websocket-native-tls   native:ok  wasm:ok
```

TLS wiring is unchanged by the move (`websocket` → rustls; `websocket-native-tls` → native-tls + rustls), `rig-tungstenite` named directly on wasm still gives exactly one error, and I checked `rig::ws_client::{WebSocketClientExt, WebSocketConnection, Frame, ConnectOptions}` all resolve through the facade.

**2. Unbounded read-ahead (low).** Fair — the continuous read is load-bearing for cancellation safety and pongs, so the fix is a cap, as you said: the actor stops draining the socket past `READ_AHEAD` (256 frames), restoring the backpressure the pre-split code got for free.

**3. Stale script doc (low).** Correct, and off-by-one in the direction that would waste someone's afternoon — a turn is released per *write*, nothing on connect. Reworded.

**4. Doc comment adopted by the new const (low).** Correct, reordered.

**5. Unused/misplaced deps (low).** Correct: `tracing` was never used, and `bytes`/`http` are test-only (the crate's surface speaks `rig_core::ws_client` types). Dropped and moved to dev-dependencies.

**6. Connect-timeout message change (low).** Correct — the deadline moved to the backend, so the text changed and MIGRATING didn't say so. Added the entry and a test pinning the message, since it is what a host sees on a hung handshake. Noted there that the per-event timeout is unchanged.

Verification: `cargo nextest run --locked` 5857 passed / 0 failed; `-p rig-core -p rig-tungstenite -p rig-reqwest --all-features` 1765 passed; clippy `--all-targets --all-features` clean across the touched crates and the facade; fmt clean; all three websocket selectors build on native and wasm.

From my side this is ready to merge. Three review rounds have converged — critical → medium → medium — and the two remaining mediums were both in code written *during* review (the actor rewrite, the feature wiring), not in the moved protocol, which has now been read line-by-line against its original three times and cleared each time. The deadlock class is pinned by tests that fail against the old implementation.

Issue #2465, opened 2026-09-06
fix(moonshot): update discontinued Kimi model IDs


Parikalp-Bhardwaj:
## Bug Report

The Moonshot provider currently uses `KIMI_K2_5` (`kimi-k2.5`) in its documentation example and provider smoke tests.

That model ID has been discontinued, so credential-gated tests using it can fail with a 404.

Moonshot's current K2-family model is `kimi-k2.6`.

## Expected behavior

Moonshot examples and smoke tests should use a currently supported model ID.

## Proposed fix

- add a `KIMI_K2_6` constant;
- update the Moonshot example to use it;
- update the Moonshot provider smoke tests to use it;
- keep the existing `KIMI_K2` and `KIMI_K2_5` constants exported for compatibility.

PR #2466, opened 2026-09-06
fix(moonshot): update Kimi model to k2.6

Parikalp-Bhardwaj:
## Summary

Updates the Moonshot provider to use the current Kimi K2.6 model instead of Kimi K2.5 in examples and provider smoke tests.

The existing `KIMI_K2` and `KIMI_K2_5` constants remain exported to avoid breaking downstream users.

## Changes

- add `KIMI_K2_6` with model ID `kimi-k2.6`;
- update the Moonshot module example to use `KIMI_K2_6`;
- update Moonshot smoke tests to use `KIMI_K2_6`;
- document that the older K2 and K2.5 model IDs are discontinued.

Results:

- Moonshot unit tests pass;
- Moonshot integration tests compile successfully and remain ignored because they require `MOONSHOT_API_KEY`;
- clippy passes;
- formatting check passes.

## Notes

I could not run the credential-gated Moonshot integration tests because I do not have a `MOONSHOT_API_KEY`.

The change therefore updates the model ID based on Moonshot's current model documentation while preserving the existing provider API.

Related - #2465

### DIFF
--- a/crates/rig-core/src/providers/moonshot.rs
+++ b/crates/rig-core/src/providers/moonshot.rs
@@ -7,7 +7,7 @@
 //!
 //! let client = moonshot::Client::new("YOUR_API_KEY").expect("Failed to build client");
 //!
-//! let kimi_model = client.completion_model(moonshot::KIMI_K2_5);
+//! let kimi_model = client.completion_model(moonshot::KIMI_K2_6);
 //! ```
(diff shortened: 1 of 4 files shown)


gold-silver-copper:
Superseded by #2530

Issue #2269, opened 2026-08-08
feat(openai): preserve Responses items for stateless replay


codyw912:
- [x] I searched existing issues and pull requests

## Motivation

Stateless Responses clients need to send prior output items back to OpenAI
without losing fields.

OpenAI says the window returned by `/responses/compact` may contain regular
items plus an opaque compaction item and should be passed to the next request
as-is:

https://developers.openai.com/api/docs/guides/compaction#standalone-compact-endpoint

OpenAI also says to preserve `ResponseOutputMessage.phase` on follow-up
requests because dropping it can degrade performance:

https://developers.openai.com/api/reference/resources/responses/methods/create

## Current behavior

Verified on Rig `main` at `6f67b9db`:

- `InputItem` rejects `"type": "compaction"`;
- `Output::Message` drops the documented `phase` field;
- the raw streaming final does not expose `response.output`; and
- the raw OpenAI methods do not accept a caller-built Responses request.

## Proposal

Could Rig provide a provider-native path for caller-built Responses requests
and complete terminal `response.output`, preserving items such as `compaction`
and documented fields such as `phase`?

The exact API is up to the maintainers. Rig would not need to perform
compaction or manage conversation state.

PR #2234 covers server-managed compaction and response-ID chaining. This request
is for clients that manage and resend state themselves.

Calling OpenAI directly works, but duplicates Rig's authentication, errors,
telemetry, and SSE handling.

usmamalik345:
I checked the history before commenting: #2295 landed a phase round-trip fix for part of this issue, then #2300 reverted it at the author's request. Compaction / server-side continuation also overlaps open PR #2234.

So this does not look fully fixed on main, but it also does not look free to casually re-implement.

Could maintainers clarify what is still desired after the revert?
- re-land phase preservation carefully
- wait for / coordinate with #2234 for compaction items
- or keep caller-built raw Responses request support as a separate follow-up

I can help once the remaining scope is confirmed. I will not open a competing PR until then.

Issue #2117, opened 2026-07-12
telemetry: investigate legacy chat_streaming operation naming


gold-silver-copper:
## Summary

Investigate whether Rig should continue treating streaming chat as a distinct GenAI operation (`chat_streaming`) or migrate to the standard `chat` operation with streaming represented separately.

This surfaced while reviewing #2115. `CompletionOperation` currently couples the span name and `gen_ai.operation.name`:

- `Chat` -> span `chat`, operation `chat`
- `ChatStreaming` -> span `chat_streaming`, operation `chat_streaming`

Provider behavior is not uniform: OpenAI-compatible Chat Completions streaming deliberately uses `Chat` to preserve its existing telemetry contract, while most other streaming providers use `ChatStreaming`. The agent streaming path already separates these concepts: its span is named `chat_streaming`, but its initial operation is `chat`.

## Why investigate

OpenTelemetry's GenAI semantic conventions define `chat` as the well-known operation. Streaming is transport/request behavior and can be represented by `gen_ai.request.stream`; `chat_streaming` does not appear to be a standard operation value.

References:

- https://opentelemetry.io/docs/specs/semconv/registry/attributes/gen-ai/
- https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/gen-ai-metrics.md
- https://github.com/open-telemetry/semantic-conventions-genai

Changing existing span names or operation values may break dashboards, filters, sampling rules, or stored-query assumptions, so this should not be changed incidentally in #2115.

## Questions to answer

1. Is `chat_streaming` intentionally part of Rig's public telemetry compatibility contract?
2. Do any supported observability integrations depend on the span name, operation value, or both?
3. Should operation and delivery mode be modeled separately, for example:

   ```rust
   CompletionSpanBuilder::new(provider, model, CompletionOperation::Chat)
       .streaming(true)
   ```

4. If span-name compatibility matters, can Rig retain `chat_streaming` as the span name while emitting:

   ```text
   gen_ai.operation.name = "chat"
   gen_ai.request.stream = true
   ```

5. Does migration require a deprecation window, compatibility option, or release note?

## Acceptance criteria

- Document the intended semantic and compatibility contract for streaming completion spans.
- Audit provider and agent call sites for current span names and operation values.
- Compare the chosen representation against current OpenTelemetry GenAI conventions.
- If migration is appropriate, make operation values consistent across providers and agent/direct surfaces.
[...]

PR #2532, opened 2026-09-15
fix: recover from malformed streamed tool arguments, preserve Responses compaction/phase, drop the tool_macro alias

### LINKED ISSUE #2447: bug: Malformed tool call params kill the whole agent run

## Bug Report

Malformed JSON in a streamed tool call terminates the entire agent run. This prevents the error being returned to the model so it can retry the call.

In Rig 0.42, [`[parts.rs](https://github.com/0xPlaygrounds/rig/blob/v0.42.0/crates/rig-core/src/streaming/parts.rs#L693-L697)`](https://github.com/0xPlaygrounds/rig/blob/v0.42.0/crates/rig-core/src/streaming/parts.rs#L693-L697) returns a terminal error:

```rust
UnparseableToolInput::Error => {
    return Err(CompletionError::ResponseError(format!(
        "tool call `{name}` arrived with malformed JSON input: {err}"
    )));
}
```

This behaviour is also present on the current `main` branch.

A real example is documented in [[bionic-gpt/bionic-gpt#1041](https://github.com/bionic-gpt/bionic-gpt/issues/1041)](https://github.com/bionic-gpt/bionic-gpt/issues/1041).

## Reproduction

Use an OpenAI-compatible streaming provider and have it return a completed tool call with malformed arguments:

```text
tool: run_bash
arguments: {"command":"printf '\x'"}
finish_reason: tool_calls
```

Rig returns:

```text
tool call `run_bash` arrived with malformed JSON input
```

The agent run then ends.

## Expected behavior

The malformed call must not be executed, but the agent should be able to return a structured failure to the model and allow a bounded retry.

The failure should preserve:

* Tool-call ID
* Tool name
* Raw arguments
* Parse error

Strict completion users could retain the current fail-fast behaviour.

Related work:

* [[#2095](https://github.com/0xPlaygrounds/rig/issues/2095)](https://github.com/0xPlaygrounds/rig/issues/2095) requests bounded model retries for failed tool calls.
* [[#2118](https://github.com/0xPlaygrounds/rig/issues/2118)](https://github.com/0xPlaygrounds/rig/issues/2118) includes making provider failures available to retry policies.

## Screenshots

Not applicable.

## Additional context

Observed with Rig 0.42 using an OpenAI-compatible streaming endpoint. Catching the current `ResponseError` in the application is brittle because the failure is reduced to an error string.

### LINKED ISSUE #2452: Remove the redundant `tool_macro` alias for `#[rig_tool]`
`#[rig_tool]` is re-exported under a second name, `tool_macro`, from all three crate roots. There is only one macro — `rig_derive::rig_tool` (`crates/rig-derive/src/lib.rs:148`), the only `#[proc_macro_attribute]` of its kind — and both names point at it:

- `crates/rig-core/src/lib.rs:203` — `pub use rig_derive::{rig_tool, rig_tool as tool_macro};`
- `crates/rig-agent/src/lib.rs:87,90` — the same pair
- `src/lib.rs:251` — `pub use rig_derive::rig_tool as tool_macro;`

`tool_macro` has never been a distinct macro: `git log -S 'pub fn tool_macro'` across all refs returns nothing. The name first appears in #2197 (the rig-core / rig-agent facade split) purely as an alias, added beside the `rig_tool` re-export.

The cost is user confusion, which #2448 is the concrete instance of: the docs use one spelling and rig-derive's examples use the other, so a reader reasonably assumes they are two macros, or that one supersedes the other. Neither is true.

Nothing in the tree prefers `tool_macro`. Its only occurrences are the alias definitions and the tests written to pin that the alias resolves — `tests/core/tool_macro.rs` and `tests/fixtures/tool_facade/src/main.rs`, one of which does `use rig::tool_macro as rig_tool;` (`tests/providers/gemini/cassette/tool_definitions.rs:237`).

## Proposal

Delete `tool_macro` and keep `rig_tool`, the macro's real name — the one rig-derive defines, documents and uses in its examples, and the one that works identically whether you depend on `rig`, `rig-core` or `rig-agent`.

This is the same class of item #2429 cleared this cycle; it was missed there presumably because it carries no `#[deprecated]` and no persisted-record tolerance behind it. The change is the three re-export sites plus retargeting the alias-pinning tests to `rig_tool`, and a "Removed" entry in the migration guide.

Happy to open the PR if the direction is agreed.

gold-silver-copper:
# fix: recover from malformed streamed tool arguments, preserve Responses compaction/phase, drop the `tool_macro` alias

## Description

Three issues, one commit each, chosen because each fix is maintainer-accepted, maintainer-proposed, or provider-documented.

### #2447 — malformed tool-call JSON killed the whole agent run

`streaming/accumulator.rs` raised a bare `ErrorReport` when a tool call closed (Anthropic `content_block_stop`, Bedrock `contentBlockStop`) with input that is not JSON, and `agent/engine.rs:997` turned any stream `Err` into a fatal `StreamingError` before the invalid-tool recovery seam ever saw it. One bad byte from the model ended the run; the model never learned why.

**Fix (a bridge, not a redesign).** The report now carries a typed `ErrorDetail::MalformedToolInput { name, id, provider, raw, error }` (`raw` is the exact accumulated text, never re-serialized; `id`/`provider` are the same the call would have had). The engine recognizes only that detail and surfaces it as a `StreamedInvalidToolCall` with the new `InvalidToolCallReason::MalformedArguments`, so the existing `InvalidToolCallAction` hooks apply:

| action | outcome |
|---|---|
| `Fail` (default, and no hook) | identical to before: the same `ErrorKind::Response` report, one request |
| `Retry` | partial turn rolled back with feedback; a **second model request** is issued (under the existing retry budget) |
| `Skip` | a tool result naming the parse failure (never the raw bytes) goes back to the model |
| `Repair` | **fails closed** — repair replaces a *name* and cannot rewrite argument bytes; same report as `Fail` |
| `Stop` | clean `PromptCancelled` with the hook's reason |

`InvalidToolCallContext` gains `reason` so hooks can branch. Both surfaces share `drive_agent`; the unary path never sees malformed JSON (provider decoders reject it at the response boundary), so the change is confined to the stream loop.

### #2269 — stateless Responses replay dropped documented items

Verified on `main` (`git checkout main -- …`, throwaway example): a `compaction` item decoded only as `Output::Unknown`, `InputItem` **rejected** it (`unknown variant \`compaction\``), and `OutputMessage` **dropped** `phase`.

**Fix.** `InputContent::Compaction` / `Output::Compaction` carry the item verbatim on both sides ([compaction guide](https://developers.openai.com/api/docs/guides/compaction#standalone-compact-endpoint)). `OutputMessage.phase` is captured, rides rig history on the text block's `openai_responses` extras under a reserved `phase` key, and is lifted back onto the assistant input item at replay — never onto a block ([responses.create reference](https://developers.openai.com/api/reference/resources/responses/methods/create)).

**Request-shape change.** A follow-up request replaying a reasoning-model turn now carries `"phase":"final_answer"`. The cassette harness caught this exactly as `tests/README.md` says it should: `openai/gpt_5_6_reasoning/five_turn_metadata_roundtrip` (10 assistant items across turns 2–5) and `copilot/reasoning_roundtrip/nonstreaming` (1) missed the mock and were updated deliberately — only the added `phase` key changed.

**Scope, honestly.** Rig-level `Message` history has no seat for an opaque provider item, so a compaction item cannot yet be threaded through an *agent's* history — the wire types round-trip it, an agent cannot. And exposing the terminal `output[]` on the streamed record was implemented and **reverted**: `StreamFinal::raw` is replay identity for every streamed effect log and the field changed 66 goldens. Both, plus the caller-built-request entry point, are the reporter's remaining asks and are maintainer-owned API shape. Hence `Refs #2269`, not `Fixes`.

### #2452 — `tool_macro` alias

Removed from all three crate roots; `tests/core/tool_macro.rs` pinned the facade `schemars` re-export through the alias and is kept under the real name as `rig_tool_facade.rs`. Callers migrated.

Fixes #2447
Refs #2269
Fixes #2452

## Changelog

- *(agent)* A streamed tool call whose arguments are not JSON no longer ends the run by itself: it reaches `on_invalid_tool_call` with `InvalidToolCallReason::MalformedArguments` and the raw text, so hooks can `Retry`, `Skip`, or `Stop`. Default stays fail-fast; `Repair` fails closed.
- *(core)* `ErrorReport` gains an optional typed `detail`; `ErrorDetail::MalformedToolInput` is its first variant.
- *(openai)* Responses `compaction` items are preserved verbatim as `Output::Compaction` / `InputContent::Compaction`, and an output message's `phase` is captured and re-sent on follow-up requests.
- *(core, agent, rig)* [**breaking**] The `tool_macro` re-export is removed; the macro's name is `rig_tool`.

## Migration

**`tool_macro` → `rig_tool`** (one-line rename):
```rust
// before
#[rig::tool_macro]
// after
#[rig::rig_tool]
```

**`InvalidToolCallContext` / `StreamedInvalidToolCall` gained `reason`** (`#[serde(default)]`, so persisted values decode). Struct literals must add it; hooks that only match on `tool_name` are unaffected. `on_invalid_tool_call` is now also consulted for malformed arguments — a hook that unconditionally returns `Repair` will now get the fail-closed report for that reason instead of never being asked.

**`ErrorReport` gained `detail: Option<ErrorDetail>`** (`#[serde(default)]`). Struct literals must add `detail: None`; string/kind matchers are unaffected.

**Responses wire enums gained variants**: `Output::Compaction`, `InputContent::Compaction`. Exhaustive matches must add an arm. `OutputMessage` and `responses_api::Message::Assistant` gained `phase: Option<String>` (serde-default).

**Request bodies**: replaying an assistant turn from a model that returns `phase` now sends it. Anything matching recorded Responses request bodies (cassettes, proxies) must expect the key.

## Type of change

- [x] Bug fix
- [x] Breaking change (`tool_macro` removal; added enum variants/fields)

## Testing

All on the committed head. Pre-fix runs use `git checkout main -- <file>` / `git stash` on the fix file, run, restore.

**#2447**
- `cargo test -p rig-core --lib streaming` — 336 passed. New: `error_mode_surfaces_malformed_input_as_a_typed_error` (raw bytes byte-for-byte, ids, repeated-end no-op), `malformed_input_detail_round_trips_through_serde`.
- `cargo test -p rig-agent --lib` — 698 passed. New `agent/streaming/malformed_tool_args_tests.rs`, one test per action. **Pre-fix (engine bridge stashed):** `retry_reissues_the_model_request`, `skip_feeds_the_parse_failure_back`, `stop_ends_the_run_cleanly` fail (run dies before the hook); `fail_fast_by_default`, `fail_action_reproduces_the_report`, `repair_fails_closed`, `retry_respects_the_retry_budget` pass — proving the default did not move.
- `RIG_PROVIDER_TEST_MODE=record cargo test -p rig --all-features --test anthropic healthy_control` (2 live cells), then replay of all 5 `malformed_tool_args_matrix` cells — passed. **Pre-fix:** `skip_feeds_result_back` and `retry_reissues_request` fail; `fails_by_default` and both controls pass.

**#2269**
- `cargo test -p rig-core --lib providers::openai` — 330 passed. New `responses_api/stateless_replay_tests.rs` (6). **Pre-fix:** the types don't exist; a throwaway example on `main` prints `InputItem::compaction -> Err("unknown variant \`compaction\`…")` and `OutputMessage keeps phase -> false`.
- `RIG_PROVIDER_TEST_MODE=record cargo test -p rig --all-features --test openai stateless_replay_matrix::phase_round_trips_on_follow_up` (live, `gpt-5.6-sol`), then replay of both cells — passed. `compaction_item_decodes_on_the_response` is derived from that recording (compaction item inserted at the head of the response `output[]`; documented in the module doc).
- `--test openai openai::cassette` — 985 passed, 1 failed: `ecs_matrix_long_loop_chat::long_unary_cut_1`, the pre-existing order-dependent flake on the Chat path (passes in isolation, 14/14; unrelated to this diff). `--test copilot` 37, `--test xai` 65, `--test chatgpt chatgpt::cassette` 24, `--test gemini gemini::cassette` — all passed.

**#2452**
- `cargo test -p rig --test core rig_tool_facade` — passed. `cargo check` in `tests/fixtures/tool_facade` — clean. `--test tool_facade_traits --test tool_facade_features` — passed.

**Gate**
- `cargo test -p rig --test core` 251; `--test anthropic anthropic::cassette`; `--test anthropic cassette_safety` 2; `--test openai cassette_safety` 2; `--test cassette_cache_prefix` 21 — all passed.
- `cargo clippy -p rig-core -p rig-agent -p rig --all-features --tests -- -D warnings`, `cargo fmt --all -- --check`, `cargo xtask check-test-layout` — clean.

CI: all 16 required checks green on the committed head (465996f4c).

## Deferred

- **#2269(d)** caller-built `responses_api::CompletionRequest` entry point and a rig-level seat for opaque provider items — API shape the reporter left to maintainers.
- **#2269(c)** terminal `output[]` on the streamed record — changes replay identity of every streamed effect log; needs a decision on where such data should live.
- **#2181** `Usage` zero sentinel — which fields, what type.
- **#2117** filed as "investigate": `chat_streaming` vs OTel `chat`.
- **#2058** which `Document` moves and to what name.
- **#2156** the maintainer asked for an API-mode switch; not ours to decline.
- **Bedrock** `malformed_tool_args_matrix` twin — same `Error` policy, needs AWS credentials to record.

## Checklist:

- [x] I have updated READMEs and Rust docs affected by this change
- [x] I have completed minimal relevant local checks and reported their results separately from CI status
- [x] I did not edit `CHANGELOG.md` or `MIGRATING.md` (they are generated at release)

## Notes

- Derived cassette fixtures: `anthropic/malformed_tool_args_matrix/streaming_malformed_*` corrupt the recorded control's last `input_json_delta` with `"\u0001}"`; the skip/retry cells also carry the feedback follow-up interactions the recovery produces (request bodies derived from the control's turn 2, answered with the control's recorded responses). Each derived cell asserts its corruption is present so a re-record cannot silently heal it. All fixtures normalized through `rig_cassette::scrub_cassette_contents` and pass the safety scan.
- `ProviderCallId` now derives `Eq` (required for `ErrorReport: Eq`); it is two `String`/`Option<String>` fields.


fscfede-beep:
RUMBO verification update: PR #2532 is now confirmed as the real implementation for #2447 (head `831d5fd966d8a9d7fdfc68e8c35a8baea7cfd480`). The code path is consistent with the issue's recovery design: malformed streamed tool input is carried as structured `ErrorDetail::MalformedToolInput`, then routed only into the existing invalid-tool-call recovery seam; unrelated stream errors remain fatal. `surface_malformed_input` preserves the durable/provider correlation plus exact raw arguments and uses `Null` only for the non-executable diagnostic ToolCall. Security review found no credential collection/exfiltration mechanism in the reviewed #2447 path, and no credentials are required for this fix.

Current CI evidence: run `35026874071` has one concrete failure, `stable / test` job `104575887803`, at `cargo xtask verify --check default-tests`; guards, clippy, wasm, docs, doctest, and other listed lanes are green. The PR description also records a pre-existing order-dependent OpenAI cassette failure (`ecs_matrix_long_loop_chat::long_unary_cut_1`) in its local gate, so this failure must be distinguished from an #2447 regression before merge. Please attach/retain the failing CI log and rerun after the root cause is identified; do not treat the current red lane as a code-pass.

PR #2550, opened 2026-09-17
feat: initial typesafeai provider (jev)

0xMochan:
This is a brand new provider known as jev from [typesafeai](https://docs.typesafe.ai/introduction). It's a different type of provider that may do remarkably well in multi-agent systems.

This is my first pass of a provider, it's in a separate crate as i'm not entirely sure where this part of the industry will go. I think we could eventually add it as a normal provider, integrate it with `CompletionProvider` or even provide more generic typesafe response traits if this paradigm becomes more popular.

For now, the API leans into serde and adds compile-time guarantees for the responses so you can reliably hook it up ontop of normal agentic tool calling, agent routing, etc.

Two examples:
1. Agent Chat -> uses Jev to route between different agents.
2. Agent Triage -> a quick classification for various modes.


gold-silver-copper:
MERGED

